PR #175, opened 2025-08-30
Support receiving for external pks

benthecarman:
Was looking into if it is possible to create a LN address server for spark and seems like it is possible. I think only sdk change needed is this?

### DIFF
--- a/crates/spark/src/services/lightning.rs
+++ b/crates/spark/src/services/lightning.rs
@@ -224,7 +224,12 @@ impl<S: Signer> LightningService<S> {
         memo: Option<String>,
         preimage: Option<Vec<u8>>,
         expiry_secs: Option<u32>,
+        identity_pubkey: Option<PublicKey>,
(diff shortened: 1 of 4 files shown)


crates/spark/src/services/lightning.rs
```diff
@@ -225,6 +225,40 @@ impl<S: Signer> LightningService<S> {
         preimage: Option<Vec<u8>>,
         expiry_secs: Option<u32>,
     ) -> Result<LightningReceivePayment, ServiceError> {
+        self.create_lightning_invoice_internal(amount_sats, memo, preimage, expiry_secs, None)
+            .await
+    }
+
+    pub async fn create_lightning_invoice_for_pubkey(
```
roeierez, line 232: Thanks @benthecarman just a small suggestion. Perhaps instead of adding another function we should add an optional publicKey parameter to the existing create_lightning_invoice? If passed None we will use the pubkey from the signer. 
WDYT?

PR #185, opened 2025-09-01
cache static deposit address

roeierez:
in case address generation fails (due to already exists) spark returns error.
Static addresses aren't changed so no reason not to cache it.
Three steps to get the address:
1. get from cache
2. query existing addresses
3. If nothing exists, generate a new one


dangeross (approved):

danielgranhao (approved):
LGTM

PR #191, opened 2025-09-01
Support spark address

roeierez:
Add missing spark address parsing support

### DIFF
--- a/crates/breez-sdk/common/Cargo.toml
+++ b/crates/breez-sdk/common/Cargo.toml
@@ -37,6 +37,9 @@ strum = { workspace = true, features = ["derive"] }
 thiserror.workspace = true
 tokio.workspace = true
 tracing.workspace = true
+spark.workspace = true
+web-time.workspace = true
+spark-wallet.workspace = true
 uniffi = { workspace = true, optional = true }
 
 # Non-Wasm dependencies
--- a/crates/breez-sdk/common/src/input/models.rs
+++ b/crates/breez-sdk/common/src/input/models.rs
@@ -158,6 +158,119 @@ pub enum InputType {
     Bip21(Bip21Details),
     Bolt12InvoiceRequest(Bolt12InvoiceRequestDetails),
     LnurlWithdraw(LnurlWithdrawRequestDetails),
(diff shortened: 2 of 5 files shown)


crates/breez-sdk/common/src/input/models.rs
```diff
@@ -158,6 +158,119 @@ pub enum InputType {
...
+    pub encoded_address: String,
+    pub decoded_address: SparkAddress,
+    pub source: PaymentRequestSource,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct SparkAddress {
+    pub identity_public_key: String,
+    pub network: BitcoinNetwork,
```
JssDWt, line 176: This should be a spark network, not a bitcoin network? In fact they are the same, so maybe it doesn't really matter.

crates/breez-sdk/common/src/input/models.rs
```diff
@@ -158,6 +158,119 @@ pub enum InputType {
     Bip21(Bip21Details),
     Bolt12InvoiceRequest(Bolt12InvoiceRequestDetails),
     LnurlWithdraw(LnurlWithdrawRequestDetails),
+    SparkAddress(SparkAddressDetails),
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct SparkAddressDetails {
+    pub encoded_address: String,
```
dangeross, line 167: Could we call it `address` like in BitcoinAddressDetails?

crates/breez-sdk/common/Cargo.toml
```diff
@@ -37,6 +37,9 @@ strum = { workspace = true, features = ["derive"] }
 thiserror.workspace = true
 tokio.workspace = true
 tracing.workspace = true
+spark.workspace = true
+web-time.workspace = true
```
JssDWt, line 41: I don't believe this is used?

dangeross (approved):

PR #196, opened 2025-09-02
Support description hashes for invoices

benthecarman:
Last PR I need for now (tm)! This just adds basic support for description hashes in invoices. I added an enum to enforce you can't set both desc hash and memo.


danielgranhao (approved):
LGTM

dangeross (approved):

PR #197, opened 2025-09-04
stateless builder

JssDWt:
By making the sdk builder stateless, there is no mutable reference needed. Therefore there is no difference between uniffi and non-uniffi builders.


danielgranhao (approved):

dangeross (commented):
The SdkBuilder needs exposing to uniffi

PR #205, opened 2025-09-05
lnurl server

JssDWt:
lnurl server implementation.

A user can register for a lightning address / lnurl pay link. Using a pubkey, username and description (for the lnurl metadata).
Then the lnurl server can create invoices on the user's behalf.

This PR contains the lnurl server, the client implementation has to be made as well. When building the client implementation I'll see if there is need to store the created invoice with preimage and transfer id or not. 

Using `diesel` for the database schema. The idea was that we would be able to swap sqlite for postgres easily. However, that is not the case unfortunately. At least it's not as easy as using some generics. There will have to be two implementations for the different databases, even though the code accessing them would be quite the same. I'm thinking it might be good to switch to `sqlx` instead, because that has async support. Currently the server only supports sqlite.

The server now uses the `verify_message` to verify the signed payload. Note that to sign you will not use the identity key, but a leaf pubkey. So you won't be able to use the equivalent `sign_message` for that.


crates/breez-sdk/lnurl/src/main.rs
```diff
@@ -0,0 +1,183 @@
...
+        .route("/lnurlpay/{pubkey}/recover", delete(LnurlServer::recover))
+        .route(
+            "/.well-known/lnurlp/{identifier}",
+            get(LnurlServer::handle_lnurl_pay),
+        )
+        .route("/lnurlp/{identifier}", get(LnurlServer::handle_lnurl_pay))
+        .route(
+            "/lnurlp/{identifier}/invoice",
+            get(LnurlServer::handle_invoice),
+        )
```
roeierez, line 151: I think we are missing the verify route: https://github.com/breez/breez-lnurl/blob/main/lnurl/pay.go#L475

crates/breez-sdk/lnurl/src/main.rs
```diff
@@ -0,0 +1,183 @@
...
+    /// Scheme prefix for lnurl urls.
+    #[arg(long, default_value = "https")]
+    pub scheme: String,
+
+    /// Domain for lnurl urls.
+    #[arg(long, default_value = "127.0.0.1:8080")]
+    pub domain: String,
+
+    /// Minimum amount that can be sent in a lnurl payment.
+    #[arg(long, default_value = "1000")]
```
roeierez, line 68: I think we can lower here to 1 sat as default?

roeierez (commented):
Looks good!

crates/breez-sdk/lnurl/src/main.rs
JssDWt: Those are actually millisatoshi. The 4000000 one should be bumped.

crates/breez-sdk/lnurl/src/main.rs
```diff
@@ -0,0 +1,183 @@
...
+    } else if sqlite::has_migrations(&mut db_connection)? {
+        return Err(anyhow::anyhow!(
+            "database has pending migrations, run with --auto-migrate to apply them, or apply them manually"
+        ));
+    }
+
+    let wallet = Arc::new(
+        spark_wallet::SparkWallet::connect(
+            SparkWalletConfig::default_config(args.network),
+            DefaultSigner::new(&[0; 32], args.network)?,
```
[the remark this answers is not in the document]
roeierez: I just felt I should mention this in case someone has a strong opinion. I don't have.

crates/breez-sdk/lnurl/src/main.rs
roeierez: I wouldn't do that ATM.

crates/breez-sdk/lnurl/src/routes.rs
```diff
@@ -0,0 +1,405 @@
...
+            .map_err(|e| {
+                error!("failed to create lightning invoice: {}", e);
+                lnurl_error("failed to create invoice")
+            })?;
+
+        debug!("Created lightning invoice: {:?}", invoice);
+
+        // TODO: Save things like the invoice/preimage/transfer id?
+        // TODO: Validate invoice?
+        // TODO: Add lnurl-verify
```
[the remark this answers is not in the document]
roeierez: Should we exclude verify from this PR? I am totally fine with that. If so let's just open an issue for that and for the lightning address support so we can remove these requirements from the review.

crates/breez-sdk/lnurl/src/routes.rs
```diff
@@ -0,0 +1,348 @@
...
+                error!("failed to execute query: {}", e);
+                (
+                    StatusCode::INTERNAL_SERVER_ERROR,
+                    Json(Value::String("internal server error".into())),
+                )
+            })?;
+
+        match user {
+            Some(user) => Ok(Json(RecoverLnurlPayResponse {
+                lnurl: format!("{}://{}/lnurlp/{}", state.scheme, state.domain, user.name),
```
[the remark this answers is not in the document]
JssDWt: Done, changed to lnurlp:// prefix (LUD-17)

dangeross (approved):
LGTM. I think the API key is a good idea, as least for the `available` request

crates/breez-sdk/lnurl/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+        .route(
+            "/lnurlpay/available/{identifier}",
+            get(LnurlServer::<DB>::available),
+        )
+        .route("/lnurlpay/{pubkey}", post(LnurlServer::<DB>::register))
+        .route("/lnurlpay/{pubkey}", delete(LnurlServer::<DB>::unregister))
+        .route(
+            "/lnurlpay/{pubkey}/recover",
+            delete(LnurlServer::<DB>::recover),
+        )
```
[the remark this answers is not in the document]
JssDWt: Done

crates/breez-sdk/lnurl/src/auth.rs
```diff
@@ -0,0 +1,71 @@
...
+    next: Next,
+) -> Result<Response, StatusCode>
+where
+    DB: Send + Sync + 'static,
+{
+    let Some(ca_cert) = state.ca_cert.as_ref() else {
+        return Ok(next.run(req).await);
+    };
+
+    let Ok((_, ca_cert)) = X509Certificate::from_der(ca_cert.as_ref()) else {
```
dangeross, line 25: Is is possible to store the X509Certificate in the state? Then we don't need to convert it for each request

roeierez (approved):
LGTM! Very nice!

Issue #206, opened 2025-09-05
Abstracting TreeState


roeierez:
Currently we are:
1. Managing a local tree state which doesn't scale with multiple workers/processes that shares the same state
2. Having a TreeService that is responsible for:
     1. Ensuring leaves are "fresh" (timelock wise)
     2. Selecting leaves from the state for external operations
     3. But has some unrelated logic (collect_nodes, transfer_leaves_to_self)

The main requirement here is to allow users pass their own implementation for:
1.  TreeState (such as Redis based) -  to better support accessing spark account from different clients/processes
2. TreeService - to allow total control over the leaf selection process

TODO:

- [ ] - Cleanup TreeService - remove collect_leaves, move transfer_leaves_to_self to the transfer service
- [ ] - Remove the mutex (wraps the TreeState) from the TreeService which implies that the TreeState is thread safe and can be accesses concurrently
- [ ] - Extract TreeState to a trait.
- [ ]  Extract TreeService to a trait.
- [ ] - Add support for passing user implementation for TreeState and TreeService

Issue #145, opened 2025-08-21
Race between claiming transfer in claim_deposit and background event


JssDWt:
There are two places where the transfers resulting from a deposit claim can be claimed.

Here at the end of the claim_deposit function:
https://github.com/breez/spark-sdk/blob/12e76100c2b6d707cf3c3daa6f82a21870f17812/crates/spark-wallet/src/wallet.rs#L336

And here in the background processing of server events:
https://github.com/breez/spark-sdk/blob/12e76100c2b6d707cf3c3daa6f82a21870f17812/crates/spark-wallet/src/wallet.rs#L943

If the server event is first, the call to claim_deposit can fail with `transfer 0198c9ca-f5ab-7cf2-b8bf-127d8ef26bd1 has already been claimed`

I have added the TransferAlreadyClaimed ServiceError for this purpose, but looks like it's not catching the error there or something changed.

PR #208, opened 2025-09-06
Tree service cleanup

roeierez:
Starting to address #206 
This PR cleanup the TreeService:
1. Move `collect_leaves` to the deposit service where it is only used. By that also makes the deposit service covers all deposit flow.
2. Move the `transfer_leaves_to_self` to the transfer service (now uses internally `transfer_leaves_to`)
3. Better encapsulate lightning payments inside lightning service (remove friction from the wallet).

Everything in Spark ultimately maps to a Transfer, which suggests that the TransferService operates at a lower level than higher-level services like Deposit or Lightning. The idea is for these higher-level services to use the TransferService internally, allowing them to encapsulate the entire flow rather than use the wallet as facade that adds friction. 
This way, SparkWallet only needs to handle the final results (ideally reserving leaves only, responding to events, etc...) and can delegate the detailed processes to the services themselves.

Wishful thinking: I believe the flakiness issue with the deposit integration tests might also be resolved.

### DIFF
--- a/crates/spark/src/services/deposit.rs
+++ b/crates/spark/src/services/deposit.rs
@@ -205,7 +205,7 @@ impl<S: Signer> DepositService<S> {
                 vout,
             )
             .await?;
-        Ok(nodes)
+        self.collect_leaves(nodes).await
     }
 
     pub async fn collect_leaves(
--- a/crates/spark/src/services/deposit.rs
+++ b/crates/spark/src/services/deposit.rs
@@ -9,7 +9,7 @@ use bitcoin::{
     secp256k1::{Message, PublicKey, ecdsa::Signature, schnorr},
 };
 use serde::Serialize;
-use tracing::{error, trace};
+use tracing::{error, trace, warn};
 
 use crate::{
     Network,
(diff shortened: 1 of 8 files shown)


crates/spark/src/services/deposit.rs
```diff
@@ -198,7 +205,59 @@ impl<S: Signer> DepositService<S> {
...
+                            "Failed to transfer leaves to self: {e:?}"
+                        )))?;
+                    }
+                };
+
+                resulting_nodes.extend(transfer.into_iter());
+            }
+        }
+
+        Ok(resulting_nodes)
```
JssDWt, line 260: Looking at this function again, I think we can replace this with a call to `check_timelock_nodes` for the `Available` leaves. What do you think?

JssDWt (commented):
I think this looks good. 

I don't think it will fix the race described in #145 though (which causes the test flake).

Made one comment for potential improvement.

roeierez:
> I don't think it will fix the race described in #145 though (which causes the test flake).

This block I think may fix it as it adds the AlreadyClaim check also to the tweak keys function: https://github.com/breez/spark-sdk/pull/208/files#diff-f70af0ccb47e7a898b9daef16b6e70f2954142e070af8861c0ca533dd2f3473aR696-R703

crates/spark/src/services/deposit.rs
JssDWt: I'll have to dive into it more. I created an issue for it: #213

JssDWt (approved):
LGTM

dangeross (approved):
LGTM

PR #214, opened 2025-09-08
Modular tree service

roeierez:
This PR extracts the TreeService and TreeState (renamed to TreeStore) into two separate traits.
By doing so, Spark Wallet can now work with these abstractions, paving the way for users to provide their own custom store or service implementations when they need maximum control (note: passing custom implementations is not yet supported in this PR).

This design is particularly useful in service-oriented architectures, where multiple clients or workers need access to the same tree state. In such scenarios, the tree can be managed as a shared state service, accessible to all clients either synchronously or asynchronously.

Additionally, the selection logic has been moved into the store itself, rather than residing in the service. This shift enables different store implementations (e.g., Redis, SQL) to leverage their native querying capabilities for selection, instead of requiring all leaves to be loaded into memory.

Some static selection utils were extracted out to a selection_helper.rs file.


crates/spark/src/tree/select_helper.rs
```diff
@@ -0,0 +1,194 @@
...
+        }
+        None => None,
+    };
+
+    Ok(TargetLeaves::new(amount_leaves, fee_leaves))
+}
+
+/// Selects leaves from the tree that sum up to exactly the target amount.
+/// If such a combination of leaves does not exist, it returns `None`.
+pub(crate) fn select_leaves_by_amount(
```
dangeross, line 45: These `pub(crate)` helper functions could also potentially be reused outside of the crate?

dangeross (approved):
Nice. LGTM (with one comment)

danielgranhao (commented):
Looks good! Just one concern about static vs dynamic dispatch

roeierez:
I changed the TreeStore trait functions to return Result ( so errors can propagate).

PR #215, opened 2025-09-08
replace rusqlite with sqlx

JssDWt:
Sqlx is async, that's nice. Potentially we'll be able to reuse the implementation for both postgres and sqlite (wishful thinking?)

Migrations are now sql files. sqlx maintains a table with migrations in the database. 
If migrating from an existing rusqlite-created database to sqlx, no tables will change, because the initial migration has 'IF EXISTS' on every table creation. Subsequent migrations won't need this.

### DIFF
--- a/docs/breez-sdk/snippets/python/src/getting_started.py
+++ b/docs/breez-sdk/snippets/python/src/getting_started.py
@@ -46,7 +46,7 @@ async def init_sdk_advanced():
 
     try:
         # Create the default storage
-        storage = default_storage(data_dir="./.data")
+        storage = await default_storage(data_dir="./.data")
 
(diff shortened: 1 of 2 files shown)


docs/breez-sdk/snippets/flutter/lib/getting_started.dart
```diff
@@ -33,7 +33,7 @@ Future<void> initSdkAdvanced() async {
       .copyWith(apiKey: "<breez api key>");
 
   // Create the default storage
-  final storage = defaultStorage(dataDir: "./.data");
+  final storage = await defaultStorage(dataDir: "./.data");
```
dangeross, line 36: I think snippets for python and swift also need updating

PR #233, opened 2025-09-12
Derive Copy for various enums

benthecarman:
Small optimization and will remove some annoyances of having to clone


JssDWt (approved):
LGTM

PR #235, opened 2025-09-14
Use dynamic dispatched signer

roeierez:
We need the flexibility here in order to expose it to the breez sdk level where we will use either default signer or user passed signer. 
Also more consistent with other implementations we let the user to customize.

### DIFF
--- a/crates/breez-sdk/core/src/sdk.rs
+++ b/crates/breez-sdk/core/src/sdk.rs
@@ -66,7 +66,7 @@ enum SyncType {
 #[cfg_attr(feature = "uniffi", derive(uniffi::Object))]
 pub struct BreezSdk {
     config: Config,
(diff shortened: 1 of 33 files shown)


dangeross (commented):
Looks good. The Breez SDK signer interface will be added in a separate PR? I'm asking because I'm thinking about how to expose the Signer trait in uniffi. 

https://mozilla.github.io/uniffi-rs/latest/types/remote_ext_types.html

roeierez:
> Looks good. The Breez SDK signer interface will be added in a separate PR? I'm asking because I'm thinking about how to expose the Signer trait in uniffi.
> 
> https://mozilla.github.io/uniffi-rs/latest/types/remote_ext_types.html

Yes. I also didn't dive into it yet. I don't know if remote objects or creating the relevant simplified structure is the right approach

PR #236, opened 2025-09-14
Use session manager also for ssp

roeierez:
Both the SSP and the operator have an identifier public key, which allows us to use the same SessionManager interface to reuse sessions across both services.
This PR extends the session manager to all services that provide a public key.
depends on #235


crates/spark/src/ssp/graphql/client.rs
```diff
@@ -105,14 +114,10 @@ impl GraphQLClient {
     where
         T: Serialize + Clone + Into<Q::Variables>,
     {
-        if needs_auth && !self.auth_provider.is_authorized().await? {
-            self.authenticate().await?;
-            tracing::debug!("Authenticated successfully with ssp");
-        }
```
dangeross, line 111: Looking at the calls to `post_query()` it seems all calls pass `needs_auth: true`. So I would say we could remove it. Otherwise this change is getting the session even if we might not need it

dangeross (approved):
LGTM

PR #237, opened 2025-09-14
Reuse connections across instances

roeierez:
This PR allows passing external ConnectionManager so operator grpc connections can be reused in multiple instances.
This is particularly useful in a service environment where wallet instances are instantiated as per request in high frequency. 
In addition the background processing is now optional given the fact there are times only sending funds is needed and claiming transfers is executed elsewhere in its own service.


JssDWt (approved):
LGTM

dangeross (approved):
LGTM

PR #241, opened 2025-09-15
Wasm Node.js sqlite packaging improvements

### LINKED ISSUE #231: Drop requirement to add better-sqlite when using Node.js package
When using the Node.js package, if the default storage implementation is used, the integrator must currently add `better-sqlite` to their project. This isn't good UX, and we should fix it by always depending on better-sqlite, despite that forcing the dependency even when it isn't used.

danielgranhao:
Closes #231

It turns out we were already packaging `better-sqlite3` for Wasm packages. This PR just makes some small improvements:

- Sets the dependency as optional so that we don't break builds where for any reason better-sqlite3 can't be used 
- Upgrades the version
- Adapts the node.js example to use a compressed package so that it doesn't need to depend directly on better-sqlite3

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ crates/breez-sdk/cli/.data
 
 # Deployment files
 fly.toml
+
+# Wasm package
+packages/wasm/breeztech-breez-sdk-spark-v0.1.0.tgz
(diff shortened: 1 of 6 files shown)


dangeross (approved):
LGTM. I think it would be good to add a README to the Wasm package like: https://github.com/breez/breez-sdk-liquid/blob/main/packages/wasm/README.md

danielgranhao:
> I think it would be good to add a README to the Wasm package

Yes, created an issue for that: #243

PR #242, opened 2025-09-15
docs: set latest version on build

JssDWt:
Sets the tag used in install.md to the latest version, so the docs always have a more or less up-to-date version documented.


danielgranhao (approved):

docs/breez-sdk/src/guide/install.md
dangeross, line 1: Could also set the android/kotlin version
```
  implementation("breez_sdk_spark:bindings-android:<version>")
```

dangeross (approved):

PR #250, opened 2025-09-16
add a script to generate dockerignore from gitignore

JssDWt:
When deploying the lnurl server the build context is copied to the docker container. There was a lot of data transferred (gigabytes), but I'm not sure where the data came from. So I decided the easiest course of action is to simply take all the gitignores and put their content in the dockerignore.


JssDWt:
> Is it worth just additionally ignoring everything in the docs, examples & packages directories?

Maybe, but the remaining context is very small. So I don't really see the need at the moment. Now this script is independent of directory structure. I like that.

dangeross (approved):

PR #257, opened 2025-09-19
Remove strict `use_spark` check

dangeross:
Currently there is a burden on the implementer to only set `use_spark` to true in the send BOLT11 payment options if the invoice contains a Spark address. This PR replaces `use_spark` with `prefer_spark` and removes the error in case there is no Spark address present.

Update: [PR 244](https://github.com/breez/spark-sdk/pull/244) introduced using `prefer_spark_over_lightning` as the default, so this PR just renames the overriding option and softens the checking


JssDWt:
Is it an idea to set this in the config instead? Seems to me like this would be a hardcoded boolean in any app?

roeierez (approved):

PR #264, opened 2025-09-21
Fix events for lightning payments

roeierez:
This PR has several fixes for lightning payments:
1. Consistently use payment_id as the transfer id, previously it was sometimes the ssp id which contained some string prefix followed by a UUID that is different than the spark id. 
2. Emit PaymentSucceed only after perimage is retrieved.
3. Emit PaymentFailed in case it is relevant.

Update:
Added the various signer configurations (taproot, etc..) to WASM.

### DIFF
--- a/crates/breez-sdk/core/src/sdk.rs
+++ b/crates/breez-sdk/core/src/sdk.rs
@@ -17,8 +17,8 @@ use breez_sdk_common::{
     rest::RestClient,
 };
 use spark_wallet::{
-    ExitSpeed, InvoiceDescription, Order, PagingFilter, PayLightningInvoiceResult, SparkAddress,
-    SparkWallet, WalletEvent, WalletTransfer,
+    ExitSpeed, InvoiceDescription, Order, PagingFilter, SparkAddress, SparkWallet, WalletEvent,
+    WalletTransfer,
 };
 use std::{str::FromStr, sync::Arc};
 use tracing::{error, info, trace};
@@ -718,9 +718,8 @@ impl BreezSdk {
                 },
             )
             .await?;
-        self.event_emitter.emit(&SdkEvent::PaymentSucceeded {
-            payment: payment.clone(),
-        });
+
+        emit_optionaly_payment_complete(&self.event_emitter, payment.clone());
         Ok(LnurlPayResponse {
             payment,
             success_action,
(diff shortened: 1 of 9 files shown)


crates/breez-sdk/core/src/sdk.rs
```diff
@@ -935,14 +941,22 @@ impl BreezSdk {
...
-                          if let Err(e) = sync_trigger.send(SyncType::PaymentsOnly) {
-                              error!("Failed to send sync trigger: {e:?}");
-                          }
-                          return;
-                    }
+                    p = spark_wallet.fetch_lightning_send_payment(&ssp_id) => {
+                      if let Ok(Some(p)) = p && let Ok(payment) = Payment::from_lightning(p.clone(), payment.amount, payment.id.clone()) {
+                        info!("Pollling payment status = {} {:?}", payment.status, p.status);
+                        match payment.status {
+                          s if s != PaymentStatus::Pending => {
```
dangeross, line 948: Seems like this match block could be just an if statement?

crates/breez-sdk/core/src/sdk.rs
```diff
@@ -1218,6 +1232,14 @@ fn process_success_action(
     Ok(Some(SuccessActionProcessed::Aes { result }))
 }
 
+fn emit_optionaly_payment_complete(event_emitter: &EventEmitter, payment: Payment) {
```
dangeross, line 1235: The fn name is a bit misleading, maybe something like `emit_final_payment_status`

crates/breez-sdk/core/src/sdk.rs
```diff
@@ -935,14 +941,22 @@ impl BreezSdk {
...
-                      if let Ok(Some(p)) = p  && p.payment_preimage.is_some(){
-                          info!("Pollling payment preimage found");
-                          if let Err(e) = sync_trigger.send(SyncType::PaymentsOnly) {
-                              error!("Failed to send sync trigger: {e:?}");
-                          }
-                          return;
-                    }
+                    p = spark_wallet.fetch_lightning_send_payment(&ssp_id) => {
+                      if let Ok(Some(p)) = p && let Ok(payment) = Payment::from_lightning(p.clone(), payment.amount, payment.id.clone()) {
+                        info!("Pollling payment status = {} {:?}", payment.status, p.status);
```
dangeross, line 946:
```suggestion
                        info!("Polling payment status = {} {:?}", payment.status, p.status);
```

crates/breez-sdk/core/src/sdk.rs
```diff
@@ -935,14 +941,22 @@ impl BreezSdk {
...
+                        match payment.status {
+                          s if s != PaymentStatus::Pending => {
+                              info!("Pollling payment completed status = {}", s);
```
dangeross, line 949:
```suggestion
                              info!("Polling payment final status = {}", s);
```

crates/breez-sdk/core/src/sdk.rs
roeierez: Done

PR #265, opened 2025-09-22
Docs: Event flows

### LINKED ISSUE #262: Docs: Add payment status tables
Add the send/receive docs payment status tables describing the status and UX/UI recommendations

dangeross:
Fixes #262


roeierez (approved):
Looks good! Do we have anywhere where we recommend to refresh the balance and payments list on events such as (Payment events, Sync event)?

dangeross:
> Looks good! Do we have anywhere where we recommend to refresh the balance and payments list on events such as (Payment events, Sync event)?

No, but I can add it above the method specific tables

PR #273, opened 2025-09-25
optional ln address description with default

JssDWt:
Default is now 'Pay to {user}@{domain}'.

This default is set on the client side, in the sdk. Another option could be to set these default descriptions on the server side instead. That could be more flexible if we want to change this default in the future?


dangeross (approved):
LGTM

PR #276, opened 2025-09-25
include lnurl metadata on wasm

### LINKED ISSUE #272: Lnurl information is not populated
On WASM the lnurl pay info is not populated on the payment. Not sure if it is wasm specific issue.

JssDWt:
Fixes #272


roeierez (approved):
Nice catch!

PR #292, opened 2025-09-30
optional account number

JssDWt:
Allow setting the account number to support the same key format as Xverse / magic eden / SparkSat.


roeierez (approved):
LGTM

PR #293, opened 2025-10-02
async event listener

JssDWt:
Make the `EventListener` interface async, to allow running async code in it. I think this is an improvement in all languages.


roeierez (approved):
LGTM

PR #294, opened 2025-10-02
Option to wait for payment to complete

JssDWt:
This PR does the following:
- Flatten the database structure of the payment details (to allow querying by invoice)
- Added a `wait_for_payment` function to allow waiting for lightning payments, whether it's sending or receiving. Mostly useful for external users for waiting for receiving a specific payment.
- Added `return_pending_after_secs` to the `send_payment` function. By default the `send_payment` function will now wait for the payment to complete, rather than returning 'pending' immediately. This can be overridden with the `return_pending_after_secs` field. Setting to 0 will make the function return immediately.

### DIFF
--- a/crates/breez-sdk/core/src/persist/sqlite.rs
+++ b/crates/breez-sdk/core/src/persist/sqlite.rs
@@ -366,6 +366,43 @@ impl Storage for SqliteStorage {
         Ok(payment)
     }
 
+    async fn get_payment_by_invoice(
+        &self,
+        invoice: String,
+    ) -> Result<Option<Payment>, StorageError> {
+        let connection = self.get_connection()?;
+
+        let mut stmt = connection.prepare(
+            "SELECT p.id
+            ,       p.payment_type
+            ,       p.status
+            ,       p.amount
+            ,       p.fees
+            ,       p.timestamp
+            ,       p.method
+            ,       p.withdraw_tx_id
+            ,       p.deposit_tx_id
+            ,       p.spark
+            ,       l.invoice AS lightning_invoice
+            ,       l.payment_hash AS lightning_payment_hash
+            ,       l.destination_pubkey AS lightning_destination_pubkey
+            ,       COALESCE(l.description, pm.lnurl_description) AS lightning_description
+            ,       l.preimage AS lightning_preimage
+            ,       pm.lnurl_pay_info
+             FROM payments p
+             LEFT JOIN payment_details_lightning l ON p.id = l.payment_id
+             LEFT JOIN payment_metadata pm ON p.id = pm.payment_id
+             WHERE l.invoice = ?",
+        )?;
+
+        let payment = stmt.query_row(params![invoice], map_payment);
+        match payment {
+            Ok(payment) => Ok(Some(payment)),
+            Err(rusqlite::Error::QueryReturnedNoRows) => Ok(None),
+            Err(e) => Err(e.into()),
+        }
+    }
+
     async fn add_deposit(
         &self,
         txid: String,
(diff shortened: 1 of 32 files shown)


crates/breez-sdk/core/src/sdk.rs
```diff
@@ -1004,6 +950,123 @@ impl BreezSdk {
...
+            identifier: WaitForPaymentIdentifier::PaymentId(payment.id.clone()),
+        });
+        let payment = if let Some(secs) = return_pending_after_secs {
+            match timeout(Duration::from_secs(secs.into()), fut).await {
+                Ok(res) => res?.payment,
+                // On timeout return the pending payment.
+                Err(_) => payment,
+            }
+        } else {
+            fut.await?.payment
```
roeierez, line 1035: I think we should have our recommended default here. Seems to me 10 seconds to wait by default before returning the pending is reasonable. @kingonly what do you think?

packages/flutter/rust/src/models.rs
```diff
@@ -270,6 +270,7 @@ pub enum _SendPaymentOptions {
     },
     Bolt11Invoice {
         prefer_spark: bool,
+        return_pending_after_secs: Option<u32>,
```
roeierez, line 273: Perhaps we can make the naming more intuitive?
wait_for_completion_interval? or anything that focus on the completed payment rather than the pending.

crates/breez-sdk/core/src/persist/sqlite.rs
```diff
@@ -133,6 +134,35 @@ impl SqliteStorage {
...
+            ALTER TABLE payments ADD COLUMN spark INTEGER;
+            CREATE TABLE payment_details_lightning (
+              payment_id TEXT PRIMARY KEY,
+              invoice TEXT NOT NULL,
+              payment_hash TEXT NOT NULL,
+              destination_pubkey TEXT NOT NULL,
+              description TEXT,
+              preimage TEXT,
+              FOREIGN KEY (payment_id) REFERENCES payments(id) ON DELETE CASCADE
+            );
```
roeierez, line 149: Seems like only the lightning details are saved in another table. If this is because we only have one field for other kind of payments then I think it will change in the future and better to be consistent.
I am OK also with use one table with nullable fields for all kind of payments.

roeierez (changes requested):
I like it! Added a few comments.

crates/breez-sdk/core/src/persist/sqlite.rs
JssDWt: It is because there is only one field, yes. Ideally everything is just in the same table. The lightning details are extracted out into their own table, so that fields don't have to be nullable. This makes reasoning about the data structure easier to understand, because you never have to consider a null invoice but a non-null payment hash for example.

I think we should move the withdraw/deposit/spark to another table when it is needed from that nullability perspective. So if you add a field basically.

crates/breez-sdk/core/src/sdk.rs
JssDWt: I was thinking about that, and I think a 10 second timeout doesn't make sense. Developers may not run into the 10 second timeout while developing and will assume payments are complete when this function returns. It's a separate case to account for, but it's kind-of hidden. So I would opt for either returning a pending payment immediately, or waiting for the payment to complete, but not something in the middle.

packages/flutter/rust/src/models.rs
roeierez: 👍

crates/breez-sdk/core/src/sdk.rs
roeierez: Seems a bit overkill?
Why would some one want to wait for infinity? When sending payment I think one always want some timeout. If tracking is needed then using events is better.

roeierez (approved):
Tested on WASM web (lnurl, lightning, withdraw, deposit).
LGTM

PR #297, opened 2025-10-03
gracefully handle already claimed transfers

### LINKED ISSUE #282: Cleanup concurrent claim
Currently if two instances attempt to claim simultaneously one of them fails and the payment stays pending for some time.
Eventually it recovers but we want to make it claim gracefully.

JssDWt:
A common case is that a transfer has already been claimed by another instance of the sdk, or even the same instance, in another flow. Currently the claim_transfer function then errors. This can leave payments in a pending state until the next sync round.

In order to fix this, we'll gracefully handle already claimed incoming transfers. If there is an error from the operator, we'll check what the state of the transfer is on the operator, by querying the transfer and determining its state. If the state is already past the expected state, we'll return OK. The TransferAlreadyClaimed error is now handled inside claim_transfer, to minimize the amount of code paths that have to handle this error.

Fixes #282

### DIFF
--- a/crates/spark/src/services/transfer.rs
+++ b/crates/spark/src/services/transfer.rs
@@ -3,8 +3,9 @@ use std::{collections::HashMap, sync::Arc};
 
 use crate::Network;
 use crate::operator::OperatorPool;
-use crate::operator::rpc::spark::TransferFilter;
+use crate::operator::rpc::spark::query_nodes_request::Source;
 use crate::operator::rpc::spark::transfer_filter::Participant;
+use crate::operator::rpc::spark::{QueryNodesRequest, TransferFilter, TreeNodeIds};
 use crate::operator::rpc::{self as operator_rpc, OperatorRpcError};
 use crate::services::models::{LeafKeyTweak, Transfer, map_signing_nonce_commitments};
 use crate::services::{ProofMap, TransferId, TransferStatus};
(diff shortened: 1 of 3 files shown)


crates/spark/src/services/transfer.rs
```diff
@@ -675,35 +701,82 @@ impl TransferService {
...
+                    "Operator {} returned connection error: {:?}",
+                    operator.identity_public_key, status
+                );
+
+                // There was an error tweaking the keys. It's likely the transfer was not in state SenderKeyTweaked after all.
+                // Let's find out what state we're in.
+                // Fetch the leaf remotely from the operator. Determine its state, and return info about it to the caller.
+                let operator_transfers = operator
+                    .client
+                    .query_all_transfers(TransferFilter {
```
roeierez, line 746: Why don't we use the `self.query_transfer` function like we do in `claim_transfer_sign_refunds`?

roeierez (changes requested):
Looks good. Still need to test it.

crates/spark/src/services/transfer.rs
JssDWt: Because that queries the coordinator, while here we specifically want to know the state of the transfer on that coordinator.

roeierez: @JssDWt  I see makes sense. A general question about the approach: We refresh the transfer inside the more high level `claim_transfer` function. We have a retry mechanism there and we could check if the transfer has already claimed.  Just wondering if you considered that and if you see advantages on the current approach.

roeierez: Thanks Jesse!
I think you are right. If we start the process with tweaks only in some of the operators  (whether we exited in the middle of the process or another client  was able to twek some);we must query the leaves in this operator and skip if already tweaked.
Thanks for clarifying.

roeierez (approved):
LGTM

PR #303, opened 2025-10-07
Fix private methods exposed to uniffi

danielgranhao:
This PR fixes an issue that caused the latest published release (0.2.6) to be broken on its golang version.

We were exposing private methods to uniffi by mistake. In this latest release, that caused issues because methods like `send_bolt11_invoice` use `sdk_common` types (`Bolt11InvoiceDetails`) as params. This caused conflicts in the go bindings.

Additionally, I also fixed some small issues with the go snippets that I found while testing this fix.

To prevent this kind of issue in the future we should work on https://github.com/breez/spark-sdk/issues/302.

### DIFF
--- a/crates/breez-sdk/core/src/sdk.rs
+++ b/crates/breez-sdk/core/src/sdk.rs
@@ -913,6 +913,273 @@ impl BreezSdk {
         self.send_payment_internal(request, false).await
     }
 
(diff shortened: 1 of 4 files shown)


dangeross (approved):
Nice catch

JssDWt (approved):

PR #306, opened 2025-10-10
Replace refresh/extend with renew leaf flow

### LINKED ISSUE #202: Use Renew leaf new API
Renew leaf API to replace refresh and extend timelock

dangeross:
Replaces the refresh/extend timelock flow with renew node/refund/zero timelock flows.
- On each transfer a leaf's refund tx sequence decremented
- When the refund tx sequence is <= 100 it calls renew_refund, which decrements the node tx sequence and resets the refund tx sequence
- When the node tx sequence is <= 100 it calls renew_node, which splits off zero timelock node txs. From these split parent txs resets the node and refund tx sequences
- If a node tx sequence is already at 0, it can't be decremented further. In this case renew_zero_timeout create a new zero timelock node tx and resets the refund tx sequence

Closes #202

### DIFF
--- a/crates/spark/src/tree/service.rs
+++ b/crates/spark/src/tree/service.rs
@@ -258,6 +258,7 @@ impl TreeService for SynchronousTreeService {
             .collect::<Vec<_>>();
         let missing_operator_leaves = missing_operator_leaves_map
             .values()
+            .filter(|leaf_id| !ignored_leaves_map.contains_key(&leaf_id.id))
             .cloned()
             .collect::<Vec<_>>();
         let refreshed_leaves = self


roeierez (approved):
LGTM. I am looking to improve the test performance but I will do that out of the scope of this PR.

dangeross:
I still see an issue after renewing the node timelock, where there is a verifying pubkey mismatch. Transfers continue to work, but we get the following warning. I believe at this point we should transfer the leaf to self. It means the tree service having some sort of interface to the transfer service to perform this during refreshing the leaves (or the wallet doing this after the leaves refresh). What do you think @roeierez @danielgranhao ?
```
2025-10-18T08:50:07.348603Z  WARN spark::tree::service: Leaf 0199f67e-6245-7b42-a7ca-068b8e09e8ff's verifying public key does not match the expected value
```

roeierez:
> I still see an issue after renewing the node timelock, where there is a verifying pubkey mismatch. Transfers continue to work, but we get the following warning. I believe at this point we should transfer the leaf to self. It means the tree service having some sort of interface to the transfer service to perform this during refreshing the leaves (or the wallet doing this after the leaves refresh). What do you think @roeierez @danielgranhao ?
> 
> ```
> 2025-10-18T08:50:07.348603Z  WARN spark::tree::service: Leaf 0199f67e-6245-7b42-a7ca-068b8e09e8ff's verifying public key does not match the expected value
> ```

Is your proposed step (claiming to self) implemented in the JS SDK and we miss it? Or the  JS SDK has the same behavior (and error)?

roeierez:
> > I still see an issue after renewing the node timelock, where there is a verifying pubkey mismatch. Transfers continue to work, but we get the following warning. I believe at this point we should transfer the leaf to self. It means the tree service having some sort of interface to the transfer service to perform this during refreshing the leaves (or the wallet doing this after the leaves refresh). What do you think @roeierez @danielgranhao ?
> > ```
> > 2025-10-18T08:50:07.348603Z  WARN spark::tree::service: Leaf 0199f67e-6245-7b42-a7ca-068b8e09e8ff's verifying public key does not match the expected value
> > ```
> 
> Is your proposed step (claiming to self) implemented in the JS SDK and we miss it? Or the JS SDK has the same behavior (and error)?

@dangeross one thing I see different in the JS SDK is that there the verify step only runs on AVAILABLE leaves while we run on all leaves: https://github.com/buildonspark/spark/blob/main/sdks/js/packages/spark-sdk/src/spark-wallet/spark-wallet.ts#L510
Could it be that you are getting this error on an UNAVAILABLE leaf?

roeierez:
@dangeross also all this logic seems to be missing on our side: https://github.com/buildonspark/spark/blob/main/sdks/js/packages/spark-sdk/src/spark-wallet/spark-wallet.ts#L510-L545

I also don't see transfer leaves to self on the js sdk after renewing timelock.

crates/spark/src/tree/service.rs
```diff
@@ -238,16 +245,18 @@ impl TreeService for SynchronousTreeService {
...
             .into_iter()
-            .filter(|leaf| !missing_operator_leaves.contains_key(&leaf.id))
+            .filter(|leaf| {
+                !missing_operator_leaves_map.contains_key(&leaf.id)
+                    && !ignored_leaves_map.contains_key(&leaf.id)
+            })
             .collect::<Vec<_>>();
-
-        let ignored_leaves = missing_operator_leaves
+        let missing_operator_leaves = missing_operator_leaves_map
```
danielgranhao: Should we also filter out ignored leaves here?

danielgranhao (approved):

PR #313, opened 2025-10-13
allow nostr zaps

### LINKED ISSUE #308: [Feature Request] Add NIP-57 Zap Support to Lightning Addresses
**Add NIP-57 (Nostr zaps) support to Breez Spark SDK Lightning addresses to enable users to receive zaps through their @breez.tips addresses.**

Breez Lightning addresses work for regular payments but cannot receive Nostr zaps because the LNURL endpoint is missing required NIP-57 fields (allowsNostr: true and nostrPubkey). 

These fields are mandatory according to the NIP-57 spec:
https://github.com/nostr-protocol/nips/blob/master/57.md

Without this, compliant wallets reject zap attempts with 'invalid lightning address' errors.

Thanks for investigating this, I look forward to being able to integrate this feature.

JssDWt:
Allow receiving nostr zaps.
- a nostr_pubkey can be configured in the config, together with the lnurl_domain. If set, nostr is supported (given the client supports responding to zap requests).
- invoices for nostr requests are stored on the lnurl server together with the zap request.
- the client synchronizes the invoices on sync
- the nostr zap request is included in the payment details

- There is now also has sender comment support (same mechanism)

NOTE: When the client receives PaymentSucceeded event, the nostr information + sender comment will not yet have been synchronized.

State: kind of tested locally a little bit. Need to test properly with a nostr client. The server side is not yet deployed. We can deploy after an initial review round.

Fixes #308

### DIFF
--- a/crates/breez-sdk/lnurl/src/postgresql/repository.rs
+++ b/crates/breez-sdk/lnurl/src/postgresql/repository.rs
@@ -43,13 +43,16 @@ impl crate::repository::LnurlRepository for LnurlRepository {
         .bind(name)
         .fetch_optional(&self.pool)
         .await?
-        .map(|row| User {
-            domain: domain.to_string(),
-            pubkey: row.get(0),
-            name: row.get(1),
-            description: row.get(2),
-            nostr_pubkey: row.get(3),
-        });
+        .map(|row| {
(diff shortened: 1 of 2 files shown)


JssDWt:
Closing because #317 was merged. Will reopen with just the client-side things.

PR #315, opened 2025-10-14
breez-sdk: Expose `get_tokens_metadata`

danielgranhao:
This exposes a new method to allow getting tokens metadata given their identifier.

TODO: 

- [x] update docs

### DIFF
--- a/docs/breez-sdk/snippets/flutter/lib/tokens.dart
+++ b/docs/breez-sdk/snippets/flutter/lib/tokens.dart
@@ -20,7 +20,11 @@ Future<void> fetchTokenBalances(BreezSdk sdk) async {
 
 Future<void> fetchTokenMetadata(BreezSdk sdk) async {
   // ANCHOR: fetch-token-metadata
-  final response = await sdk.getTokensMetadata(request: GetTokensMetadataRequest(tokenIdentifiers: ['<token identifier 1>', '<token identifier 2>']));
+  final response = await sdk.getTokensMetadata(
+    request: GetTokensMetadataRequest(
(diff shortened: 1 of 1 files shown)


crates/breez-sdk/core/src/sdk.rs
```diff
@@ -1179,6 +1179,31 @@ impl BreezSdk {
...
         })
     }
+
+    /// Returns the metadata for the given token identifiers.
+    ///
+    /// Results are not guaranteed to be in the same order as the input token identifiers.    
+    ///
+    /// If the metadata is not found locally in cache, it will be queried from
+    /// the Spark network and then cached.
+    pub async fn get_tokens_metadata(
```
dangeross: I feel like it should be `list_tokens_metadata`, so it fits with the listing payments, unclaimed desposits and fiat currencies. What do you think?

docs/breez-sdk/snippets/flutter/lib/tokens.dart
```diff
@@ -18,6 +18,23 @@ Future<void> fetchTokenBalances(BreezSdk sdk) async {
   // ANCHOR_END: fetch-token-balances
 }
 
+Future<void> fetchTokenMetadata(BreezSdk sdk) async {
+  // ANCHOR: fetch-token-metadata
+  final response = await sdk.getTokensMetadata(request: GetTokensMetadataRequest(tokenIdentifiers: ['<token identifier 1>', '<token identifier 2>']));
```
dangeross: Can you break it into multiple lines so its more readable

dangeross (approved):
LGTM

PR #319, opened 2025-10-15
spark: Only use available leaves when computing balance

### LINKED ISSUE #316: GetInfoResponse shows outdated balanceSats after L1 withdrawal, doesn’t reflect actual balance
When calling GetInfoResponse in the Spark Swift SDK, the balanceSats field returns a value that appears correct based on the last Spark balance, but it does not reflect the actual current balance, especially after a withdrawal to L1.

For example:
	•	I withdrew funds from Spark to a Bitcoin (L1) address.
	•	The BTC transaction is confirmed and visible on-chain.
	•	However, even after syncing, the balanceSats still shows the pre-withdrawal Spark balance.
	•	On Spark Scan, it shows the same balance as GetInfoResponse, but also shows a “hard balance” which is accurate.

Expected Behavior:
GetInfoResponse.balanceSats should reflect the actual available balance, ideally matching the “hard balance” shown on Spark Scan.

Actual Behavior:
The response seems to cache or reuse the previous balance and doesn’t update after a withdrawal, leading to an incorrect representation of available funds.

Steps to Reproduce:
	1.	Withdraw funds from Spark to an external L1 BTC address.
	2.	Confirm transaction on chain.
	3.	Call GetInfoResponse.
	4.	Observe that balanceSats is not updated, even after a sync.

danielgranhao:
Closes #316

The issue occurred because we were considering unavailable leaves (in this case, `TransferLocked`) when calculating the balance. We missed this because we didn't filter for `Available` in the `missing_from_operators` set of leaves.

### DIFF
--- a/crates/spark/src/tree/mod.rs
+++ b/crates/spark/src/tree/mod.rs
@@ -20,7 +20,6 @@ pub struct Leaves {
     pub available: Vec<TreeNode>,
     pub not_available: Vec<TreeNode>,
     pub available_missing_from_operators: Vec<TreeNode>,
-    pub not_available_missing_from_operators: Vec<TreeNode>,
     pub reserved: Vec<TreeNode>,
 }
 
--- a/crates/spark/src/tree/store.rs
+++ b/crates/spark/src/tree/store.rs
@@ -53,12 +53,6 @@ impl TreeStore for InMemoryTreeStore {
                 .filter(|leaf| leaf.status == TreeNodeStatus::Available)
                 .cloned()
                 .collect(),
-            not_available_missing_from_operators: leaves
-                .missing_operators_leaves
-                .values()
-                .filter(|leaf| leaf.status != TreeNodeStatus::Available)
-                .cloned()
-                .collect(),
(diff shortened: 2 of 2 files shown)


crates/spark/src/tree/mod.rs
```diff
@@ -19,7 +19,8 @@ use uuid::Uuid;
 pub struct Leaves {
     pub available: Vec<TreeNode>,
     pub not_available: Vec<TreeNode>,
-    pub missing_from_operators: Vec<TreeNode>,
+    pub available_missing_from_operators: Vec<TreeNode>,
+    pub not_available_missing_from_operators: Vec<TreeNode>,
```
roeierez: Perhaps we can omit this field? We don't use it and I know spark are on their way to fix the "missing from operators" issue soon.

danielgranhao: Done in [8d32e1d](https://github.com/breez/spark-sdk/pull/319/commits/8d32e1dfb84e308ca79a705141d097fc5221367d)

PR #334, opened 2025-10-18
Real-time sync

JssDWt:
This PR adds real-time sync functionality.
Only the payment metadata is synchronized currently.

I did my best to create a proper separation of concerns. The sync is completely standalone, outside the domain of the relational data model. I wanted to put all that code in the `common` crate, but I didn't want to create a new persister trait for this. So some code is in the `core` package because of that.

The only part of the code that knows about sync is the wrapper repository around our existing repository. It calls the sync service to set a record _before_ (that is important) inserting in the relational data model. It also listens to incoming changes through a callback mechanism. This wrapper repository would be the only thing needed in the core package if persistence was in common.

The sync basically works like this:
Outgoing changes:
- first insert into outgoing_sync, then in relational data model. If we would do that the other way around, you would not be able to track the change anymore.
- In the background these changes are pushed to remote.
- After pushing they are removed from the outgoing table in the same transaction as they are inserted in the sync state.

Incoming changes:
- First they are inserted into an incoming table
- Then the pending outgoing changes are 'rebased'. Currently they only need a new revision number. In the future 'rebase' could mean more complex logic.
- Then the incoming change is inserted in the sync state. Note the incoming change is not deleted yet, so we know there is a pending change.
- Then the callback is invoked to update the relational data model. 
- Only then is the incoming change deleted.

These sequences of events ensure consistency at all times. Just the order is important.

TODO: 
- [x] Record encryption
- [x] real-time sync related events
- [x] Fix backoff storm
- [x] Tests
- [x] Docs

### DIFF
--- a/crates/breez-sdk/core/src/sdk.rs
+++ b/crates/breez-sdk/core/src/sdk.rs
@@ -280,8 +280,6 @@ impl BreezSdk {
     /// 1. `periodic_sync`: the wallet with the Spark network    
     /// 2. `monitor_deposits`: monitors for new deposits
     fn start(&self, initial_synced_sender: watch::Sender<bool>) {
-        // TODO: update the local database from outgoing sync records.
-        // ensure there will be no races. (if we would just take the pending outgoing records, it's possible the remote update happens before the local update. Then we have an issue at startup, because the update would no longer be 'pending', while the local data model hasn't been updated yet.)
         self.periodic_sync(initial_synced_sender);
         self.try_recover_lightning_address();
     }


roeierez:
@JssDWt thanks for that! Indeed a lot of thinking and the separation of concerns is just great! 
I have noticed there are some concepts that are different from the liquid implementation which I am trying to understand and evaluate the tradeoffs so I will start by some high level questions here if that's OK.

* It seems that we now save every local change as a pending outgoing change in the db. Given the fact that we merge these changes eventually (before pushing) what is the advantage of persisting them as separate changes and not merge as soon as we have a new local change with some pending outgoing change? I hope this is clear.

crates/breez-sdk/common/src/utils/mod.rs
```diff
@@ -1,3 +1,5 @@
+use std::time::SystemTime;
```
dangeross: Should probably use web_time::SystemTime for Wasm

crates/breez-sdk/core/src/persist/sqlite.rs
```diff
@@ -62,7 +66,7 @@ impl SqliteStorage {
         };
         migrations.to_latest(&mut conn)?;
 
-        if previous_version < 6 {
+        if previous_version < 7 {
```
dangeross: Isn't this fine to stay as 6?

crates/breez-sdk/core/src/sdk.rs
```diff
@@ -174,7 +177,8 @@ pub async fn connect(request: crate::ConnectRequest) -> Result<BreezSdk, SdkErro
         .join(path_suffix);
 
     let storage = default_storage(storage_dir.to_string_lossy().to_string())?;
-    let builder = crate::SdkBuilder::new(request.config, request.seed, storage);
+    let mut builder = super::sdk_builder::SdkBuilder::new(request.config, request.seed, storage);
+    builder = builder.with_real_time_sync(BREEZ_SYNC_SERVICE_URL.to_string());
```
dangeross: I'm wondering if we need the option to run without the sync like in the Liquid SDK? The Breez itests run on regtest without an API key, so it would cause errors authenticating with the default sync service url. Could the sync service url be an optional config field and the sync service be initialized in the SDK instead of the builder?

crates/breez-sdk/core/src/persist/sqlite.rs
JssDWt: Yes, I got confused about the numbers. This change was not necessary.

crates/breez-sdk/core/src/sdk.rs
JssDWt: I made the url an optional field. Initializing in the SDK instead of the builder became a little more problematic now, because this function now also takes sync storage as a parameter.

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -243,17 +257,59 @@ impl SdkBuilder {
...
+
+            synced_storage.listen(incoming_callback_receiver, outgoing_callback_receiver);
+            let storage: Arc<dyn Storage> = synced_storage;
+            let sync_client = BreezSyncerClient::new(server_url, self.config.api_key.as_deref())
+                .map_err(|e| SdkError::Generic(e.to_string()))?;
+            let sync_signer = DefaultSyncSigner::new(
+                Arc::clone(&signer),
+                "m/448201320'/0'/0'/0/0".parse().map_err(|_| {
+                    SdkError::Generic(
+                        "Someone put an invalid static derivation path here".to_string(),
```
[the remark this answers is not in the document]
JssDWt: Done

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -265,17 +285,68 @@ impl SdkBuilder {
...
+            );
+            let signing_sync_client = SigningClient::new(
+                Arc::new(sync_client),
+                Arc::new(sync_signer),
+                Uuid::now_v7().to_string(),
+            );
+            let sync_processor = Arc::new(SyncProcessor::new(
+                signing_sync_client,
+                sync_service.get_sync_trigger(),
+                incoming_callback_sender,
```
roeierez: It seems to me that passing a trait instead of callbacks can simplify here?

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -265,17 +285,68 @@ impl SdkBuilder {
         };
         let shutdown_sender = watch::channel::<()>(()).0;
 
+        let (storage, sync_processor) = if let Some(server_url) = &self.real_time_sync_server_url {
+            debug!("Real-time sync is enabled.");
```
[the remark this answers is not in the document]
JssDWt: Yes, done

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -265,16 +284,38 @@ impl SdkBuilder {
...
 
+        let event_emitter = Arc::new(EventEmitter::new());
+        let storage = if let Some(server_url) = &self.real_time_sync_server_url {
+            let Some(sync_storage) = self.sync_storage else {
+                return Err(SdkError::Generic("Sync storage is not available".into()));
+            };
+
+            init_and_start_real_time_sync(RealTimeSyncParams {
+                server_url: server_url.clone(),
+                api_key: self.config.api_key.clone(),
```
dangeross: Just want to check what happens when no API key is provided (e.g the Breez itests run in regtest with no API key), does it gracefully handle failed calls to the data sync server?

crates/breez-sdk/core/src/sdk.rs
JssDWt: Can you check whether the current state has what you want here @dangeross? Currently on `connect`, if the url is `None`, I initialize it with the default. Should I do that in `default_config` instead?

dangeross: I think ideally we want:
- Real-time sync enabled by default (including using the basic `connect()` initialization)
- Can be opted-out by setting the config param to None
- Can be customized with a different storage implementation using the SDK builder

What do you think @roeierez?

crates/breez-sdk/common/src/sync/background.rs
```diff
@@ -0,0 +1,1365 @@
...
+                _ = pull_trigger.changed() => {
+                    debug!("Received incoming sync notification");
+                    match self.pull_sync_once().await {
+                        Ok(count) => (count, None),
+                        Err(e) => {
+                            error!("Failed to sync once: {}", e);
+                            (None, None)
+                        }
+                    }
+                }
```
danielgranhao: Awaiting futures within `select` arms prevents those futures from being canceled by the shutdown receiver. In the past, we saw issues with that on Wasm Node.js (resulting in memory leaks). A simple way to avoid it is to put all logic in the async expression and avoid it in the handler.

This also applies below

crates/breez-sdk/common/src/sync/model.rs
```diff
@@ -0,0 +1,506 @@
...
+
+impl Display for RecordId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{:x}",
+            sha256::Hash::hash(format!("{}:{}", self.r#type, self.data_id).as_bytes())
+        )
+    }
+}
```
danielgranhao: Nit: Should we implement a `hash` method? IMO, we should leave Display for readable formats

crates/spark/src/signer/mod.rs
```diff
@@ -20,6 +21,12 @@ pub trait Signer: Send + Sync + 'static {
         message: &[u8],
     ) -> Result<Signature, SignerError>;
 
+    fn sign_message_ecdsa_recoverable_from_path(
+        &self,
+        message: &[u8],
+        derivation_path: &DerivationPath,
+    ) -> Result<RecoverableSignature, SignerError>;
+
```
danielgranhao: Wouldn't it make sense to create a new signer trait on the Breez SDK level that implements the Spark signer methods and these new ones? Then spark crate users wouldn't need to implement these and Breez SDK users would still have a single trait to implement.

JssDWt: Yes, that makes sense to have a signer trait on the Breez SDK level that contains the additional functions. Then create an implementation of the spark signer in the Breez SDK that pipes everything to the SDK signer. It's definitely the cleaner approach. I'll do that. Agreed @roeierez?

crates/breez-sdk/common/src/sync/model.rs
```diff
@@ -0,0 +1,506 @@
+use std::{collections::HashMap, fmt::Display};
+
+use bitcoin::hashes::{Hash, sha256};
+use semver::Version;
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+const CURRENT_SCHEMA_VERSION: Version = Version::new(0, 2, 6);
```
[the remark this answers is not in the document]
danielgranhao: Ahh 🙂 

Seems to me that the sync schema may evolve differently from the SDK's API, so I think a separate version, like we have in the liquid SDK, makes sense.

crates/breez-sdk/common/src/sync/background.rs
danielgranhao: Let's leave as is. We can fix it in another PR. There are likely other places in the codebase where similar issues are prone and we can apply a common solution like we did in the liquid sdk: https://github.com/breez/breez-sdk-liquid/blob/973fd2327a77ef1bceaec4502aeace8e053ec358/lib/core/src/utils/mod.rs#L196

crates/breez-sdk/common/src/sync/model.rs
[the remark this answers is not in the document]
JssDWt: Made it 1.0.0

danielgranhao (approved):
LGTM

roeierez (approved):
LGTM

PR #336, opened 2025-10-19
improve breez-itest performance

roeierez:
Some fixes to make the long timelock tests run in 24 minutes. In total 400 payments.
Total time ~24 minutes (3.7 seconds in average per payment).
Every payment time split into:
1. ~1.5 seconds sending
2. ~2 seconds claiming and updating balance

Changes:
1. Remove explicit sync on every iteration (balances should be updated as part of the payment/claiming).
2. Add one extra wait_for_payment for Alice (for the initial deposit) otherwise it is receives mistakenly as the spark payment later.
3. Reduce poll interval to 100 milliseconds.

Let's see how it performs on the CI.


dangeross (approved):
Nice

danielgranhao (approved):

PR #338, opened 2025-10-20
remove error log on startup

JssDWt:
This would log for every operator:

```
ERROR spark::operator::rpc::spark_rpc_client: 670: Failed to get
operator session from session manager: Generic error: Session not found
```

### DIFF
--- a/crates/spark/src/ssp/graphql/client.rs
+++ b/crates/spark/src/ssp/graphql/client.rs
@@ -6,7 +6,7 @@ use reqwest::header::HeaderMap;
 use reqwest::header::{AUTHORIZATION, CONTENT_TYPE, HeaderValue};
 use serde::Serialize;
 use std::sync::Arc;
-use tracing::error;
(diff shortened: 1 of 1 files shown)


crates/spark/src/operator/rpc/spark_rpc_client.rs
```diff
@@ -667,7 +667,14 @@ impl SparkRpcClient {
         let valid_session = match current_session {
             Ok(session) => self.auth.get_authenticated_session(Some(session)).await,
             Err(e) => {
-                error!("Failed to get operator session from session manager: {}", e);
+                match e {
```
dangeross: could you add the same to the graphql client?
https://github.com/breez/spark-sdk/blob/62eb94e7f5e85f1420ff91784ff621a3c8aba99c/crates/spark/src/ssp/graphql/client.rs#L500-L503

roeierez (approved):

PR #349, opened 2025-10-22
Spark invoices support

### LINKED ISSUE #212: Spark invoice support
Spark addresses can be used to invoice sat or token amounts. When this is used, they should also be signed (or have their signature verified).

Worth noting:
- #199 already introduces the use of the invoice data from the payer side, but doesn't include signature validation.
- At the time of writing, invoicing sat amounts is still under development

danielgranhao:
Closes #212

Changes:

- Breez SDK
    - spark invoices can now be created and paid
    - Separate input types returned by `parse` for spark addresses and invoices
    - Spark and Token payment details include the invoice the payment fulfills
    - `wait_for_payment` can also be called with a spark invoice as the payment request
- Spark/Spark wallet
    - Query invoice status implemented to match js sdk. Can be tried using the internal cli
    - Updated token identifier representation at `spark`'s `SparkAddress` to bech32m instead of keeping raw byte array
    - Implemented invoice signing as part of the string serialization (`to_invoice_string()`) and signature verification as part of `from_str()`.
    - Implemented token transactions v2 (needed to include the invoice attachments field)
    - Fixed spark address protobuf encoding (SO expects a canonical encoding)

### DIFF
--- a/crates/breez-sdk/core/src/persist/sqlite.rs
+++ b/crates/breez-sdk/core/src/persist/sqlite.rs
@@ -188,8 +188,8 @@ impl SqliteStorage {
             "DROP TABLE payments;",
             "ALTER TABLE payments_new RENAME TO payments;",
             "CREATE TABLE payment_details_spark (
-              payment_id TEXT PRIMARY KEY,
-              invoice_details TEXT,
+              payment_id TEXT NOT NULL PRIMARY KEY,
+              invoice_details TEXT NOT NULL,
               FOREIGN KEY (payment_id) REFERENCES payments(id) ON DELETE CASCADE
             );
             ALTER TABLE payment_details_token ADD COLUMN invoice_details TEXT;",
@@ -373,9 +373,11 @@ impl Storage for SqliteStorage {
                     "UPDATE payments SET spark = 1 WHERE id = ?",
                     params![payment.id],
                 )?;
-                tx.execute("INSERT OR REPLACE INTO payment_details_spark (payment_id, invoice_details) VALUES (?, ?)",
-                    params![payment.id, invoice_details.map(|d| serde_json::to_string(&d)).transpose()?],
-                )?;
+                if let Some(invoice_details) = invoice_details {
+                    tx.execute("INSERT OR REPLACE INTO payment_details_spark (payment_id, invoice_details) VALUES (?, ?)",
+                        params![payment.id, serde_json::to_string(&invoice_details)?],
+                    )?;
+                }
             }
             Some(PaymentDetails::Token {
                 metadata,
--- a/crates/breez-sdk/wasm/js/node-storage/index.cjs
+++ b/crates/breez-sdk/wasm/js/node-storage/index.cjs
@@ -265,7 +265,10 @@ class SqliteStorage {
           spark: payment.details?.type === "spark" ? 1 : null,
         });
 
-        if (payment.details?.type === "spark") {
+        if (
+          payment.details?.type === "spark" &&
+          payment.details.invoiceDetails != null
+        ) {
           sparkInsert.run({
             id: payment.id,
             invoiceDetails: payment.details.invoiceDetails
--- a/crates/breez-sdk/wasm/js/node-storage/migrations.cjs
+++ b/crates/breez-sdk/wasm/js/node-storage/migrations.cjs
@@ -215,8 +215,8 @@ class MigrationManager {
         name: "Add spark invoice details",
         sql: [
           `CREATE TABLE payment_details_spark (
-              payment_id TEXT PRIMARY KEY,
-              invoice_details TEXT,
+              payment_id TEXT NOT NULL PRIMARY KEY,
+              invoice_details TEXT NOT NULL,
               FOREIGN KEY (payment_id) REFERENCES payments(id) ON DELETE CASCADE
             )`,
           `ALTER TABLE payment_details_token ADD COLUMN invoice_details TEXT`,


crates/breez-sdk/core/src/sdk.rs
```diff
@@ -606,11 +610,40 @@ impl BreezSdk {
...
+                description,
+                sender_public_key,
+            } => {
+                let invoice = self.spark_wallet.create_spark_invoice(
+                    amount,
+                    token_identifier,
+                    expiry_time.map(|time| {
+                        SystemTime::UNIX_EPOCH
+                            .checked_add(Duration::from_secs(time))
+                            .expect("Invalid expiry time")
```
dangeross: `expect` can potentially panic, better to handle gracefully

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -484,6 +504,16 @@ pub enum SendPaymentMethod {
         /// If empty, it is a Bitcoin payment
         token_identifier: Option<String>,
     },
+    SparkInvoice {
+        invoice: String,
```
dangeross: Do we need the `invoice` here if we have the `spark_invoice_details`?

crates/breez-sdk/core/src/persist/sqlite.rs
```diff
@@ -187,6 +187,12 @@ impl SqliteStorage {
              FROM payments;",
             "DROP TABLE payments;",
             "ALTER TABLE payments_new RENAME TO payments;",
+            "CREATE TABLE payment_details_spark (
+              payment_id TEXT PRIMARY KEY,
+              invoice_details TEXT,
```
dangeross: Do we need to store the whole invoice details object? vs the essentials (description and invoice string)

crates/breez-sdk/core/src/sdk.rs
```diff
@@ -1286,6 +1362,73 @@ impl BreezSdk {
...
+            .await?;
+
+        self.map_and_persist_token_transaction(&token_transaction)
+            .await
+    }
+
+    async fn send_spark_invoice(
+        &self,
+        invoice: &str,
+        _token_identifier: Option<String>,
```
dangeross: Can be removed?

crates/spark-wallet/src/wallet.rs
```diff
@@ -491,6 +497,23 @@ impl SparkWallet {
         &self,
         amount_sat: u64,
         receiver_address: &SparkAddress,
+    ) -> Result<WalletTransfer, SparkWalletError> {
+        if receiver_address.is_invoice() {
+            return Err(SparkWalletError::Generic(
+                "Receiver address is a Spark invoice. Use fulfill_spark_invoice instead."
```
dangeross:
```suggestion
                "Receiver address is a Spark invoice. Use `fulfill_spark_invoice` instead."
```

dangeross (commented):
Lookin good

crates/breez-sdk/core/src/persist/sqlite.rs
danielgranhao: Not really. In any case all data can be recovered from the invoice string. Changed to store just description and invoice string as suggested

crates/breez-sdk/core/src/sdk.rs
danielgranhao: Yes! That was a leftover. Removed now. Also changed the implementation on the breez sdk level to force the inclusion of a token id in token invoices in order to consider them valid. That was the case in the spark wallet already.

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -518,7 +556,12 @@ pub struct ReceivePaymentRequest {
 #[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
 pub struct ReceivePaymentResponse {
     pub payment_request: String,
-    pub fee_sats: u64,
+    /// Fee to pay to receive the payment
+    /// Denominated in sats if token identifier is empty, otherwise in the token base units
+    pub fee: u128,
+    /// The presence of this field indicates that the payment is for a token
+    /// If empty, it is a Bitcoin payment
+    pub token_identifier: Option<String>,
```
roeierez: Why is this field needed?

roeierez (commented):
This is really huge and looking very good!

docs/breez-sdk/snippets/kotlin_mpp_lib/shared/src/commonMain/kotlin/com/example/kotlinmpplib/ReceivePayment.kt
```diff
@@ -34,43 +36,85 @@ class ReceivePayment {
...
+            // handle error
+        }
+        // ANCHOR_END: receive-payment-spark-address
+    }
+
+    suspend fun receiveSparkInvoice(sdk: BreezSdk) {
+        // ANCHOR: receive-payment-spark-invoice
         try {
+            val optionalDescription = "<invoice description>"
+            val optionalAmountSats = BigInteger.fromLong(5_000L)
```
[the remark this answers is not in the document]
danielgranhao: Thanks. Fixed

crates/breez-sdk/core/src/models/mod.rs
danielgranhao: Dropped in [ad979bd](https://github.com/breez/spark-sdk/pull/349/commits/ad979bd7250b24ca36b336e8acb1adeff11b9011)

PR #355, opened 2025-10-25
log all sdk errors

JssDWt:
Ensure all sdk errors are logged, so we're able to find them later.

Also some debug logging added here and there.


danielgranhao:
Have you considered using `tracing::instrument`? I've never used it, but IIUC it would achieve what you're proposing here and more. Also, we wouldn't need to remember to use the error constructor methods.

JssDWt:
> Have you considered using `tracing::instrument`? I've never used it, but IIUC it would achieve what you're proposing here (and more) in a clean way. E.g. we wouldn't need to remember to use the error constructor methods.

I was trying to prove to you that `tracing::instrument` only creates a span, and doesn't log anything. But then I found `#[instrument(err)]`, which actually does log an event. So you're right, and we'll recreate this PR in another form.

PR #358, opened 2025-10-26
C# bindings

JssDWt:
Status: Working C# bindings (see https://github.com/breez/spark-sdk/actions/runs/19249120034)

This PR adds C# bindings to the Breez SDK. 

In order to achieve this, it changes the way the project is structured. Only types from `core` are exposed through uniffi now. This is a rather large change, mostly because the types (InputType and children, SyncStorage and children, FiatService and children, RestClient and children) all have to be duplicated and mapped, including traits.

- A change to uniffi-bindgen-cs is needed, see https://github.com/breez/uniffi-bindgen-cs/tree/jssdwt-fix-async-event-listener


roeierez (approved):
LGTM

Issue #368, opened 2025-11-02
Support non static deposit address


roeierez:
Goal is more privacy. 
The challenge here is how to manage this in an effective with good UX. This is giving the fact we are monitoring automatically user addresses. We don't want this list to get larger enforcing the sdk to do a lot of work where most of it is for nothing (Sending twice to the same dynamic address is really a negative flow).

One option is:
* As long as address doesn't have associated transactions, return it (this can be checked in realtime using the spark endpoint).
* Monitor only these addresses:
     1. Addresses without transactions (should be single one in most cases)
     2. Addresses with non zero balance (Only those that in progress, either claimable or refundable).

We should also document better the deposit flow and fix some semantic things that will better reflect best practices:
* Don't use the term FailedDepoist (we use it in FailedDepositEvent) should be more like UnclaimedDeposit as it is still part of the positive flow (user may claim manually).
* We should use some default in the max deposit fees. If user sets to None it should opt in for always claim manually (right now we always claim in this case).
* A link to the claim/refund documentation should be from within the Bitcoin section of the "Receiving Payments" section and not only from the events table (hidden at the bottom).
* In "Handling Refunds" page, which should be renamed to "Handling Unclaimed Deposits" or similar, we should instruct the user positively on how to complete the flow by claiming rather than just list the options (claim, refund).

danielgranhao:
From #362 (closing that and moving discussion to this issue):

Currently, fees for automatic deposit claims are managed by max_deposit_claim_fee in Config, allowing either a static max absolute fee or a max fee rate (sat/vB).

We should improve the UX here by having a dynamic fee leeway. This is a great match for single-use deposit addresses. Maybe it can also be applied to static deposit ones?

PR #370, opened 2025-11-03
Handling unclaimed deposits

dangeross:
This PR addresses the second part of #368 better improve the exposed terminology of unclaimed deposits and best practices.
- Renames events to `UnclaimedDeposits` / `ClaimedDeposits`
- Sets a default 1 sat/vb max_deposit_claim_fee (when unset, deposits need to be manually claimed)
- Updates docs

### DIFF
--- a/packages/flutter/rust/src/errors.rs
+++ b/packages/flutter/rust/src/errors.rs
@@ -7,12 +7,7 @@ pub enum _DepositClaimError {
     DepositClaimFeeExceeded {
         tx: String,
         vout: u32,
-        max_fee: Fee,
-        actual_fee: u64,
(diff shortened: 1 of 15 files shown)


packages/flutter/rust/src/errors.rs
```diff
@@ -30,7 +30,7 @@ pub enum _SdkError {
     DepositClaimFeeExceeded {
         tx: String,
         vout: u32,
-        max_fee: Fee,
+        max_fee: Option<Fee>,
```
danielgranhao: The `max_fee` in `_DepositClaimError` is missing the update

PR #375, opened 2025-11-04
Use a consistent db path for default storage

### LINKED ISSUE #369: python: using `default_storage()` causes event loop issues
When using the advanced SDK initialization, calling `default_storage()` results in errors when calling the async functions of the storage because it is no longer in the main event loop. The code:
```python
async def main():  
    try:
        storage = default_storage(data_dir="./.data")
        builder = SdkBuilder(config=config, seed=seed, storage=storage)
        sdk = await builder.build()
        info = await sdk.get_info(request=GetInfoRequest(ensure_synced=True))
        logging.info(f"Balance: {info.balance_sats} sats")
        return sdk
    except Exception as error:
        logging.error(error)
        raise

if __name__ == "__main__":
    asyncio.run(main())
```
causes:
```bash
Exception ignored on calling ctypes callback function <function _UniffiTraitImplStorage.set_cached_item at 0x103a31f80>:
Traceback (most recent call last):
  File "/Users/rosssavage/Source/Projects/breez-sdk-spark-python-test/.venv/lib/python3.13/site-packages/breez_sdk_spark/breez_sdk_spark.py", line 4107, in set_cached_item
    uniffi_out_return[0] = _uniffi_trait_interface_call_async_with_error(make_call, handle_success, handle_error, StorageError, _UniffiConverterTypeStorageError.lower)
  File "/Users/rosssavage/Source/Projects/breez-sdk-spark-python-test/.venv/lib/python3.13/site-packages/breez_sdk_spark/breez_sdk_spark.py", line 10959, in _uniffi_trait_interface_call_async_with_error
    eventloop = _uniffi_get_event_loop()
  File "/Users/rosssavage/Source/Projects/breez-sdk-spark-python-test/.venv/lib/python3.13/site-packages/breez_sdk_spark/breez_sdk_spark.py", line 10892, in _uniffi_get_event_loop
    return asyncio.get_running_loop()
RuntimeError: no running event loop
```

This can be fixed by setting the uniffi event loop, but this needs to be better documented for python (and perhaps we can expose this function more conveniently)
```python
from breez_sdk_spark.breez_sdk_spark import uniffi_set_event_loop

async def main():  
    uniffi_set_event_loop(asyncio.get_event_loop())
```

### LINKED ISSUE #373: sdk: using `default_storage()` directly doesn't isolate the mnemonic/network
With `connect()` the mnemonic fingerprint and network are added to the data dir path, but using `default_storage()` with the SdkBuilder uses the data dir directly without mnemonic fingerprint and network added. This could cause issues in the advanced initialization when switching mnemonics/networks

dangeross:
The used db storage path differs between when using `connect()` and `default_storage()` directly. 

This PR changes the SDK builder to allow either a default storage path or storage implementation to be set. As the `connect()` flow also uses the SDK builder's `with_default_storage()` the used directory is consistent for both flows.

Fixes #373
Closes #369 (`default_storage()` is removed)

### DIFF
--- a/packages/flutter/test/breez_sdk_spark_test.dart
+++ b/packages/flutter/test/breez_sdk_spark_test.dart
@@ -23,16 +23,15 @@ void main() {
   });
 
   test('Connect', () async {
-    const storageDir = './.data';
-    const mnemonic =
+    String mnemonic =
         "abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about";
-    const apiKey = "<api key>";
-    final storage = await defaultStorage(dataDir: storageDir);    
-    final config = defaultConfig(
-      network: Network.regtest,
-    ).copyWith(apiKey: apiKey);
+    final seed = Seed.mnemonic(mnemonic: mnemonic, passphrase: null);
 
-    final sdkBuilder = SdkBuilder(config: config, mnemonic: mnemonic, storage: storage);
+    final config = defaultConfig(network: Network.regtest)
+        .copyWith(apiKey: "<api key>");
+
+    final sdkBuilder = SdkBuilder(config: config, seed: seed);
+    sdkBuilder.withDefaultStorage(storageDir: './.data');
     final sdk = await sdkBuilder.build();
 
     /*final sdk = await connect(
--- a/packages/flutter/test/helper.dart
+++ b/packages/flutter/test/helper.dart
@@ -8,6 +8,8 @@ extension ConfigCopyWith on Config {
     Fee? maxDepositClaimFee,
     String? lnurlDomain,
     bool? preferSparkOverLightning,
+    List<ExternalInputParser>? externalInputParsers,
+    bool? useDefaultExternalInputParsers,
   }) {
     return Config(
       apiKey: apiKey ?? this.apiKey,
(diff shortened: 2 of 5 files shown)


crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -43,6 +44,45 @@ pub struct ConnectRequest {
     pub storage_dir: String,
 }
 
+#[derive(Clone)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct DefaultStorageRequest {
+    pub storage_dir: String,
+    pub network: Network,
+    pub seed: Seed,
```
danielgranhao: Not that it makes a world of difference, but it feels a bit odd to pass the seed when creating the storage. We could instead provide just the hash suffix. The logic for getting the hash could be moved to a `impl Seed` block. Wdyt?

danielgranhao (commented):
Nice! Small comment

packages/flutter/rust/src/sdk_builder.rs
danielgranhao: Missing the `with_storage` method here?

roeierez: Right but for now we are ok to omit it from flutter. No regression from previous.

PR #381, opened 2025-11-05
Expose private mode config

### LINKED ISSUE #380: Expose Spark privacy setting and enable by default

danielgranhao:
Closes #380 

Spark's private mode hides the user's balances and tx history in Spark indexers.

TODO:

- [x] Update wasm and flutter bindings
- [x] Update docs

### DIFF
--- a/crates/breez-sdk/core/src/models/mod.rs
+++ b/crates/breez-sdk/core/src/models/mod.rs
@@ -347,9 +347,9 @@ pub struct Config {
     /// Url to use for the real-time sync server. Defaults to the Breez real-time sync server.
     pub real_time_sync_server_url: Option<String>,
 
-    /// Wether the Spark private mode is enabled by default.
+    /// Whether the Spark private mode is enabled by default.
     ///
-    /// If set to true, the Spark private will be enabled on the first initialization of the SDK.
+    /// If set to true, the Spark private mode will be enabled on the first initialization of the SDK.
     /// If set to false, no changes will be made to the Spark private mode.
     pub private_enabled_default: bool,
 }


crates/breez-sdk/core/src/sdk.rs
```diff
@@ -228,6 +229,21 @@ impl BreezSdk {
     fn start(&self, initial_synced_sender: watch::Sender<bool>) {
         self.periodic_sync(initial_synced_sender);
         self.try_recover_lightning_address();
+        self.apply_spark_private_mode_setting();
```
roeierez: That may fail and users would think they are in private mode.
Perhaps an event such as UpdateWalletSettingFailed?
Also should we expose update_wallet_settings function in case users want to retry?

danielgranhao: We may add an event but I fear the UX will be suboptimal. I don't like that users would be required to handle/look for the event to know something is wrong.

I'm looking into the possibility of adding automatic retries and preventing payments before the it eventually succeeds. Wdyt?

crates/breez-sdk/core/src/sdk.rs
```diff
@@ -228,6 +229,21 @@ impl BreezSdk {
...
+    }
+
+    /// Applies the Spark private mode setting to the wallet
+    fn apply_spark_private_mode_setting(&self) {
+        let sdk = self.clone();
+        tokio::spawn(async move {
+            if let Err(e) = sdk
+                .spark_wallet
+                .update_wallet_settings(sdk.config.spark_private_mode_enabled)
```
JssDWt: I think there could be an issue with multi-app support here. Should this perhaps be a user-defined setting eventually? If one app uses privacy mode, and another uses non-privacy mode (for example for offline nostr zap receipts), the setting may flap. The lnurl server might not be able to send zap receipts when the privacy-mode app came online the latest. The privacy-mode app may leak payments when the non-privacy-mode app came online latest.

So eventually I think it's the user (not the developer) (or should be the user) that decides its privacy mode, depending on the apps it uses.

danielgranhao: Yes, there will be issues if multiple apps/instances disagree on the privacy mode. We may recommend integrators to expose this to users instead of deciding on their behalf?

> The privacy-mode app may leak payments when the non-privacy-mode app came online latest.

Yes, it will definetely leak. Setting the privacy mode off will leak all payments done while it was on.

If we want to optimize for intuitive behavior when multiple apps are used, we should expose the query/update API instead of a startup config that we attempt to apply remotely independently of the current remote state.

crates/breez-sdk/core/src/sdk.rs
danielgranhao: Added retries and a check before payments. Also, I propose we only enforce the setting was applied in case it is on. If turned off, I don't think we need to prevent payments.

crates/breez-sdk/core/src/sdk.rs
danielgranhao: So providing no setting would effectively mean: "I would prefer to go private, but if another app at any point disables private mode, I won't care again". I see how it would reduce flaps, but I fear we would still prefer to keep `Some(true)` as the default and I find it hard to reason when would someone opt for None. Also, users would have poor visibility into what's the actual current setting.

Coming back to the fact that disabling private mode will leak all payments from when it was on -> this by itself leads me to believe that mnemonics really shouldn't be shared across apps that disagree on the privacy setting.

I think there's 2 main options for a solution where app A needs privacy off and app B needs/prefers privacy on:

- We strongly recommend agaisnt sharing the mnemonic between A and B and simply document the issues (matching current implementation)
- We provide full support for sharing the mnemonic  between A and B. For this, I really think we should expose the query/update API. App A could query settings and prompt user to disable privacy. App B can show the current privacy setting and allow user to change.

kingonly: We should push for privacy. Are we relying non-private mode for zap receipts @JssDWt? I thought we've decided to also implement it in the client (ofc the UX isn't the same).

roeierez: > We should push for privacy. Are we relying non-private mode for zap receipts @JssDWt? I thought we've decided to also implement it in the client (ofc the UX isn't the same). 

Currently we only deployed the server side solution. Before we deploy this PR we should probably finalize and deploy the client side solution that will work with this private mode

crates/breez-sdk/core/src/persist/mod.rs
```diff
@@ -183,6 +189,25 @@ pub trait Storage: Send + Sync {
...
+    async fn set_user_settings_as_applied(
+        &self,
+        settings: Vec<UserSetting>,
+    ) -> Result<(), StorageError>;
+
+    /// Gets the applied user settings
+    /// # Returns
+    ///
+    /// A vector of applied user settings or a `StorageError`
+    async fn get_applied_user_settings(&self) -> Result<Vec<UserSetting>, StorageError>;
```
roeierez: Instead of extending the Storage trait can we use the key value table for this?

crates/breez-sdk/core/src/realtime_sync/storage.rs
```diff
@@ -13,20 +13,22 @@ use tracing::{debug, error};
...
-    SdkEvent, Storage, StorageError, UpdateDepositPayload,
+    SdkEvent, Storage, StorageError, UpdateDepositPayload, persist::UserSetting,
 };
 use tokio_with_wasm::alias as tokio;
 
 const INITIAL_SYNC_CACHE_KEY: &str = "sync_initial_complete";
 
 enum RecordType {
     PaymentMetadata,
+    DefaultUserSettingIsApplied,
```
roeierez: How about just name it `UserSetting`? The way we treat it internally can be kept as implementation details as we might change it in the future.

roeierez (changes requested):
I think that works. The disadvantage is that it is a bit complicated to reason about. I dropped some comments that might help with this. It would be nice to try and have the logic central as much as we can.

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -346,6 +346,16 @@ pub struct Config {
 
     /// Url to use for the real-time sync server. Defaults to the Breez real-time sync server.
     pub real_time_sync_server_url: Option<String>,
+
+    /// The initial user settings to be applied.
+    ///
+    /// This configuration is only relevant when the SDK is initialized for the first time for any given seed.
+    /// To adjust the settings after that, use [`BreezSdk::update_user_settings`](crate::sdk::BreezSdk::update_user_settings).
+    /// To get the current settings, use [`BreezSdk::get_user_settings`](crate::sdk::BreezSdk::get_user_settings).
```
dangeross: Is this over all devices/instances or the current device? What I'm asking really is if it's for all devices how does the partner/implementer know they should call update_user_settings instead?

crates/breez-sdk/core/src/sdk.rs
```diff
@@ -505,6 +649,26 @@ impl BreezSdk {
         );
         Ok(transfer)
     }
+
+    fn ensure_default_spark_private_mode_applied_if_on(&self) -> Result<(), SdkError> {
+        if !*self.default_user_settings_applied_watcher.borrow()
```
dangeross: I see an edge case where if the real-time sync is not available for some reason, maybe connectivity/server issues, an SDK without initial settings applied would be blocked. Are we ok with that?

roeierez: Right @dangeross  but it will only do so if the current config attempting to set the mode to private. It seems something we can't avoid based on the product decision?

crates/breez-sdk/core/src/models/mod.rs
danielgranhao: If the partner knows the seed is new, they can take advantage of this configuration. If not, and they need some user setting to be applied, they will need to use `update_user_settings`.

crates/breez-sdk/common/src/sync/model.rs
```diff
@@ -5,7 +5,7 @@ use semver::Version;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 
-const CURRENT_SCHEMA_VERSION: Version = Version::new(1, 0, 0);
+const CURRENT_SCHEMA_VERSION: Version = Version::new(2, 0, 0);
```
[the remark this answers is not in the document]
danielgranhao: Fixed

danielgranhao:
Ok, so another big refactor. Thanks @roeierez for realizing we could do something much simpler. 

Changes:

- Got rid of rtsync usage
- `Config` now has a `private_enabled_defaukt` boolean. If true, we try to set private mode on on first initialization. If false, no changes are made. We block payments until we succeed at applying the private mode if it's enabled by default and first init hasn't been made.

This leads to a practically equal UX and is much simpler. Most partners don't need to do anything and get privacy by default. Partners that need privacy off can set the default off to avoid having to use `update_user_settings` (only when seed is new).

roeierez (commented):
Very nice

crates/breez-sdk/wasm/src/models/mod.rs
```diff
@@ -991,3 +992,13 @@ pub struct OutgoingChange {
...
 }
+
+#[macros::extern_wasm_bindgen(breez_sdk_spark::UserSettings)]
+pub struct UserSettings {
+    pub spark_private_mode_enabled: bool,
+}
+
+#[macros::extern_wasm_bindgen(breez_sdk_spark::UpdateUserSettingsRequest)]
+pub struct UpdateUserSettingsRequest {
+    pub enable_spark_private_mode: Option<bool>,
```
roeierez: Why optional? Also shouldn't we allow the user change it to pubic also?

roeierez (changes requested):
One comment otherwise LGTM

PR #382, opened 2025-11-06
Emit pending payment events

### LINKED ISSUE #312: Pending payment mode for faster receiving
Receiving a payment is currently rather slow. This can be sped up by triggering a 'PaymentPending' event the moment a transfer event is received. Currently a PaymentSucceeded event is triggered after the payment is claimed, which takes time.

danielgranhao:
Closes #312

TODO:

- [x] Docs

### DIFF
--- a/crates/breez-sdk/core/src/sdk.rs
+++ b/crates/breez-sdk/core/src/sdk.rs
@@ -366,8 +366,8 @@ impl BreezSdk {
                     error!("Failed to sync wallet: {e:?}");
                 }
             }
-            WalletEvent::FoundClaimableTransfer(transfer) => {
-                info!("Found a claimable transfer");
+            WalletEvent::TransferClaimStarting(transfer) => {
+                info!("Transfer claim starting");
                 if let Ok(payment) = Payment::try_from(transfer) {
                     // Insert the payment into storage to make it immediately available for listing
                     if let Err(e) = self.storage.insert_payment(payment.clone()).await {
--- a/crates/internal/src/main.rs
+++ b/crates/internal/src/main.rs
@@ -104,7 +104,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                         spark_wallet::WalletEvent::StreamDisconnected => println!("Disconnected from Spark server."),
                         spark_wallet::WalletEvent::Synced => println!("Synced"),
                         spark_wallet::WalletEvent::TransferClaimed(transfer) => println!("Transfer claimed: {}", transfer.id),
-                        spark_wallet::WalletEvent::FoundClaimableTransfer(transfer) => println!("Found claimable transfer: {}", transfer.id),
+                        spark_wallet::WalletEvent::TransferClaimStarting(transfer) => println!("Transfer claim starting: {}", transfer.id),
                     }
                 }
                 else => eprintln!("Event stream closed."),
--- a/crates/spark-wallet/src/wallet.rs
+++ b/crates/spark-wallet/src/wallet.rs
@@ -1225,7 +1225,7 @@ impl BackgroundProcessor {
         };
 
         self.event_manager
-            .notify_listeners(WalletEvent::FoundClaimableTransfer(
+            .notify_listeners(WalletEvent::TransferClaimStarting(
                 WalletTransfer::from_transfer(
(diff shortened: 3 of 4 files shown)


roeierez (commented):
Should we also emit this event for outgoing payments as we do with other payment events?

danielgranhao:
> Should we also emit this event for outgoing payments as we do with other payment events?

I was wondering about that. Felt that such event would be redundant as payments are started by the user, so the event wouldn't be needed? But maybe for consistency we should. Or at least rename to `IncomingPendingPayment`.

roeierez:
> > Should we also emit this event for outgoing payments as we do with other payment events?
> 
> I was wondering about that. Felt that such event would be redundant as payments are started by the user, so the event wouldn't be needed? But maybe for consistency we should. Or at least rename to `IncomingPendingPayment`.

One advantage for having pending event for outgoing is when the operation takes time (such when the user pass a long timeout for lightning payments in send_payment method) the event allows the app to react. Not critical though as usually apps that sets this timeout are waiting in a payment screen (not showing the transactions list).

crates/internal/src/main.rs
```diff
@@ -104,6 +104,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                         spark_wallet::WalletEvent::StreamDisconnected => println!("Disconnected from Spark server."),
                         spark_wallet::WalletEvent::Synced => println!("Synced"),
                         spark_wallet::WalletEvent::TransferClaimed(transfer) => println!("Transfer claimed: {}", transfer.id),
+                        spark_wallet::WalletEvent::FoundClaimableTransfer(transfer) => println!("Found claimable transfer: {}", transfer.id),
```
roeierez: We have TransferedClaimed, perhaps we can rename that to TransferClaiming or TransferClaimStarted ?

crates/breez-sdk/core/src/sdk.rs
```diff
@@ -362,6 +362,17 @@ impl BreezSdk {
                     error!("Failed to sync wallet: {e:?}");
                 }
             }
+            WalletEvent::FoundClaimableTransfer(transfer) => {
+                info!("Found a claimable transfer");
+                if let Ok(payment) = transfer.try_into() {
+                    self.event_emitter
+                        .emit(&SdkEvent::PaymentPending { payment })
```
roeierez: There is a small issue here I think that when app receives this event probably the logic would be to refresh payments list and then the pending payment will not be there since we are not synced yet. Should we add it to the db as pending at this point? If so we also need to make sure the index for next payments page stays correct.

crates/internal/src/main.rs
danielgranhao: Renamed to `TransferClaimStarting`

crates/spark-wallet/src/wallet.rs
```diff
@@ -1227,6 +1224,19 @@ impl BackgroundProcessor {
                 .next()
```
JssDWt: It would be nice if we could emit the pending payment event before querying the SSP. Or even not query the SSP at all. It's just metadata we're querying, which seems cacheable. Do you think we could cache that? (I didn't look at it long)

JssDWt: My idea for the payment pending event was to notify immediately when hearing about the payment, so you can provide quick feedback in your UI. So I find the API call disturbing for that purpose. An api call adds time, so payment latency. I'm thinking maybe the details are not at all important for the PaymentPending event actually? Can we pass None instead? Or do 2 PaymentPending events in a row?

roeierez: I see what you mean Jesse but since spark operator gives us a payment with partial data I think it will be too confusing to emit it with partial data. Perhaps we can start this way (it is a nice progress over what we have now) and if we will see it is necessary we can add another event which emits quicker?

danielgranhao: Yes, not saying it is perfect. Agree that ideally we would push the pending even asap. On the other hand, I also feel it's important to provide consistent and correct payment data and we already make significant savings with the current approach as the heavist process is claiming the transfer.

JssDWt (approved):

PR #389, opened 2025-11-07
client side zaps

JssDWt:
This PR adds zap functionality if privacy mode is enabled. It also enables lnurl comments up to 255 characters. In a nutshell it works as follows
- client side registers nostr pubkey along with lightning address
- client synchronizes invoice metadata from lnurl server when receiving a payment with a description hash, and periodically.
- in the background, client publishes zap receipts where necessary. Zap receipts are published through the lnurl server, so they don't have to be real-time synced.

Other comments
- I changed the way sync works a bit. Parallelizing it, and making it so you can more precisely set which things to sync by using bitflags.
- the nostr pubkey is set in the register lightning address request based on the privacy setting on the operator. This means before registering the lightning address, we now query the operator for the privacy setting.
- the nostr pubkey is derived from the seed, according to NIP-06 (I think)
- the sender comment and the nostr zap request are immediately available when the PaymentSucceeded event is published. The zap receipt is available later, after it is published.

### DIFF
--- a/crates/breez-sdk/core/src/sdk.rs
+++ b/crates/breez-sdk/core/src/sdk.rs
@@ -1573,9 +1573,10 @@ impl BreezSdk {
                 self.storage
                     .delete_deposit(detailed_utxo.txid.to_string(), detailed_utxo.vout)
                     .await?;
-                if let Err(e) = self.sync_trigger.send(SyncRequest::no_reply(
-                    SyncType::WalletState | SyncType::Deposits,
-                )) {
+                if let Err(e) = self
+                    .sync_trigger
+                    .send(SyncRequest::no_reply(SyncType::WalletState))
+                {
                     error!("Failed to execute sync after deposit claim: {e:?}");
                 }
                 Ok(ClaimDepositResponse {


JssDWt:
Question @roeierez Should we have a websocket with the lnurl server to stream created invoices? So the data is ready before the payment arrives and we can publish the PaymentSucceeded event on TransferClaimed including the lnurl comment and nostr event? I don't see another way to include that data in that event. Even with the websocket it could be racy. 

Another option is to check whether the invoice has a description hash. If it does, sync the lnurl receive metadata before emitting PaymentSucceeded.

roeierez:
> Question @roeierez Should we have a websocket with the lnurl server to stream created invoices? So the data is ready before the payment arrives and we can publish the PaymentSucceeded event on TransferClaimed including the lnurl comment and nostr event? I don't see another way to include that data in that event. Even with the websocket it could be racy.
> 
> Another option is to check whether the invoice has a description hash. If it does, sync the lnurl receive metadata before emitting PaymentSucceeded.

I would prefer not to complicate things atm for this feature (we will see later how we can do better). I think it is fine if the event won't include the nostr event and will be added to the payment later.

crates/breez-sdk/core/src/nostr.rs
```diff
@@ -0,0 +1,43 @@
...
+            secp,
+            nostr_key: nostr_key.private_key,
+        })
+    }
+
+    pub fn nostr_pubkey(&self) -> String {
+        let (xonly_pubkey, _) = XOnlyPublicKey::from_keypair(&self.nostr_key.keypair(&self.secp));
+        xonly_pubkey.to_string()
+    }
+}
```
roeierez: Are we using this only to calculate the nostr_pubkey? Perhaps instead of passing the seed only for this purpose we can use the signer directly to derive the pubkey?

roeierez:
> Another option is to check whether the invoice has a description hash. If it does, sync the lnurl receive metadata before emitting PaymentSucceeded.

Gave it another thought and I think this suggestion might be not too complicated so I would go with this option.

crates/breez-sdk/core/src/nostr.rs
roeierez: Just notifying about this @JssDWt  in case you missed it.

JssDWt: @roeierez this way it's compatible with NIP-06, which I thought was a nice feat:
https://nostr-nips.com/nip-06

crates/breez-sdk/lnurl/migrations/sqlite/20251107_sender_comments.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE sender_comments
+(
+    payment_hash   TEXT NOT NULL PRIMARY KEY,
+    user_pubkey    TEXT NOT NULL,
+    sender_comment TEXT NOT NULL,
+    invoice_expiry BIGINT NOT NULL,
```
roeierez: What is the purpose of invoice_expiry at this context?

roeierez (commented):
Small question, otherwise LGTM

crates/breez-sdk/lnurl/migrations/sqlite/20251107_sender_comments.sql
JssDWt: Removed that field

crates/breez-sdk/core/src/persist/mod.rs
```diff
@@ -1094,6 +1152,19 @@ pub mod tests {
...
+                        (Some(r_info), Some(e_info)) => {
+                            assert_eq!(r_info.nostr_zap_request, e_info.nostr_zap_request);
+                            assert_eq!(r_info.sender_comment, e_info.sender_comment);
+                        }
+                        (None, None) => {}
+                        _ => panic!(
+                            "LNURL receive metadata mismatch for payment {}",
+                            expected_payment.id
+                        ),
+                    }
```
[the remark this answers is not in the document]
JssDWt: Well spotted! Done.

crates/breez-sdk/core/src/sdk.rs
```diff
@@ -281,6 +309,131 @@ impl BreezSdk {
...
+        let mut offset = 0;
+        let limit = 100;
+        loop {
+            let payments = sdk
+                .storage
+                .list_payments(ListPaymentsRequest {
+                    offset: Some(offset),
+                    limit: Some(limit),
+                    ..Default::default()
+                })
```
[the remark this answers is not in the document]
JssDWt: Yes, done

PR #396, opened 2025-11-11
Fix token txid hex encoding

### LINKED ISSUE #383: Sending a 2nd tokens payment fails until sync
When doing a 2nd payment using the same token of the 1st, the SO returns an error that states `failed to fetch input TTXOs` and `no transaction found for prev tx hash <hash>`.

After syncing, the payment succeeds. 

This should be a bug in how we update the set of owned tx outputs.

danielgranhao:
Closes #383

`Txid::to_string()` encodes the txid in hex but in the reverse order of what we need.

This also fixes the payment observer as it was previously also using the reversed hex string.

### DIFF
--- a/crates/breez-sdk/core/src/models/payment_observer.rs
+++ b/crates/breez-sdk/core/src/models/payment_observer.rs
@@ -118,7 +118,7 @@ impl spark_wallet::TransferObserver for SparkTransferObserver {
 
     async fn before_send_token(
         &self,
-        tx_id: &bitcoin::Txid,
+        tx_id: &str,
         token_id: &str,
         receiver_outputs: Vec<spark_wallet::ReceiverTokenOutput>,
     ) -> Result<(), TransferObserverError> {
--- a/crates/spark/src/services/tokens.rs
+++ b/crates/spark/src/services/tokens.rs
@@ -1,7 +1,6 @@
 use std::{collections::HashMap, ops::Not, sync::Arc};
 
 use bitcoin::{
-    Txid,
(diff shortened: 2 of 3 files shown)


dangeross (commented):
I just did a quick test to check the payment observer and it also looks like the payment_id is reversed:
```
INFO:root:Preparing to send payment: eec988d3ed0fbf367b8de71fef7d2d088b604b9a182d8ed5ba77e20f5d6b2429:0
INFO:root:Payment sent: 29246b5d0fe277bad58e2d189a4b608b082d7def1fe78d7b36bf0fedd388c9ee:0
```

On the receiving side and in sparkscan the txid is 29246b5d0fe277bad58e2d189a4b608b082d7def1fe78d7b36bf0fedd388c9ee

Could you address that also in your PR? Could our usage of Txid::from_splice be wrong?

danielgranhao:
@JssDWt @dangeross 

From the docs of `Txid`:

"For compatibility with the existing Bitcoin infrastructure and historical and current versions of the Bitcoin Core software itself, this and other `sha256d::Hash` types, are serialized in reverse byte order when converted to a hex string via `std::fmt::Display` trait operations. See `hashes::Hash::DISPLAY_BACKWARD` for more details."

Spark tokens (at least on sparkscan) don't adher to this reverse byte order display. Since `Txid` is really meant to represent Bitcoin tx ids, I propose we don't use it entirely. See [2532c54](https://github.com/breez/spark-sdk/pull/396/commits/2532c54cbf06060c543229a7bbb303702f90283d)

PR #398, opened 2025-11-12
C# docs

JssDWt:
Add documentation for C#.

The amount types were actually strings. I have changed those into proper BigInteger types in this PR too.

### DIFF
--- a/docs/breez-sdk/snippets/csharp/ListPayments.cs
+++ b/docs/breez-sdk/snippets/csharp/ListPayments.cs
@@ -37,9 +37,14 @@ namespace BreezSdkSnippets
                 // Filter by status
                 statusFilter: new List<PaymentStatus> { PaymentStatus.Completed },
                 assetFilter: assetFilter,
-                // Offset and limit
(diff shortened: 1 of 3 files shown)


docs/breez-sdk/snippets/csharp/GettingStarted.cs
```diff
@@ -0,0 +1,109 @@
...
+        void SetLogger(SdkLogger logger)
+        {
+            BreezSdkSparkMethods.InitLogging(logDir: null, appLogger: logger, logFilter: null);
+        }
+        // ANCHOR_END: logging
+
+        // ANCHOR: add-event-listener
+        class SdkListener : EventListener
+        {
+            public async Task OnEvent(SdkEvent sdkEvent)
```
danielgranhao: Nit: this is throwing a warning in CI. It's fine, but it will always show the warning which is annoying. Can we add some rule to ignore this?

dangeross (approved):
LGTM aside of Daniel's comments

PR #409, opened 2025-11-12
Add idempotency key check to bitcoin/spark payments

### LINKED ISSUE #402: Add Idempotency to sdk.SendPayment() for Spark transactions
We need some form of idempotency on the SDK for spark payments.

Currently, if the mobile doesnt function properly, multiple SendPayment calls can be made with no controls in place.

We would like a key that creates a NOOP if seen before.

For example

`sdk.SendPayment(prepareResponse, uuid)`

dangeross:
This PR adds to the send payment options an idempotency_key for Bitcoin and Spark payments. This key can be used to ensure a payment is not made multiple times to a Bitcoin address or Spark address/invoice. This is ensured in two ways:
- If a payment already exists with this id, return the payment immediately
- If no payment exists yet but a transfer has been started using this key as the transfer id, the SO will throw an error for this transfer. In this case we query for the started transfer and return it as a payment

TODO:
- [x] Docs

Fixes #402

### DIFF
--- a/crates/breez-sdk/core/src/sdk.rs
+++ b/crates/breez-sdk/core/src/sdk.rs
@@ -1504,6 +1504,12 @@ impl BreezSdk {
         request: SendPaymentRequest,
         suppress_payment_event: bool,
     ) -> Result<SendPaymentResponse, SdkError> {
+        if request.idempotency_key.is_some() && request.prepare_response.token_identifier.is_some()
+        {
+            return Err(SdkError::InvalidInput(
+                "Idempotency key is not supported for token payments".to_string(),
+            ));
+        }
         if let Some(idempotency_key) = &request.idempotency_key {
             // If an idempotency key is provided, check if a payment with that id already exists
             if let Ok(payment) = self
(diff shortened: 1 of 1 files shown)


dangeross:
> what is the advantage of having the idempotency key in the SendOptions vs directly on the SendPaymentRequest?

@roeierez it started as a Spark only option and I later added it to Bitcoin also. My rational was as it doesn't apply to all payment methods (not lightning or tokens), best to keep it in the options.

I'm also not strongly opinionated and fine moving it to SendPaymentRequest

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -737,6 +737,13 @@ pub struct PrepareSendPaymentResponse {
 pub enum SendPaymentOptions {
     BitcoinAddress {
         confirmation_speed: OnchainConfirmationSpeed,
+
+        /// If set, providing the same idempotency key for multiple requests will ensure that only one
+        /// payment is made.
+        ///
+        /// If an idempotency key is re-used, the same payment will be returned if available; otherwise,
+        /// the payment request will result in an error.
```
danielgranhao: I think we should mention it needs to be an UUID

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -745,6 +752,14 @@ pub enum SendPaymentOptions {
         /// number of seconds. If unset, the function will return immediately after initiating the payment.
         completion_timeout_secs: Option<u32>,
     },
+    Spark {
+        /// Providing the same idempotency key for multiple requests will ensure that only one
+        /// payment is made. The idempotency key is only valid for non-token Spark transfers.
+        ///
+        /// If an idempotency key is re-used, the same payment will be returned if available; otherwise,
+        /// the payment request will result in an error.
```
danielgranhao: Same here

PR #410, opened 2025-11-13
Domain check using fly.io certificates

yaslama:


crates/domain-validator/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
```
roeierez: I think it is better to move the domain-validator and fly-api crates under the breez sdk folder where lnurl crate is located. 
Also would consider moving the domain-validator content into lnurl-models crate and get rid of the domain-validtor crate (I am not a fan of one trait crate).

PR #420, opened 2025-11-15
spark: Add token output reservation

### LINKED ISSUE #118: Support concurrent token operations/transactions
Currently, the implementation of `transfer_tokens()` holds a lock on owned token outputs for the entire duration of the transaction preparation and broadcast. This prevents concurrent token transactions or any other operations on the outputs (e.g. refreshing them).

We should optimize this. There are several options, but one of them is to implement a reservation mechanism similar to what we have for spark leaves.

dangeross:
Adds service/state management of token outputs to support concurrent token operations through token output reservation.

TODO:
- [x] Add tests

Closes #118

### DIFF
--- a/crates/spark/src/token/service.rs
+++ b/crates/spark/src/token/service.rs
@@ -1,6 +1,7 @@
-use std::{collections::HashMap, sync::Arc};
+use std::{collections::HashMap, sync::Arc, time::Duration};
 
-use tracing::warn;
+use tokio_with_wasm::alias as tokio;
+use tracing::{info, warn};
 
 use crate::{
     Network,
@@ -18,6 +19,8 @@ use crate::{
         error::TokenOutputServiceError,
     },
 };
+
+const SELECT_TOKEN_OUTPUTS_MAX_RETRIES: u32 = 3;
 
 pub struct SynchronousTokenOutputService {
     network: Network,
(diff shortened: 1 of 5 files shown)


crates/spark/src/token/store.rs
```diff
@@ -0,0 +1,273 @@
...
+        }
+
+        trace!(
+            "Updated {} token outputs in the local state",
+            token_outputs.len()
+        );
+        Ok(())
+    }
+
+    async fn get_token_metadata(
```
dangeross: This can be performed by the service

danielgranhao (changes requested):
Looks great! Just one issue

crates/spark/src/token/service.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    async fn reserve_token_outputs(
+        &self,
+        token_identifier: &str,
+        amount: u128,
+        preferred_outputs: Option<Vec<TokenOutputWithPrevOut>>,
+    ) -> Result<TokenOutputsReservation, TokenOutputServiceError> {
+        self.state
+            .reserve_token_outputs(token_identifier, amount, preferred_outputs)
```
roeierez: In select_leaves function of the TreeService we do some retries and refresh the leaves on first failure to mitigate sending right on start (before sync is completed). Do we want to apply a similar pattern here?

roeierez (changes requested):
Looks good.

Issue #425, opened 2025-11-17
Allow initializing SDK with custom account number


danielgranhao:
This allows issuing multiple tokens from the same mnemonic.

danielgranhao:
Didn't realize we already had this implemented in #292. Let's improve the documentation to make it clear how it can be used.

PR #428, opened 2025-11-18
Docs: creating multiple tokens

danielgranhao:
Related: #425

### DIFF
--- a/docs/breez-sdk/src/guide/issuing_tokens.md
+++ b/docs/breez-sdk/src/guide/issuing_tokens.md
@@ -14,9 +14,13 @@ The Breez SDK provides a specialized Token Issuer interface for managing custom 
 
 Create a custom token with configurable parameters. Define the decimal precision, max supply and if the token can be frozen.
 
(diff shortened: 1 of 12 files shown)


docs/breez-sdk/src/guide/issuing_tokens.md
```diff
@@ -14,10 +14,14 @@ The Breez SDK provides a specialized Token Issuer interface for managing custom
 
 Create a custom token with configurable parameters. Define the decimal precision, max supply and if the token can be frozen.
 
-**Note:** Token creation is limited to one token per issuer wallet
-
 {{#tabs issuing_tokens:create-token}}
 
+### Creating multiple tokens
+
+Token creation is limited to one token per issuer wallet. If you need to create and then manage more than one token using the same mnemonic, we recommend using different account numbers when initializing the SDK.
```
roeierez: Why do we need to document this? Isn't it the same as multiple sdk instances where each instance actually mapped to a different spark user?

PR #435, opened 2025-11-21
breez-itest improvements

danielgranhao:
This PR:
- Adds `--no-fail-fast` so that any failures do not stop us from seeing all failures
- Moves the renew timelocks test to `spark-itest`, increasing its reliability and speeding it up from ~15 minutes to under 100 seconds
- Fixes a bug in the renew timelocks implementation that would prevent root nodes from being renewed when they reached zero timelock - this was never noticed before because a root node would almost immediately be swapped during the optimization process (using SSP, which isn't available on `spark-itest`)


JssDWt (commented):
It looks good.

It seems nice to have nextest do retries. But also it may hide _real_ races/flakes. I would prefer to keep it at a single shot per test for that reason. And fix our tests such that they always succeed. 

For the renew_timelocks test, that seems like a test we could run against a local SO, because no SSP is needed?

danielgranhao:
> It seems nice to have nextest do retries. But also it may hide _real_ races/flakes. I would prefer to keep it at a single shot per test for that reason. And fix our tests such that they always succeed.

I see your point and share that concern, yet I think this may be an efficient way to address resource exhaustion issues. It may happen, but I think in most cases, flakiness will not be caused by faults with our implementation. Would reducing the number of retries to just 1 help reduce your concern?

We may also open an issue to resolve the underlying causes of flakiness and eliminate retries, but keep them for now so integration tests still provide value. 

> For the renew_timelocks test, that seems like a test we could run against a local SO, because no SSP is needed?

Good idea! I'll address that

JssDWt (approved):

crates/spark/src/services/timelock_manager.rs
```diff
@@ -148,11 +148,24 @@ impl TimelockManager {
         &self,
         renew_type: RenewType,
         node: &TreeNode,
-        parent_node: &TreeNode,
+        parent_node: Option<&TreeNode>,
```
JssDWt: What do you think about putting the parent node inside the RenewType enum?

```
enum RenewType {
    Node { parent: TreeNode },
    Refund  { parent: TreeNode },
    ZeroTimelock,
}
```

danielgranhao: Good idea! Done

JssDWt (approved):

PR #439, opened 2025-11-24
Fix coop exit idempotency

danielgranhao:
This fixes the failing test `test_03_bitcoin_idempotency_key`.

### DIFF
--- a/crates/spark/src/services/coop_exit.rs
+++ b/crates/spark/src/services/coop_exit.rs
@@ -163,11 +163,14 @@ impl CoopExitService {
         trace!("Leaf external IDs for cooperative exit: {leaf_external_ids:?}");
         trace!("Fee leaf external IDs for cooperative exit: {fee_leaf_external_ids:?}");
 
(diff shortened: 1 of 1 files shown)


crates/spark/src/services/coop_exit.rs
```diff
@@ -177,19 +181,29 @@ impl CoopExitService {
...
                 withdrawal_address: withdrawal_address.to_string(),
-                idempotency_key: None,
+                idempotency_key: Some(unwrapped_transfer_id.to_string()),
                 exit_speed: exit_speed.into(),
                 withdraw_all,
                 fee_leaf_external_ids,
                 fee_quote_id,
-                user_outbound_transfer_external_id: Some(transfer_id.to_string()),
+                user_outbound_transfer_external_id: None,
             })
```
JssDWt: Does this actually set the transfer id of the resulting transfer now? If not, I think we should rename the transfer_id field here and in upper layers to idempotency_key instead.

PR #443, opened 2025-11-25
Use SSP pubkey to identify Lightning vs HTLC payments

danielgranhao:
This proposes an improvement over #440. It drops the reliance on the SSP in order to distinguish Lightning from HTLC  payments.

Instead, we infer if a payment is Lightning if the transfer was sent to or received from the SSP. This only works while there is a single static SSP, and we'll need to find another solution when that stops being the case.

### DIFF
--- a/crates/breez-sdk/core/src/models/adaptors.rs
+++ b/crates/breez-sdk/core/src/models/adaptors.rs
@@ -16,30 +16,17 @@ use crate::{
 
 impl PaymentMethod {
     fn from_transfer(transfer: &WalletTransfer) -> Self {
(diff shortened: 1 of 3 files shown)


roeierez:
@danielgranhao what is the advantage of this method given the fact the user request is populated automatically by the SO?

danielgranhao:
> @danielgranhao what is the advantage of this method given the fact the user request is populated automatically by the SO?

Increased resiliency. If, for any reason, the SSP fails to provide the lightning user request, we end up with empty `PaymentDetails` instead of showing the incorrect `Spark` details with HTLC.

roeierez (approved):
LGTM

PR #445, opened 2025-11-26
Unclaimed deposit handling improvements

### LINKED ISSUE #437: Improve on-chain claim documentation
Discussed with @danielgranhao offline. Developers do not understand how to handle on-chain claims. 
1. We need to make it simpler and explain that we allow auto claim by setting a x sats/vbyte or an absolute fee (which developers can expose or not in the app settings).
2. Describe that best practice could be simply to show a UI to the user to accept the actual fee and show a sample code that does it.
3. Also explain that developers can have their own business logic for automating claims like for example checking that the actual fees are close to the recommended fees and automatically accept such claim. Show a sample code that does it.

danielgranhao:
Closes #437 

Changes:
- Renamed `DepositClaimFeeExceeded` to `MaxDepositClaimFeeExceeded` (breaking)
- Changed `max_fee` inner type from `Fee` to `u64` (breaking) - for an integrator, the currently configured fee rate isn't useful as they have no way to determine which absolute max fee it results in
- Renamed `actual_fee` to `required_fee` (breaking)
- Docs improvements
    - Event handling snippet now shows how to get event data
    - Recommended fees snippet shows correct units (sats/vbyte)
    - Fix Spark lightning docs link
    - Handling unclaimed deposit docs major refactor
    - Added moving to production page (to state that unclaimed deposits should be handled)

### DIFF
--- a/crates/breez-sdk/core/src/persist/sqlite.rs
+++ b/crates/breez-sdk/core/src/persist/sqlite.rs
@@ -248,6 +248,9 @@ impl SqliteStorage {
                 nostr_zap_receipt TEXT,
                 sender_comment TEXT
             );",
+            // Delete all unclaimed deposits to clear old claim_error JSON format.
+            // Deposits will be recovered on next sync.
+            "DELETE FROM unclaimed_deposits;",
         ]
     }
 }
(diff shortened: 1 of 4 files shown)


JssDWt (commented):
Looking good Daniel. Just some docs nits.

PR #448, opened 2025-11-27
Enable BOLT11 spark address hint

### LINKED ISSUE #432: No paid event emitted when invoice is settled via Spark
Now that private mode is added we are looking to enable including the Spark address by default when creating a Lightning invoice. One gap we see is being able to account (from an implementation POV) that the invoice is paid when paid via transfer (e.g. a merchant waiting for an "invoice paid" event). Possible ways of solving is by using MRH.

dangeross:
This PR: 
- Creates BOLT11 invoices with spark address included by default
- Excludes spark address in LNURL invoice requests including a nostr zap request to ensure the invoice is paid over lightning

Closes #432

### DIFF
--- a/crates/breez-sdk/core/src/sdk.rs
+++ b/crates/breez-sdk/core/src/sdk.rs
@@ -1236,7 +1236,10 @@ impl BreezSdk {
                     amount: request.prepare_response.amount_sats.into(),
                     token_identifier: None,
                 },
-                options: None,
+                options: Some(SendPaymentOptions::Bolt11Invoice {
+                    prefer_spark: false,
+                    completion_timeout_secs: None,
+                }),
                 idempotency_key: request.idempotency_key,
             },
             true,


crates/breez-sdk/core/src/sdk.rs
```diff
@@ -949,7 +943,7 @@ impl BreezSdk {
     async fn initialize_spark_private_mode(&self) -> Result<(), SdkError> {
         if !self.config.private_enabled_default {
             ObjectCacheRepository::new(self.storage.clone())
-                .save_spark_private_mode_initialized()
+                .save_spark_private_mode(false)
```
JssDWt: Here you could actually be in private mode? The private mode was 'initialized', because we don't have to tell the server if we want to be private. But if we don't care if we're private or not (`private_enabled_default == false`), that doesn't mean were 'not private'.

dangeross:
> And if it's public mode - the account is already public, so there's no reason not to include the address.

I thought we could at least be preventative in making the address public over Lightning if the history/balance is publicly available. 

I will refactor the PR

JssDWt:
This change is not 'only' a config change. 

When a user creates a lightning invoice, it will now - by default - contain a MRH. When the sender pays an invoice with MRH, it will now - by default - pay through Spark. On both sides, however, the payment cannot be associated with the invoice. So when the payment is fetched from the SO, and is displayed in the list, it will not contain the invoice.

Let's put this PR on hold until we know what we want to do with that?

PR #457, opened 2025-12-01
add timestamp to lnurl api signatures

JssDWt:
The timestamp is optional temporarily, so old clients can still work. We'll have to remove the backward compatibility after some time.


dangeross (approved):
LGTM

roeierez (approved):
Small comment , otherwise LGTM.

PR #459, opened 2025-12-02
Prep for token to lightning payments

dangeross:
Preparation PR for token to lightning payments

- Adds Flashnet client
- Handles failed conversions by manually checking and performing refunds if necessary

### DIFF
--- a/crates/breez-sdk/cli/src/command/mod.rs
+++ b/crates/breez-sdk/cli/src/command/mod.rs
@@ -5,11 +5,11 @@ use breez_sdk_spark::{
     AssetFilter, BreezSdk, CheckLightningAddressRequest, ClaimDepositRequest,
     ClaimHtlcPaymentRequest, Fee, GetInfoRequest, GetPaymentRequest, GetTokensMetadataRequest,
     InputType, LightningAddressDetails, ListPaymentsRequest, ListUnclaimedDepositsRequest,
-    LnurlPayRequest, LnurlWithdrawRequest, MaxFee, OnchainConfirmationSpeed, PaymentStatus,
-    PaymentType, PrepareLnurlPayRequest, PrepareSendPaymentRequest, ReceivePaymentMethod,
-    ReceivePaymentRequest, RefundDepositRequest, RegisterLightningAddressRequest,
-    SendPaymentMethod, SendPaymentOptions, SendPaymentRequest, SparkHtlcOptions, SparkHtlcStatus,
-    SyncWalletRequest, TokenIssuer, UpdateUserSettingsRequest,
+    LnurlPayRequest, LnurlWithdrawRequest, MaxFee, OnchainConfirmationSpeed, PaymentDetailsFilter,
+    PaymentStatus, PaymentType, PrepareLnurlPayRequest, PrepareSendPaymentRequest,
+    ReceivePaymentMethod, ReceivePaymentRequest, RefundDepositRequest,
+    RegisterLightningAddressRequest, SendPaymentMethod, SendPaymentOptions, SendPaymentRequest,
+    SparkHtlcOptions, SparkHtlcStatus, SyncWalletRequest, TokenIssuer, UpdateUserSettingsRequest,
 };
 use clap::Parser;
 use rand::RngCore;
@@ -60,6 +60,10 @@ pub enum Command {
         /// Filter by Spark HTLC status
         #[arg(long)]
         spark_htlc_status_filter: Option<Vec<SparkHtlcStatus>>,
+
+        /// Filter by token transaction hash
+        #[arg(long)]
+        tx_hash: Option<String>,
 
         /// Only include payments created after this timestamp (inclusive)
         #[arg(long)]
@@ -278,11 +282,30 @@ pub(crate) async fn execute_command(
             type_filter,
             status_filter,
             spark_htlc_status_filter,
+            tx_hash,
             asset_filter,
             from_timestamp,
             to_timestamp,
             sort_ascending,
         } => {
+            let mut payment_details_filter = Vec::new();
+            if let Some(statuses) = spark_htlc_status_filter {
+                payment_details_filter.push(PaymentDetailsFilter::Spark {
+                    htlc_status: Some(statuses),
+                    conversion_refund_needed: None,
+                });
+            }
+            if let Some(tx_hash) = tx_hash {
+                payment_details_filter.push(PaymentDetailsFilter::Token {
+                    conversion_refund_needed: None,
+                    tx_hash: Some(tx_hash),
+                });
+            }
+            let payment_details_filter = if payment_details_filter.is_empty() {
+                None
+            } else {
+                Some(payment_details_filter)
+            };
             let value = sdk
                 .list_payments(ListPaymentsRequest {
                     limit,
@@ -290,7 +313,7 @@ pub(crate) async fn execute_command(
                     type_filter,
                     status_filter,
                     asset_filter,
-                    spark_htlc_status_filter,
+                    payment_details_filter,
                     from_timestamp,
                     to_timestamp,
                     sort_ascending,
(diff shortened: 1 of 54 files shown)


crates/flashnet/src/api.rs
```diff
@@ -0,0 +1,417 @@
...
+            .filter(|ma| ma.enabled)
+            .map(|ma| (ma.asset_identifier, ma.min_amount))
+            .collect::<HashMap<_, _>>();
+        if let Some(&min_in) = min_amounts_map.get(asset_in_address) {
+            if amount_in < min_in {
+                return Err(FlashnetError::Generic(format!(
+                    "Amount in {amount_in} is less than minimum required {min_in} for asset {asset_in_address}",
+                )));
+            }
+            return Ok(());
```
roeierez: We skip here without checking the min_amount_out ?

roeierez (changes requested):
Looks really good!

crates/flashnet/src/api.rs
roeierez: To me it looks like the min_amount_out is not validated in this case. It looks like a bug unless it is not used when min_in is set.

crates/breez-sdk/cli/src/command/mod.rs
```diff
@@ -585,6 +605,38 @@ pub(crate) async fn execute_command(
             print_value(&res)?;
             Ok(true)
         }
+        Command::TokenTransfer {
```
danielgranhao: Nit:
```suggestion
        Command::TransferToken {
```

crates/breez-sdk/cli/src/command/mod.rs
```diff
@@ -585,6 +605,38 @@ pub(crate) async fn execute_command(
...
+                ConvertType::ToBitcoin
+            };
+            let prepare_response = sdk
+                .prepare_convert_token(PrepareConvertTokenRequest {
+                    amount,
+                    token_identifier,
+                    convert_type,
+                })
+                .await?;
+            println!("Prepared transfer: {prepare_response:#?}\n Do you want to continue? (y/n)");
```
danielgranhao:
```suggestion
            println!("Prepared conversion: {prepare_response:#?}\n Do you want to continue? (y/n)");
```

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -238,12 +238,20 @@ pub enum PaymentDetails {
         invoice_details: Option<SparkInvoicePaymentDetails>,
         /// The HTLC transfer details if the payment fulfilled an HTLC transfer
         htlc_details: Option<SparkHtlcDetails>,
+        /// Conversion information if this was a successful conversion
+        conversion_info: Option<ConversionInfo>,
+        /// Conversion refund information if this was a conversion that was refunded
+        conversion_refund_info: Option<ConversionRefundInfo>,
```
danielgranhao: Have you considered merging these 2 into an enum?

crates/breez-sdk/core/src/sdk.rs
```diff
@@ -2372,6 +2762,177 @@ impl BreezSdk {
...
+                if token_balance < amount {
+                    return Err(SdkError::InsufficientFunds);
+                }
+                (token_identifier, BTC_ASSET_ADDRESS)
+            }
+        };
+        Ok((asset_in_address.to_string(), asset_out_address.to_string()))
+    }
+
+    /// Fetches a payment by its covert token identifier.
```
danielgranhao:
```suggestion
    /// Fetches a payment by its convert token identifier.
```

crates/breez-sdk/core/src/sdk.rs
```diff
@@ -860,6 +897,199 @@ impl BreezSdk {
...
+                            "Failed to update payment with conversion info for swap {}: {e:?}",
+                            swap.id
+                        );
+                    }
+                } else {
+                    error!(
+                        "Failed to find payment for swap {} with inbound transfer id {}",
+                        swap.id, swap.inbound_transfer_id
+                    );
+                }
```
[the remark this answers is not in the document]
dangeross: I've removed `sync_conversion_info` from the parallel join, so it syncs after the wallet. I also now break out of the conversion info sync if the payment's not found and store the last updated offset. [feaaf00](https://github.com/breez/spark-sdk/pull/459/commits/feaaf007dea946038f9b7bd97c580113afff4724)

crates/flashnet/src/api.rs
```diff
@@ -0,0 +1,423 @@
...
+        config: FlashnetConfig,
+        spark_wallet: Arc<SparkWallet>,
+        cache_store: Arc<CacheStore>,
+    ) -> Self {
+        Self {
+            config,
+            cache_store,
+            spark_wallet,
+        }
+    }
```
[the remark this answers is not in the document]
dangeross: Removed the `connect`, just using the `new` instead [feaaf00](https://github.com/breez/spark-sdk/pull/459/commits/feaaf007dea946038f9b7bd97c580113afff4724)

crates/breez-sdk/core/src/sdk.rs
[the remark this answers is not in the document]
danielgranhao: Nice. I realize this is an edge case, but if, for any reason, a payment is missing and stays missing, we'll never fully complete syncing conversion info 🤔 

Wondering how we may address that. Could we still store the conversion info in storage even if the corresponding payment is not there?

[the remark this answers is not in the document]
dangeross: We could put it in the storage cache. How likely do you think it is to happen?

crates/breez-sdk/core/src/models/mod.rs
dangeross: Done [0b0b642](https://github.com/breez/spark-sdk/pull/459/commits/0b0b6420eee3fe4a8ecc9bf6c56341c6e1d9409e)

crates/breez-sdk/core/src/sdk.rs
```diff
@@ -736,6 +753,25 @@ impl BreezSdk {
 
         tokio::join!(sync_wallet, sync_lnurl, sync_deposits);
 
+        if sync_type.contains(SyncType::ConversionInfo) {
```
roeierez: My concern here is that we always attempting to approach flashnet API while lots of users won't use it. Slowing the sync for most cases for no reason.
Perhaps we can add a config flag (enable_flashnet) to signal we need to sync with it?

crates/breez-sdk/core/src/sdk.rs
```diff
@@ -1510,6 +1677,205 @@ impl BreezSdk {
...
+        let response = self
+            .flashnet_client
+            .simulate_swap(SimulateSwapRequest {
+                asset_in_address: asset_in_address.clone(),
+                asset_out_address: asset_out_address.clone(),
+                pool_id: pool.lp_public_key,
+                amount_in: request.amount,
+                integrator_bps: None,
+            })
+            .await?;
```
roeierez: Now that we don't need to sync we can treat flashnet as a swapper (swap provider) from the sdk perspective and this code (and some more )could be extracted out to its own swapper trait/service. It it a nice to have for now that will get the flashnet specifics out of the sdk generic code. I wouldn't hold the merge for this feedback.

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -866,9 +887,9 @@ pub struct ListPaymentsRequest {
     pub status_filter: Option<Vec<PaymentStatus>>,
     #[cfg_attr(feature = "uniffi", uniffi(default=None))]
     pub asset_filter: Option<AssetFilter>,
-    /// Only include payments with specific Spark HTLC statuses
+    /// Only include payments matching these payment details filters
```
danielgranhao: Nit:
```suggestion
    /// Only include payments matching at least one of these payment details filters
```

crates/breez-sdk/core/src/sdk.rs
roeierez: I agree but I would rather wait for another implementation before attempting to generalize as a plugin. This has complexities also around integration with the payment list, sync etc... perhaps it is too soon.

roeierez (approved):
LGTM

JssDWt (approved):
LGTM

PR #461, opened 2025-12-02
Add network recommended max fee option

### LINKED ISSUE #460: Allow setting max claim fee to recommended fees in config
To allow for a straightforward way of setting up the max claim fee in the SDK's config, we should allow setting the max fee to the current recommended fees. #451 proposed doing this by moving `recommended_fees` out of the SDK into a standalone function, but this breaks the model of being able to start the SDK without network access. 

Instead, we can add a variant to `Fee` (e.g. `Recommended {leeway_sat_per_vbyte: u64 }`) that makes the SDK internally use the recommended fees to define the max claim fee.

danielgranhao:
Closes #460 

TODO:

- [x] Docs

### DIFF
--- a/docs/breez-sdk/snippets/csharp/Config.cs
+++ b/docs/breez-sdk/snippets/csharp/Config.cs
@@ -17,10 +17,14 @@ namespace BreezSdkSnippets
             config = config with { maxDepositClaimFee = null };
 
             // Set a maximum feerate of 10 sat/vB
-            config = config with { maxDepositClaimFee = new Fee.Rate(satPerVbyte: 10) };
+            config = config with { maxDepositClaimFee = new MaxFee.Rate(satPerVbyte: 10) };
 
             // Set a maximum fee of 1000 sat
-            config = config with { maxDepositClaimFee = new Fee.Fixed(amount: 1000) };
+            config = config with { maxDepositClaimFee = new MaxFee.Fixed(amount: 1000) };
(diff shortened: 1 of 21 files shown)


crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -449,22 +449,47 @@ impl Config {
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 #[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
-pub enum Fee {
+pub enum MaxFee {
     // Fixed fee amount in sats
     Fixed { amount: u64 },
     // Relative fee rate in satoshis per vbyte
     Rate { sat_per_vbyte: u64 },
+    // Current network recommended fee with a leeway in satoshis per vbyte
```
dangeross: Feels like it needs more explanation that it's calculated at request/claim time

danielgranhao:
@roeierez @dangeross Made a new change: reverted the `max_fee` type in the claim error to `Fee` because I noticed while writing docs that providing the variant `NetworkRecommended` when a claim fails isn't very helpful. Now, if the new option is selected, the error contains the fee rate that ends up being used as max, which is much more helpful, as it can be compared to the required fee rate.

Docs were also added.

dangeross (approved):
LGTM

PR #462, opened 2025-12-03
Docs: add custom claim logic snippet

danielgranhao:
This adds a snippet to show that failed claims include a feerate that can be compared with the recommended fees in order to implement a custom claim logic.

Also shows that recommended fees can be used to set fee rate when refunding.

### DIFF
--- a/docs/breez-sdk/snippets/react-native/refunding_payments.ts
+++ b/docs/breez-sdk/snippets/react-native/refunding_payments.ts
@@ -3,7 +3,7 @@ import {
   type ListUnclaimedDepositsRequest,
   type ClaimDepositRequest,
   type RefundDepositRequest,
-  DepositClaimError,
+  DepositClaimError_Tags,
   Fee,
   type DepositInfo,
   Fee_Tags,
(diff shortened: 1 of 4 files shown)


roeierez (approved):
Looks good!

docs/breez-sdk/snippets/react-native/refunding_payments.ts
```diff
@@ -104,6 +105,25 @@ const setMaxFeeToRecommendedFees = () => {
   console.log('Config:', config)
 }
 
+const customClaimLogic = async (sdk: BreezSdk, deposit: DepositInfo) => {
+  // ANCHOR: custom-claim-logic
+  if (deposit.claimError instanceof DepositClaimError.MaxDepositClaimFeeExceeded) {
```
dangeross: I'm wondering if the `DepositClaimError_Tags` need to be used here. Not sure

```suggestion
  if (deposit.claimError.tag === DepositClaimError_Tags.MaxDepositClaimFeeExceeded) {
```

PR #467, opened 2025-12-04
lnurl: store allowed domains in database

JssDWt:
Move allowed domains from config-only to database storage with config bootstrap. Domains are loaded from the database at startup, to avoid doing database calls every time. Newly configured domains through the --domains config variable are inserted in the database if they don't exist yet.


JssDWt:
> There was an LNURL itest failure. I wonder if it's related?

I'm looking at it. Running locally they pass. I'm having a hard time finding relevant parts of the logs. Let's get #468 in.

PR #473, opened 2025-12-08
Emit token payment events

### LINKED ISSUE #418: Token payment events
At the moment, we don't emit token payment events except for `PaymentPending` when sending. We should make the event UX on par with Bitcoin payments.

Also, we don't learn of new incoming token payments until we sync. We could use sparkscan's live updates for this.

dangeross:
Emits the SDK payment events for the synced token transactions. 

I didn't yet consider using sparkscan to receive stream events for token transactions (could be shown as pending/initiate a sync). 

Closes #418


roeierez (commented):
LGTM

roeierez (approved):
LGTM

PR #479, opened 2025-12-12
Optimize HTLC query using ReceiverAndSender role

danielgranhao:
Spark introduced a new preimage request role, which allows us to avoid an additional network request.


roeierez (approved):
LGTM

PR #485, opened 2025-12-16
Async Signer

roeierez:


dangeross (commented):
Maybe not in the scope of the PR, but I'm wondering if we should start on a service level storing the identity public key so we don't have to keep getting it from the signer

crates/spark/src/services/deposit.rs
```diff
@@ -333,9 +341,20 @@ impl DepositService {
...
         );
         // Sign the payload with the identity key
-        let signature = self.signer.sign_message_ecdsa_with_identity_key(&payload)?;
+        let signature = self
+            .signer
+            .sign_message_ecdsa_with_identity_key(&payload)
+            .await?;
 
         // Create the UTXO swap request
+        // Create the UTXO swap request
```
danielgranhao:
```suggestion
```

danielgranhao (approved):
Agree with Ross that it would be nice to minimize calls to get the identity pub key as they may result in network requests depending on the signer implementation. Other than that LGTM

roeierez:
The reason I am not optimizing now is that I want first to get to a stage we have a BreezSigner and ExternalSigner. Then I think to create a SparkSigner adapter implementation where such optimization may be implemented once and not in every service.
My point is that I share your concern but feels premature to do it ATM.

dangeross (approved):

PR #489, opened 2025-12-18
Send payments with token conversion

### LINKED ISSUE #335: Add BTC <> USDB swaps (and vice versa)
* using Flashnet 
* No new API, like we did in Liquid (integrate into send/receive flows)

dangeross:
Adds the ability to pay using different funds than the destination receives using token conversions before the payment.
- Pay Bitcoin spark address/invoice, lightning, onchain using Token funds
- Pay Token spark address/invoice using Bitcoin funds
- Convert Bitcoin from/to Token using self payment to a spark address/invoice

TODO:
- [x] Finalize interface
- [x] ~Structuring of payments with child payments. E.g. Lightning send with [Spark receive, Token send] children~
- [x] Docs

Closes #335

### DIFF
--- a/crates/breez-sdk/core/src/sdk.rs
+++ b/crates/breez-sdk/core/src/sdk.rs
@@ -20,8 +20,8 @@ use breez_sdk_common::{
     rest::RestClient,
 };
 use flashnet::{
-    BTC_ASSET_ADDRESS, ClawbackRequest, ClawbackResponse, ExecuteSwapRequest, FlashnetClient,
-    FlashnetError, GetMinAmountsRequest, ListPoolsRequest, PoolSortOrder, SimulateSwapRequest,
+    ClawbackRequest, ClawbackResponse, ExecuteSwapRequest, FlashnetClient, FlashnetError,
+    GetMinAmountsRequest, ListPoolsRequest, PoolSortOrder, SimulateSwapRequest,
 };
 use lnurl_models::sanitize_username;
 use spark_wallet::{
@@ -1793,18 +1793,9 @@ impl BreezSdk {
         &self,
         request: FetchTokenConversionLimitsRequest,
     ) -> Result<FetchTokenConversionLimitsResponse, SdkError> {
-        self.ensure_spark_private_mode_initialized().await?;
-        let (asset_in_address, asset_out_address) = match request.conversion_type {
-            TokenConversionType::FromBitcoin => (
-                BTC_ASSET_ADDRESS.to_string(),
-                request.token_identifier.ok_or(SdkError::InvalidInput(
-                    "Token identifier is required for from Bitcoin conversion".to_string(),
-                ))?,
-            ),
-            TokenConversionType::ToBitcoin {
-                from_token_identifier,
-            } => (from_token_identifier, BTC_ASSET_ADDRESS.to_string()),
-        };
+        let (asset_in_address, asset_out_address) = request
+            .conversion_type
+            .as_asset_addresses(request.token_identifier.as_ref())?;
         let min_amounts = self
             .flashnet_client
             .get_min_amounts(GetMinAmountsRequest {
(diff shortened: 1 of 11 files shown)


crates/breez-sdk/core/src/sdk.rs
```diff
@@ -2127,8 +2211,162 @@ impl BreezSdk {
...
+        if matches!(
+            token_conversion_options.conversion_type,
+            TokenConversionType::FromBitcoin
+        ) {
+            let _ = self
+                .sync_trigger
+                .send(SyncRequest::no_reply(SyncType::WalletState));
+        }
+        // Wait for the received token conversion payment to complete
+        let payment = self
```
roeierez: This 2 steps flow (conversion, waiting for receive and then send) may fail in the middle.
How would this affect the user if the conversion succeed but then sending the payment bellow fails?

roeierez (changes requested):
Looking good! 
I see a point in exposing an option to list payments in "aggregated" mode where conversion payments are "sub items" of their parent but since we have all the info in the db I think we should first use it as is now (which is also reasonable) and can discuss/add the aggregated feature afterwards.

crates/breez-sdk/core/src/sdk.rs
```diff
@@ -98,6 +106,7 @@ const BREEZ_SYNC_SERVICE_URL: &str = "https://datasync.breez.technology";
 const BREEZ_SYNC_SERVICE_URL: &str = "https://datasync.breez.technology:442";
 
 const SYNC_PAGING_LIMIT: u32 = 100;
+const DEFAULT_TOKEN_CONVERSION_MAX_SLIPPAGE_BPS: u32 = 50;
```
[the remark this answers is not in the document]
dangeross: I've seen when testing different conversions on mainnet/regtest that it can slip more than 50bps. It depends on the token liquidity

crates/flashnet/src/models.rs
```diff
@@ -411,6 +411,93 @@ pub struct Pool {
...
+            &self.asset_a_address
+        };
+        let amount_out = if asset_out_address == BTC_ASSET_ADDRESS {
+            amount_out.saturating_add(63) & !63
+        } else {
+            amount_out
+        };
+
+        // Add slippage buffer to amount_out first
+        let amount_out_with_slippage = amount_out
```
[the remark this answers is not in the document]
dangeross: Will test that. When iterating through the reverse calculation, I was finding float arithmetic more consistent to the Flashnet calculations

crates/breez-sdk/core/src/sdk.rs
dangeross: The send payment request will fail if we either timeout waiting for the conversion receive or the final send payment fails. The user will end up with the sent payment for the token conversion (with token conversion details set) and a received payment for the token conversion (eventually, if the waiting times out)

dangeross:
I will follow up this with another PR adding itests

crates/breez-sdk/core/src/sdk.rs
```diff
@@ -2130,8 +2216,164 @@ impl BreezSdk {
...
+        .await?;
+
+        Ok(response)
+    }
+
+    async fn send_payment_internal(
+        &self,
+        request: &SendPaymentRequest,
+    ) -> Result<SendPaymentResponse, SdkError> {
+        // Perform a token conversion before sending the payment
```
danielgranhao: Leftover comment?

docs/breez-sdk/src/SUMMARY.md
```diff
@@ -28,6 +28,7 @@
 - [End-user fees](guide/end-user_fees.md)
 - [Handling tokens](guide/tokens.md)
   - [Token payments](guide/token_payments.md)
+  - [Token conversion](guide/token_conversion.md)
```
danielgranhao: nit: maybe `Converting tokens`?

danielgranhao (commented):
It's looking really good! A few comments.

crates/breez-sdk/core/src/sdk.rs
```diff
@@ -2506,6 +2734,50 @@ impl BreezSdk {
         timeout_res?
     }
 
+    async fn merge_payment_metadata(
```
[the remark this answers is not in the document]
dangeross: Yeah, or preferably as it's own module, but there's a bit of back-and-forth between the SDK which needs refactoring. Will address separately if that's ok

PR #494, opened 2025-12-20
Implement LUD-21 LNURL-verify

### LINKED ISSUE #278: LNURL Verify

andrerfneves:
## Summary

  Implements [LUD-21](https://github.com/lnurl/luds/blob/luds/21.md) which allows services to check whether a Lightning invoice has been paid by querying the LNURL server.

  **What is LUD-21?**
  When a wallet pays an LNURL invoice, the payer may want to verify that the payment was received (e.g., for proof of payment in e-commerce). LUD-21 defines a `verify` URL that returns the payment status and preimage.

  ## Changes

  ### Database Schema
  Added new table `lnurl_pay_invoices`

  ### Invoice Creation (`/lnurlp/{user}/invoice`)
  When creating an invoice, we now:
  1. Store the `res.id` from `create_lightning_invoice()` as `lightning_receive_id`
  2. Store the `res.invoice` as `bolt11_invoice`
  3. Return a `verify` URL in the response

  ### Verify Endpoint (`/lnurlp/{user}/verify?pr=lnbc...`)
  1. Parse the invoice from `pr` query param to extract `payment_hash`
  2. Look up the stored `lightning_receive_id` from the database
  3. Call `wallet.fetch_lightning_receive_payment()` to get current payment status
  4. Return LUD-21 compliant response with `settled` and `preimage`

##  Assumptions

  1. Payment is settled when preimage exists: We determine settled: true by checking if payment.payment_preimage.is_some(). This assumes Spark only populates the preimage field once the payment is fully received.
  2. Verify requires the full invoice: Per LUD-21, the verify endpoint requires the pr (invoice) as a query parameter. We parse it to extract the payment hash for lookup.
  3. Domain/username validation: We verify the invoice belongs to the user in the URL path to prevent cross-user invoice status leakage.

## Test Plan

- Create an invoice via /lnurlp/{user}/invoice?amount=10000
- Verify the response includes a verify URL
- Call verify endpoint before payment → returns {"settled": false, "preimage": null, ...}
- Pay the invoice via Lightning Network
- Call verify endpoint after payment → returns {"settled": true, "preimage": "...", ...}
- Verify that querying with an invoice from a different user returns an error
- Verify that querying with an unknown invoice returns "invoice not found"

### DIFF
--- a/crates/breez-sdk/breez-itest/tests/lnurl.rs
+++ b/crates/breez-sdk/breez-itest/tests/lnurl.rs
@@ -8,6 +8,7 @@ use rand::RngCore;
 use rstest::*;
 use tempdir::TempDir;
 use tracing::{debug, info};
+use hex;
 
 // ---------------------
 // Fixtures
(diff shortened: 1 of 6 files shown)


andrerfneves:
@JssDWt, thanks for the feedback! You're right - the original approach of querying the server's wallet doesn't work since it's not the user's wallet.

### What I've done

Basically replicated the `zap receipt` pattern but for LNURL-pay invoices / LUD-21 verify:

**Database changes (lnurl_pay_invoices table):**
- Added `preimage` - stores the payment preimage once we know the invoice is paid
- Added `is_privacy_mode` - captures whether the user was in private mode when the invoice was created (need this at creation time since user settings can change)

**How I determine privacy mode:**
We check `user.nostr_pubkey.is_some()` - this works because the client only sends their nostr pubkey during registration when `spark_user_settings.private_enabled = true` (see `register_lightning_address_internal` in sdk.rs around line 2559).

Let me know if the above is incorrect as it’s a key conditional.

**Non-private mode flow (server monitors):**
- New `lnurl_pay.rs` module - basically a copy of `zap.rs` but simpler (no nostr stuff)
- When invoice is created and user is non-private, server subscribes to their events
- When payment comes in, server grabs preimage from SSP and stores it
- Verify endpoint returns `settled: true` with preimage

**Private mode flow (client notifies):**
- New endpoint: `POST /lnurlpay/{pubkey}/invoice/{payment_hash}/paid`
- Client calls this with the preimage after they receive payment
- Server validates preimage matches payment_hash (sha256) and stores it
- Verify endpoint returns `settled: true` with preimage

**Client-side (sdk.rs / lnurl.rs):**
- Added `mark_invoice_paid` to `LnurlServerClient` trait
- Added `process_pending_lnurl_payments` background task that runs on startup and after metadata sync
- Added `lnurl_verify_notified` field to `LnurlReceiveMetadata` to track which payments we've already notified the server about (avoids redundant API calls, similar to how zap receipts track `nostr_zap_receipt`)

**Integration tests:**
- `test_07_lnurl_verify_non_private_mode` - Tests server-side monitoring flow
- `test_08_lnurl_verify_private_mode` - Tests client notification flow

Both tests verify the `/verify` endpoint returns correct `settled` status and validate that `sha256(preimage) == payment_hash`.

Let me know if this is more inline with what your team has in mind internally.

andrerfneves:
Also lmk if you want me to remove these code comments

crates/breez-sdk/lnurl/src/routes.rs
```diff
@@ -626,7 +632,8 @@ where
                 _ => (None, false),
             };
 
-        let desc_hash = if let Some(event) = &params.nostr {
+        // Compute description hash and capture metadata for LNURL-verify
```
JssDWt: Let's remove this comment. This is also used elsewhere.

roeierez:
I see this replacement PR was opened by @JssDWt  https://github.com/breez/spark-sdk/pull/606
Probably took some inspiration from this one (thank you @andrerfneves ) but we needed to move faster with this.
Closing this one.

PR #501, opened 2025-12-29
Token outputs optimization

### LINKED ISSUE #469: Token outputs optimization
Implement token outputs optimization to match the JS implementation. Goal is to consolidate outputs to avoid being unable to send (there is a max number of inputs per TX)

danielgranhao:
Closes #469.

Changes:
- Implement token optimization - i.e. send to self to consolidate outputs
- Run optimization automatically in the background
- Run optimization when payments would fail due to too many selected inputs
- Fix bug that would lead to wrong balance when outputs of a certain token exceeded 500 (no pagination on token querying)
- Prevent Spark self-payments. We didn't prevent this in the past, but we didn't support it properly either. Bitcoin self-payments would result in the list containing only outgoing payments without a matching incoming payment. Token self-payments wouldn't appear as payments at all. As I don't currently see a use case for self-payments, I think it's simpler and appropriate to simply prevent them.

Notes:
- I didn't expose configuration or control of the token optimization process at the Breez SDK level. Unlike leaf optimization, token optimization has a low likelihood of affecting payments due to it being a quicker process that needs to occur less often. I propose we wait before exposing it to see if we find a compelling reason to do it.

TODO:
- [x] Add reservation purpose to token store to make sure balance isn't affected by ongoing optimizations
- [x] Make sure payment sync logic isn't affected by self sends

### DIFF
--- a/crates/flashnet/src/api.rs
+++ b/crates/flashnet/src/api.rs
@@ -437,6 +437,7 @@ impl FlashnetClient {
                         spark_invoice: None,
                     }],
                     None,
+                    None,
                 )
                 .await?
                 .hash


crates/spark-wallet/src/config.rs
```diff
@@ -67,8 +74,13 @@ impl SparkWalletConfig {
...
                 tokens_config: Self::default_tokens_config(),
-                optimization_options: OptimizationOptions::default(),
-                auto_optimize_enabled: true,
+                leaf_optimization_options: LeafOptimizationOptions::default(),
+                leaf_auto_optimize_enabled: true,
+                token_outputs_optimization_options: TokenOutputsOptimizationOptions {
+                    min_outputs_threshold: 50,
+                    auto_optimize_interval: Some(Duration::from_secs(60 * 2)),
+                },
+                self_payment_allowed: false,
```
roeierez: How this will work when we introduce self payments for token conversion purpose? Why not handle self payments properly instead?

roeierez (commented):
Still reviewing but wanted to pass an early feedback.

crates/spark-wallet/src/config.rs
dangeross: > For token conversions, there are 2 separate transfers/txs and no self-payments. Could I be missing something?

This is right

crates/spark/src/services/tokens.rs
```diff
@@ -500,6 +525,110 @@ impl TokenService {
...
+                "min_outputs_threshold must be greater than 1".to_string(),
+            ));
+        }
+
+        info!(
+            "Optimizing token outputs starting (optional token identifier: {:?})",
+            token_identifier
+        );
+
+        self.token_output_service.refresh_tokens_outputs().await?;
```
roeierez: Should we explicitly refresh_tokens_outputs here? Given the fact it is already refreshed every sync I wonder if this is necessary.

PR #502, opened 2025-12-30
Expose token tx type

### LINKED ISSUE #481: Expose token transaction type in Breez SDK
Minting or burning a token results in an incoming or outgoing payment, but the payment does not indicate that it corresponds to a mint/burn operation. 

We should expose this information as well as allow filtering by type in `list_payments`.

danielgranhao:
Closes #481.

Changes:
- Expose token tx type (Transfer vs Mint vs Burn) on payment details
- Allow filtering by token tx type when listing payments

To make sure existing payments get the correct tx type, a migration was added that resets the token sync offset, causing token payments to be synced from scratch. 

TODO: 

- [x] Docs
- [x] Rebase

### DIFF
--- a/crates/breez-sdk/core/src/persist/postgres.rs
+++ b/crates/breez-sdk/core/src/persist/postgres.rs
@@ -684,6 +684,10 @@ impl PostgresStorage {
                 "CREATE INDEX IF NOT EXISTS idx_payment_details_lightning_invoice ON payment_details_lightning(invoice)",
                 "CREATE INDEX IF NOT EXISTS idx_payment_metadata_parent ON payment_metadata(parent_payment_id)",
             ],
+            // Migration 4: Add tx_type to token payments
+            &[
+                "ALTER TABLE payment_details_token ADD COLUMN tx_type TEXT NOT NULL DEFAULT 'transfer'",
+            ],
         ]
     }
 }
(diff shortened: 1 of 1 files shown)


dangeross (commented):
Looks good, needs a rebase

crates/breez-sdk/core/src/persist/sqlite.rs
```diff
@@ -553,20 +572,23 @@ impl Storage for SqliteStorage {
...
                      ON CONFLICT(payment_id) DO UPDATE SET 
                         metadata=excluded.metadata,
                         tx_hash=excluded.tx_hash,
+                        tx_type=excluded.tx_type,
                         invoice_details=COALESCE(excluded.invoice_details, payment_details_token.invoice_details)",
                     params![
                         payment.id,
                         serde_json::to_string(&metadata)?,
                         tx_hash,
+                        serde_json::to_string(&tx_type)?,
```
dangeross: Is there a particular reason to use serde_json instead of Display/FromStr. Don't we end up with quote wrapped strings?

danielgranhao: Mainly consistency with how other types are (de)serialized

dangeross: I think structs are converted to json, but most enums are just strings (see PaymentType, PaymentStatus, PaymentMethod)

JssDWt (commented):
Just a small thing that comes from leaving this PR up for too long. Otherwise LGTM.

Issue #497, opened 2025-12-22
Reduce dependencies


TheBlueMatt:
I was looking at using this crate as a part of a larger project, but noticed that the dependency list is more than a little absurd. to the point that I don't really feel comfortable trusting that many third party developers with the money of people who might use my software. Obviously dependencies are not, in principle, a bad thing, and save lots of time, but dependencies where the project is almost exclusively built by a single third party with limited review is just an opportunity for that third-party (or a threat actor who compromises the computer/Github account of that third party) to steal the money of everyone who uses `breez-sdk-spark`.

Just scrolling through the list of `breez-sdk-spark` dependencies from top to bottom, there's lots of crates that look like they get limited if any outside contribution, things like `dotenvy`, `ecies` (which is a trivial algorithm you can write yourself!), `nostr`, `qrcode-rs`, `shellwords`, `tokio-with-wasm` (which certainly doesn't need to be built for non-wasm targets!), `web-time` (you can just use `std` time when building with `wasm-pack`!), `x509-parser`, and `base32` (come on, just copy the code...) probably all don't make sense in a Bitcoin wallet, and that's not even looking at recursive dependencies or other sub-crates here.

While cleaning these up all at once may be a chunk of work (or may not be, depending on how these are used!), making progress on this over time while restricting new dependencies is almost certainly a good idea.

roeierez:
We will go over our dependencies and start removing whatever makes sense.

roeierez:
> One fairly-easy way to cut down a lot of dependencies is to switch from `reqwest` to `bitreq` (https://docs.rs/bitreq/latest/bitreq/). Its maintained as a part of `rust-bitcoin`'s `corepc` crate and relies on a ton fewer dependencies (basically only `tokio` and `rustls`) and avoids anti-features like IDNs. You'll likely still have some transitive dependencies on `reqwest`, but hopefully over time all the `rust-bitcoin` world HTTP clients move over and you lose a lot of deps.

@TheBlueMatt We have a PR that replaces reqwest with bitreq. While it works fine natively I suspect (still didn't test) it will not work on Android. The reason is that with `reqwest` we had to use "native-tls" feature and I don't see such feature in bitreq which uses rust-tls ( I don't think existing feature `rustls-native-certs `has the same effect). Are you aware of such issue?

roeierez:
@TheBlueMatt I have tested with rust-tls and with your PR and native-tls and both works.
[...]

PR #503, opened 2026-01-02
Remove 3 dependencies from breez-sdk-core

roeierez:
Starting to address: https://github.com/breez/spark-sdk/issues/497
Went first for some quick wins by removing 3 dependencies from the workspace(were actually unused in  breez-sdk-core) and declaring them only where they are used (internal testing and internal tool):
* dotenvy
* qrcode-rs
* shellwords


roeierez:
> I think these crates can be kept in the workspace? The only change is they are removed from `breez-sdk-core`

I am leaning towards clean the workspace and leave there only crates that are shared and used as a runtime dependency. for tools and testing I prefer to declare them in specific crates.

PR #504, opened 2026-01-02
Remove reqwest

roeierez:
Partially addresses: https://github.com/breez/spark-sdk/issues/497

This PR replaces reqwest with bitreq. Since bitreq lacks complete WASM support (work was started but remains incomplete), we're taking a phased approach:
Phase 1 (this PR): Keep reqwest as a dependency only for WASM targets; remove it entirely from non-WASM builds.
Phase 2 (future PR): Evaluate and implement a better long-term solution for removing reqwest from WASM as well.
To abstract away the HTTP client implementation details, this PR updates:

* `spark`: Adds two HttpClient implementations (Bitreq and Reqwest), with Reqwest available only for WASM
* `breez-sdk-common`: Adds two RestClient implementations (Bitreq and Reqwest), with Reqwest available only for WASM
* Move all url parsing code from reqwest to url crate ( bitreq has url parsing functionality but it is internal atm).
* Higher-level crates now depend on the breez-sdk-common abstraction, agnostic to the underline library in use.

It removes arround ~120 transient dependencies.

TODO:

- [x] Test lnurl client (affected by this change).
- [x] Test android/ios build (we had some issues with tls in the past.


TheBlueMatt:
We shipped `bitreq 0.3` at https://github.com/rust-bitcoin/corepc/pull/459 which cleans up a number of issues and improves performance in async. bitreq 0.2 was mostly a vibe-coded adaptation of the minreq crate, bitreq 0.3 is now a real crate with reasonable review and lots of cleanups.

crates/breez-sdk/common/src/rest/rest_client.rs
```diff
@@ -1,38 +1,30 @@
...
+/// Response from a REST request
 pub struct RestResponse {
     pub status: u16,
     pub body: String,
 }
 
 impl RestResponse {
     pub fn is_success(&self) -> bool {
-        self.status >= 200 && self.status < 300
+        (200..300).contains(&self.status)
```
JssDWt: I don't think iterating a list of numbers is an improvement over testing inequalities.

crates/breez-sdk/common/src/rest/rest_client.rs
```diff
@@ -122,8 +225,8 @@ impl RestClient for ReqwestRestClient {
         headers: Option<HashMap<String, String>>,
         body: Option<String>,
     ) -> Result<RestResponse, ServiceConnectivityError> {
-        debug!("Making DELETE request to: {url}");
-        let mut req = self.client.delete(url).timeout(REQUEST_TIMEOUT);
+        tracing::debug!("Making DELETE request to: {url}");
+        let mut req = self.client.delete(&url);
```
JssDWt: The timeout was removed here, is that intended? Also for get and post.

crates/spark/src/ssp/graphql/http.rs
```diff
@@ -0,0 +1,299 @@
+//! HTTP client abstraction for GraphQL and REST requests.
```
JssDWt: The HttpClient in this file seems to do the exact same thing as the RestClient in common. Should we combine them into a small new crate?

roeierez: Yes, that's the way but since this is a pure refactoring of replacing one library in another I prefer to do it as a follow up PR.

crates/breez-sdk/common/src/rest/rest_client.rs
```diff
@@ -53,37 +43,144 @@ pub trait RestClient: Send + Sync {
...
+    }
+
+    async fn post_request(
+        &self,
+        url: String,
+        headers: Option<HashMap<String, String>>,
+        body: Option<String>,
+    ) -> Result<RestResponse, ServiceConnectivityError> {
+        tracing::debug!("Making POST request to: {url}");
+        let mut req = bitreq::post(&url).with_timeout(30);
```
erdemyerebasmaz: **[Suggestion]** Hardcoded timeout

The `REQUEST_TIMEOUT` constant is defined at line 5 but POST uses `30` directly here while GET uses the constant. Is this intentional?

```rust
let mut req = bitreq::post(&url).with_timeout(REQUEST_TIMEOUT);
```

erdemyerebasmaz (commented):
> 🧪 Experimental PR review using Claude Code.

---

### Summary
Replaces reqwest with bitreq for non-WASM targets to reduce ~120 transitive dependencies. Creates HTTP client abstractions in both `breez-sdk-common` and `spark` crates.

### Notes
- All CI checks pass
- Existing review comments have been addressed

**Recommendation:** APPROVE

This is a solid refactoring that achieves the dependency reduction goal. The parallel abstractions in `common/rest` and `spark/http` are noted for future consolidation.

---

> ⚠️ **Review limitation:** `Cargo.lock` exceeds token limits and was not reviewed in detail.

crates/spark/src/ssp/graphql/http.rs
```diff
@@ -0,0 +1,299 @@
...
+
+#[cfg(all(target_family = "wasm", target_os = "unknown"))]
+impl ReqwestHttpClient {
+    pub fn new(user_agent: Option<String>) -> Self {
+        let mut builder = reqwest::Client::builder();
+        if let Some(ua) = user_agent {
+            builder = builder.user_agent(ua);
+        }
+        Self {
+            client: builder.build().expect("Failed to create reqwest client"),
```
erdemyerebasmaz: **[Suggestion]** Consider returning `Result` instead of panicking

`expect()` here will panic if the client fails to build. This is the WASM-only code path, so it only affects browser builds. Is there a scenario where this could fail in practice?

```rust
// Alternative: propagate the error
pub fn new(user_agent: Option<String>) -> Result<Self, HttpError> {
    let client = builder.build()
        .map_err(|e| HttpError::Other(e.to_string()))?;
    Ok(Self { client })
}
```

crates/spark/Cargo.toml
```diff
@@ -33,7 +28,7 @@ macros.workspace = true
 prost.workspace = true
 prost-types.workspace = true
 rand.workspace = true
-reqwest.workspace = true
+reqwest = { workspace = true, optional = true }
```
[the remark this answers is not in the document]
roeierez: I left them for the integration tests and wasm but I am convinced now that we can use bitreq also in the integration tests so removed all optional dependencies and left the reqwest only in common, spark and flashnet crates scoped for wasm only.
I will follow up with a PR for "platform-utils" crate that contains such kind of different platform implementation and will add a generic simple http client there where all crates can reference.

danielgranhao (approved):
LGTM - there's a CI failure in test_06_client_side_zap_receipt. Most likely is a flakyness issue, but best to double check given the changes here could plausibly affect it.

TheBlueMatt:
It was rewritten and exported in the latest version from today :).

crates/spark-itest/src/fixtures/bitcoind.rs
```diff
@@ -245,23 +246,28 @@ impl BitcoindFixture {
             "params": params,
         });
 
+        let auth_header = make_basic_auth_header(&self.rpcuser, &self.rpcpassword);
+        let body = serde_json::to_string(&request)?;
+
+        let mut headers = HashMap::new();
+        headers.insert("Authorization".to_string(), auth_header);
```
danielgranhao: Nit: Instead of `make_basic_auth_header`, we could have an `add_basic_auth_header` that mutates the headers directly.

crates/internal/src/command/deposit.rs
```diff
@@ -171,14 +172,21 @@ async fn get_transaction(
...
-    let hex = response.text().await?;
-    let tx = deserialize_hex(&hex)?;
+    let auth_header = make_basic_auth_header(
+        mempool_config.username.as_deref().unwrap_or(""),
+        mempool_config.password.as_deref().unwrap_or(""),
+    );
+
+    let mut headers = HashMap::new();
+    headers.insert("Authorization".to_string(), auth_header);
+
```
danielgranhao: We should just skip inserting the header if username/password are empty

crates/internal/src/command/deposit.rs
```diff
@@ -188,19 +196,24 @@ async fn broadcast_transaction(
...
-    let response = request.send().await?;
-    let text = response.text().await?;
-    let txid = Txid::from_str(&text).map_err(|_| {
+    let auth_header = make_basic_auth_header(
+        mempool_config.username.as_deref().unwrap_or(""),
+        mempool_config.password.as_deref().unwrap_or(""),
+    );
+
+    let mut headers = HashMap::new();
+    headers.insert("Authorization".to_string(), auth_header);
```
danielgranhao: Same

danielgranhao (commented):
Looks good. Just some minor details

danielgranhao (approved):
LGTM

PR #509, opened 2026-01-05
docs(timelock): clarify intentional lower-16-bit extraction

pulkit-999:
- Add TIMELOCK_MASK constant for clarity.
- Add comment explaining that only lower 16 bits are used for timelock.
- Maintain exact same truncation behavior as before.


danielgranhao (commented):
Please format using `cargo fmt`. Otherwise LGTM. The integration test failure is not related to the changes.

PR #511, opened 2026-01-05
fix: counter swap resulting in pending Payment

danielgranhao:
If a leaf swap ( during optimization or at payment time) fails to have the counter swap transfer claimed, the counter swap will end up being claimed later in the background by `claim_pending_transfers`. 

Given that claiming all pending transfers can result in `TransferClaimed` events, we would create a Payment for the transfer and persist it. The payment would be pending because we were also missing updating the transfer status after claiming

Fixes in this PR:
- Converting a transfer into a Payment now fails if the transfer is associated with a swap (failed conversions are ignored when handling wallet events)
- We now update all claimed transfer status to `Completed`
- I noticed we would only claim these counter swap transfers when restarting the stream, so now we check for pending transfers on sync

### DIFF
--- a/crates/spark-wallet/src/wallet.rs
+++ b/crates/spark-wallet/src/wallet.rs
@@ -718,6 +718,11 @@ impl SparkWallet {
             self.maybe_start_optimization();
         }
 
+        for transfer in &transfers {
+            self.event_manager
+                .notify_listeners(WalletEvent::TransferClaimed(transfer.clone()));
+        }
+
         Ok(transfers)
     }
 
@@ -973,6 +978,8 @@ impl SparkWallet {
     pub async fn sync(&self) -> Result<(), SparkWalletError> {
         self.tree_service.refresh_leaves().await?;
         self.token_output_service.refresh_tokens_outputs().await?;
+        // Claiming here any transfers that may have been missed in the event stream handling (e.g. counter swap transfers are intentionally not claimed there)
+        self.claim_pending_transfers().await?;
         Ok(())
     }
 
@@ -1902,12 +1909,12 @@ impl BackgroundProcessor {
                     "Claimed {} pending transfers on stream reconnection",
                     transfers.len()
                 );
+                if !transfers.is_empty() {
+                    self.maybe_start_optimization();
+                }
                 for transfer in &transfers {
                     self.event_manager
                         .notify_listeners(WalletEvent::TransferClaimed(transfer.clone()));
-                }
-                if !transfers.is_empty() {
-                    self.maybe_start_optimization();
                 }
             }
             Err(e) => {


crates/spark-wallet/src/wallet.rs
```diff
@@ -1548,6 +1555,11 @@ async fn claim_pending_transfers(
         );
     }
 
+    let mut transfers = transfers;
```
roeierez: This will solve some flakiness in the integration tests.  👍

roeierez (approved):
Dropped one question. Other than that LGTM

crates/spark-wallet/src/wallet.rs
```diff
@@ -973,6 +978,8 @@ impl SparkWallet {
     pub async fn sync(&self) -> Result<(), SparkWalletError> {
         self.tree_service.refresh_leaves().await?;
         self.token_output_service.refresh_tokens_outputs().await?;
+        // Claiming here any transfers that may have been missed in the event stream handling (e.g. counter swap transfers are intentionally not claimed there)
+        self.claim_pending_transfers().await?;
```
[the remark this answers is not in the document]
danielgranhao: The comment there tries to answer that question, but clearly failed :) 

Counter swap transfers are skipped and not claimed in the real-time event handling logic. If they fail to get claimed synchronously during the swap process, they will only be claimed on reconnect. This helps by ensuring we claim them before the next restart or stream reconnection.

Please share if you have a proposal for how to make the comment clearer.

crates/breez-sdk/core/src/models/adaptors.rs
```diff
@@ -129,6 +130,19 @@ impl From<SparkInvoiceDetails> for SparkInvoicePaymentDetails {
...
+            TransferType::Swap,
+            TransferType::PrimarySwapV3,
+        ]
+        .contains(&transfer.transfer_type)
+        {
+            debug!("Tried to convert swap-related transfer to payment. Transfer: {transfer:?}");
+            return Err(SdkError::Generic(
+                "Swap-related transfers are not considered payments".to_string(),
+            ));
+        }
```
dangeross: How does this effect the payment sync? Wouldn't the swap related transfer break the loop? Maybe I'm missing something
https://github.com/breez/spark-sdk/blob/6468f1b0651513c1fdba72678da72928408b5e4a/crates/breez-sdk/core/src/sync.rs#L78-L80

danielgranhao: When we query transfers, we query only non swap related ones, so we shouldn't get any there.

dangeross: Thanks for clarifying

dangeross (approved):

PR #514, opened 2026-01-06
fix: flawed connector refund txs creation

danielgranhao:
This fixes 2 bugs that were caught in CI when the SOs started running stricter validation on regtest:
- Anchor output not included in CPFP connector refund tx 
- Wrong output value used in direct connector refund tx. This had been fixed for regular refund txs in #311, but was wrong on connector txs


dangeross (approved):
LGTM

PR #519, opened 2026-01-07
itest: add payment sync test

### LINKED ISSUE #441: Add synchronization integration tests
Currently, all tests create new identities and new payments. This doesn't give the opportunity to test that we can sync/recover older payments correctly.

We should set up a wallet with some payments (ideally at least one of each kind) and add a test where we just sync and verify we can recover the payments.

danielgranhao:
Closes #441


roeierez (approved):
LGTM

PR #523, opened 2026-01-08
LNURL-pay with conversion options

dangeross:
Feeds token conversion options into LNURL-pay flow


danielgranhao (approved):
LGTM (edit, need to check failing snippet CI)

PR #535, opened 2026-01-11
swift: plist min OS version fix

dangeross:
Changes the minimum OS version to be inline with the swift package/podfile. This fixes an issue when uploading the iOS application for distribution the process fails with a validation error, invalid bundle

~Draft while waiting for user testing~


dangeross:
@claude short review of this PR

danielgranhao (approved):

roeierez (approved):

PR #541, opened 2026-01-12
fix: handle broadcast receiver lag to prevent event processing termination

pulkit-999:
Previously, the event processing loop would silently exit when the broadcast
receiver lagged behind, causing permanent failure of the event system. Now
properly handles RecvError::Lagged by logging a warning and continuing to
process subsequent events.

### DIFF
--- a/crates/spark-wallet/src/wallet.rs
+++ b/crates/spark-wallet/src/wallet.rs
@@ -1855,23 +1855,38 @@ impl BackgroundProcessor {
     }
 
     async fn process_events(&self, mut event_stream: broadcast::Receiver<SparkEvent>) {
-        while let Ok(event) = event_stream.recv().await {
-            debug!("Received event: {event}");
-            trace!("Received event: {event:?}");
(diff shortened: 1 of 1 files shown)


roeierez (approved):
LGTM

crates/spark-wallet/src/wallet.rs
```diff
@@ -1855,23 +1855,39 @@ impl BackgroundProcessor {
...
+                    debug!("Processed event: {event}");
+
+                    if let Err(e) = result {
+                        error!("Error processing event: {e:?}");
+                    }
+                }
+
+                Err(RecvError::Lagged(skipped)) => {
+                    error!("Event stream lagged, skipped {} messages", skipped);
+                    // Continue processing - THIS IS THE CRITICAL FIX
```
JssDWt: nit: Please remove this comment

JssDWt:
Thanks, can you please squash your changes into a single commit?

PR #544, opened 2026-01-13
Drain all funds

### LINKED ISSUE #499: Allow drain all funds
* when sending to a btc address 
* when using lnurl/lightning address (this is feasible only if we can overpay fees in spark. If not we need to open a separate issue and discuss with the spark team)

danielgranhao:
Closes #499.

Changes:
- Add drain feature to send entire Bitcoin balance
- Introduce new PayAmount enum with Bitcoin, Token, and Drain variants. For LNURL, BitcoinPayAmount is used - same but without Token variant.
- **Breaking change**: Replace amount/token_identifier fields with pay_amount in PrepareSendPaymentRequest and PrepareLnurlPayRequest

LNURL implementation notes:
- For LNURL, fee overpayment is used to make sure we get to 0 balance. A simple approach is used: we get an invoice for the total balance, check fee, then get the actual invoice we intend to pay for (balance - fee). I did some experiments, and saw fee overpayments of at most 1 sat or less than 0.01% for big amounts (e.g. 1 bitcoin). 
- I added a protection/sanity check that prevents fee overpayment amount to be higher than the needed fee. In any case, the user also reviews the total fee before committing to the drain.

### DIFF
--- a/crates/breez-sdk/core/src/models/mod.rs
+++ b/crates/breez-sdk/core/src/models/mod.rs
@@ -853,9 +853,8 @@ pub struct ReceivePaymentResponse {
 
 #[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
 pub struct PrepareLnurlPayRequest {
-    /// The amount to send. Use `PayAmount::Drain` to drain all funds.
-    /// Note: `PayAmount::Token` is not supported for LNURL (returns error).
-    pub pay_amount: PayAmount,
+    /// The amount to send. Use `BitcoinPayAmount::Drain` to drain all funds.
+    pub pay_amount: BitcoinPayAmount,
     pub pay_request: LnurlPayRequestDetails,
     #[cfg_attr(feature = "uniffi", uniffi(default=None))]
     pub comment: Option<String>,
@@ -869,7 +868,7 @@ pub struct PrepareLnurlPayRequest {
 #[derive(Debug, Clone)]
 #[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
 pub struct PrepareLnurlPayResponse {
-    pub pay_amount: PayAmount,
+    pub pay_amount: BitcoinPayAmount,
     pub comment: Option<String>,
     pub pay_request: LnurlPayRequestDetails,
     /// The fee in satoshis. For drain operations, this represents the total drain fee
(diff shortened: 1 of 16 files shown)


crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -740,6 +742,21 @@ impl SendOnchainSpeedFeeQuote {
     }
 }
 
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct EstimateOnchainSendFeeQuotesRequest {
```
JssDWt: Can we improve this name?

crates/breez-sdk/core/src/sdk.rs
```diff
@@ -1529,17 +1543,191 @@ impl BreezSdk {
...
+                return Err(SdkError::Generic(format!(
+                    "Fee overpayment ({overpayment} sats) exceeds allowed maximum ({max_allowed_overpayment} sats)"
+                )));
+            }
+
+            if overpayment > 0 {
+                tracing::info!(
+                    overpayment_sats = overpayment,
+                    drain_fee_sats = drain_fee,
+                    current_fee_sats = current_fee,
```
JssDWt: Do these end up in the log?

JssDWt:
I have one remark about the interface. There are now many options/combinations where you can get an error back. Reading the code there are some payment requests that allow drain and some not. There are some payment requests that allow no amount, and some not. It's starting to feel like a bit of a minefield, where the caller has to know about all these combinations in order to avoid errors. I wonder if we could improve the interface somehow, in a way that makes it more obvious which options there are. I'm not sure how, just wanted to throw that thought out here.

crates/breez-sdk/core/src/sdk.rs
danielgranhao: Yes. I also just found out about this feature.

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -868,14 +906,16 @@ pub enum OnchainConfirmationSpeed {
...
     #[cfg_attr(feature = "uniffi", uniffi(default=None))]
-    pub amount: Option<u128>,
-    /// If provided, the payment will be for a token.
-    /// May only be provided if the payment request is a spark address.
+    pub pay_amount: Option<PayAmount>,
+    /// Confirmation speed for on-chain payments.
+    /// Returns an error if provided for non-on-chain payment types.
     #[cfg_attr(feature = "uniffi", uniffi(default=None))]
-    pub token_identifier: Option<String>,
+    pub onchain_speed: Option<OnchainConfirmationSpeed>,
```
dangeross: Is there any why to avoid having to configure this here, I it breaks the notion of using the SendPaymentOptions (disjoint in flows depending on the payment type).

It looks like the main reason to add it here is to set the `amount` in the response. What if the `amount` becomes a `pay_amount`, fees can still be returned in the response and we can use the prepare_response `pay_amount` in the send_payment() to calculate the actual amount.

Could also take the same approach for `PrepareLnurlPayResponse`

roeierez (commented):
Looking good! 
I think worth putting some effort in make the interface more concise.

PR #549, opened 2026-01-14
ConversionInfo refactor

dangeross:
This PR:
- Renames structs/enums/fields from TokenConversion to Conversion
- Returns also the estimated amount converted in a ConversionEstimate struct in the prepared response
- Refactors the ConversionInfo to:
  - Link related conversions with a generated conversion_id, including the manual refund
  - Add a purpose to the received conversion payment of SelfTransfer or OngoingPayment(payment_request)
  - Add a status to track if the conversion needs a manual refund
- Unifies caching of payment metadata for LNURL-withdraw to allow caching of metadata by tx hash for token payments which are applied when the token transaction is synced


dangeross:
@claude short review of this PR

PR #557, opened 2026-01-16
fix: drop direct tx and direct refund tx from tree creation

danielgranhao:
This fixes the use of `start_deposit_tree_creation`, since providing `direct_root_tx` and `direct_refund_tx` has been deprecated. Validation to prevent providing these fields has already been deployed on regtest. 

This only affects non-static deposits, which are not exposed in the breez-sdk API. Because of this, the validation change was not caught by our breez-itest tests. 

The spark-itest tests run agaisnt a local SO, but the default configuration still doesn't apply the validation. So they didn't caught the change as well. This PR also adds a spark-itest non-static deposit test using regtest.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -37,6 +37,9 @@ jobs:
   itest:
     name: Integration test
     runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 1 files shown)


dangeross (approved):
LGTM

PR #565, opened 2026-01-19
fix(parser): validate BIP-353 domain DNS labels correctly.

pulkit-999:
RFC 1035 limits DNS labels to 63 characters, not entire domain names.

This change validates each domain label individually and rejects
empty labels, fixing incorrect rejection of valid BIP-353 addresses.


pulkit-999:
> But we don't support BIP353, so can you please explain this PR?

We don’t add or expand BIP-353 support in this PR.

The parse_bip_353 logic already exists and is currently executed as part of the input parsing flow.
This change only fixes an incorrect DNS validation check inside that existing code path.

Previously, the code rejected valid multi-label domains because it applied the 63-character limit to the entire domain instead of individual DNS labels.
RFC 1035 specifies that the 63-character limit applies per label, not per full domain.

PR #573, opened 2026-01-21
Lock-Free Tree Store

roeierez:
This PR replaces the mutex-based InMemoryTreeStore with a queue-based command processing pattern to address severe performance issues under load.

**Problem**
The previous implementation had significant drawbacks when handling concurrent payments:

* High lock contention: Mutex-based synchronization created bottlenecks under load
* Leaf racing:  Concurrent payments competed for the same leaves, causing excessive failed retries
* Swap inefficiency: When a reservation required a swap, leaves were added to state before re-selection, forcing the current * reserve call to race for the very leaves it just swapped, significantly reducing its chance of success

**Solution**
The new architecture introduces:

* Single-threaded state processor: All state mutations flow through an mpsc channel, eliminating mutex contention
* Balance change notifications: Watch channel enables efficient waiting, retries only occur when success is possible
* Concurrency limiting: Semaphore restricts concurrent reservations to 10 (currently hardcoded; configuration options will be exposed once the architecture stabilizes)
* In-place swap updates: Reservations are updated directly after swap rather than requiring a new reservation call

Testing
Added parallel-perf, a new parallel benchmark tool as a binary in the breez-bench crate for throughput testing.

### DIFF
--- a/crates/breez-sdk/breez-bench/src/bin/parallel_perf.rs
+++ b/crates/breez-sdk/breez-bench/src/bin/parallel_perf.rs
@@ -1,0 +1,608 @@
+//! Parallel Payment Performance Testing Tool
+//!
+//! Tests parallel payment throughput in regtest by executing
+//! multiple Spark transfers and Lightning payments concurrently.
+
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+
+use anyhow::{Result, bail};
+use clap::Parser;
+use rand::seq::SliceRandom;
+use rand::{Rng, SeedableRng};
+use tokio::sync::mpsc;
+use tracing::{info, warn};
+use tracing_subscriber::EnvFilter;
+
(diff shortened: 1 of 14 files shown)


crates/breez-sdk/breez-bench/src/bin/parallel_perf.rs
```diff
@@ -0,0 +1,676 @@
...
+            Err(_) => bail!(
+                "Timeout waiting for Synced event after {} seconds",
+                timeout_secs
+            ),
+        }
+    }
+}
+
+/// Fund wallet via regtest faucet
+async fn fund_via_faucet(sdk_instance: &mut BenchSdkInstance, amount: u64) -> Result<()> {
```
danielgranhao: Same here and in `wait_for_claimed_event`

PR #575, opened 2026-01-22
fix: lower minimum fee threshold for refunds to 194 sats

### LINKED ISSUE #574: Lower minimum required fees for refund from 300 to 194 sats
SDK currently enforces a hardcoded minimum fee of [300 sats](https://github.com/breez/spark-sdk/blob/95923a07989b7e4cce372a07b43b6b6acac7e8e7/crates/spark/src/services/deposit.rs#L314) for refund transactions, but this value has been lowered to [194 sats](https://github.com/buildonspark/spark/blob/a9f1e408bbc77a6f0bf3e94bf5d69d8704513087/sdks/js/packages/spark-sdk/src/spark-wallet/spark-wallet.ts#L2199) based on an approximate minimum fee needed for a 1-in/1-out transaction on the upstream Spark SDK(`@a9f1e408bbc77a6f0bf3e94bf5d69d8704513087`).

This causes issues when using the `recommendedFees` API, which can return fee rates(≤2 sats/vB) that create fees smaller than 300 sats, leading to refunds being aborted even though the fee would be sufficient according to Spark's actual requirements.

erdemyerebasmaz:
Fixes #574

This PR aligns minimum fee threshold of refunds with upstream Spark SDK[@a9f1e40](https://github.com/buildonspark/spark/blob/a9f1e408bbc77a6f0bf3e94bf5d69d8704513087/sdks/js/packages/spark-sdk/src/spark-wallet/spark-wallet.ts#L2199).

### DIFF
--- a/docs/breez-sdk/snippets/csharp/RefundingPayments.cs
+++ b/docs/breez-sdk/snippets/csharp/RefundingPayments.cs
@@ -83,9 +83,6 @@ namespace BreezSdkSnippets
             // or using a fixed amount
             //var fee = new Fee.Fixed(amount: 500);
             //
-            // Important: The total fee must be at least 194 sats to ensure the
-            // transaction can be relayed by the Bitcoin network. If the fee is
-            // lower, the refund request will be rejected.
 
             var request = new RefundDepositRequest(
                 txid: txid,
--- a/docs/breez-sdk/snippets/flutter/lib/refunding_payments.dart
+++ b/docs/breez-sdk/snippets/flutter/lib/refunding_payments.dart
@@ -61,9 +61,6 @@ Future<void> refundDeposit(BreezSdk sdk) async {
   // or using a fixed amount
   //Fee fee = Fee.fixed(amount: BigInt.from(500));
   //
-  // Important: The total fee must be at least 194 sats to ensure the
-  // transaction can be relayed by the Bitcoin network. If the fee is
-  // lower, the refund request will be rejected.
 
   final request = RefundDepositRequest(
     txid: txid,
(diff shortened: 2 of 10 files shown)


crates/spark/src/services/deposit.rs
```diff
@@ -311,9 +311,9 @@ impl DepositService {
         refund_tx.input[0].witness = witness;
 
         let fee_sats = fee.to_sats(refund_tx.vsize() as u64);
-        if fee_sats <= 300 {
+        if fee_sats <= 194 {
             return Err(ServiceError::Generic(
-                "fee must be more than 300 sats".to_string(),
+                "fee must be more than 194 sats".to_string(),
             ));
         }
```
erdemyerebasmaz: **Extract magic number to named constant**

The value 194 appears twice (comparison and error message) and should be extracted to a module-level constant for maintainability and clarity.

**Suggestion:**
```rust
const MIN_REFUND_FEE_SATS: u64 = 194;

// Then use:
if fee_sats <= MIN_REFUND_FEE_SATS {
    return Err(ServiceError::Generic(
        format!("fee must be more than {} sats", MIN_REFUND_FEE_SATS),
    ));
}
```

This makes the rationale clear (minimum fee for a 1-in/1-out transaction) and prevents the two values from drifting apart in future edits. Similar to how `CLAIM_STATIC_DEPOSIT_ACTION` is defined at line 40.

erdemyerebasmaz: 6273a4573900666c2375d562bcddb1d66498a098

PR #577, opened 2026-01-23
Add PostgreSQL storage backend support to Breez SDK

danielgranhao:
Changes:
  - Add PostgreSQL storage as an alternative to SQLite on native builds
  - Configurable connection pool (size, timeouts)
  - Gated behind postgres feature flag, included in bindings where it makes sense

Postgres not supported on:
- Flutter: I didn't think it made sense given is primarily for mobile/desktop client apps. But we can add it without too much effort if we want
- react native: disabled for same reason
- Wasm Node.js: makes sense to implement, but requires having another PostgreSQL storage implementation in JS. I propose we consider doing this later if/when there's demand.

Other notes:
- SSL is supported but skips certificate verification. IIUC this is common practice and is fine as long as the postgres instance is in the same private network as in most deployment setups.

### DIFF
--- a/crates/breez-sdk/core/src/persist/postgres.rs
+++ b/crates/breez-sdk/core/src/persist/postgres.rs
@@ -923,9 +923,9 @@ impl Storage for PostgresStorage {
         let limit = i64::from(request.limit.unwrap_or(u32::MAX));
         let offset = i64::from(request.offset.unwrap_or(0));
 
+        let offset_idx = param_idx + 1;
         let query = format!(
-            "{SELECT_PAYMENT_SQL} {where_sql} ORDER BY p.timestamp {order_direction} LIMIT ${param_idx} OFFSET ${}",
-            param_idx + 1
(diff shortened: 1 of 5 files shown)


roeierez (approved):
LGTM

crates/breez-sdk/core/src/persist/postgres.rs
```diff
@@ -0,0 +1,2038 @@
...
+            "ASC"
+        } else {
+            "DESC"
+        };
+
+        let limit = i64::from(request.limit.unwrap_or(u32::MAX));
+        let offset = i64::from(request.offset.unwrap_or(0));
+
+        let query = format!(
+            "{SELECT_PAYMENT_SQL} {where_sql} ORDER BY p.timestamp {order_direction} LIMIT ${param_idx} OFFSET ${}",
```
JssDWt: nit: Give the offset an idx too?

JssDWt (commented):
It's all looking good. The transition to including the sync storage in the storage trait seems fine. Although perhaps not necessary? Someone running with postgres on a backend would generally not need any real-time sync, so would run with real-time sync disabled? Is that a fair assumption?

PR #592, opened 2026-01-30
fix: ensure unique payment matching in recovery test

pulkit-999:
Add timestamp to payment matching criteria to prevent false
positives when multiple payments share the same method, type,
amount, and status.


danielgranhao:
False positives (at least due to timestamp mismatches) would be caught by the timestamp assertion that follows the code block you changed, right @pulkit-999?

PR #594, opened 2026-01-30
bindings: Remove build dependencies

dangeross:
The uniffi_bindgen, goblin, clap_builder, clap, cargo_metadata, textwrap and askama dependencies are being statically linked into the runtime library because of the uniffi dependency features "bindgen-tests" and "cli" in the bindings crate.

Removing them makes the most effect on the iOS static library, dropping ~20MB from each arch:
- Single archs: ~172MB -> ~152MB
- Universal: ~347MB -> ~303MB


roeierez (approved):
Awsome!

danielgranhao (approved):
Nice!

PR #595, opened 2026-02-01
fix: add username validation for lightning addresses

pulkit-999:
Add validation to ensure usernames are non-empty and between 3-32 
characters after sanitization. Apply consistently across check and 
register methods to catch errors early.

### DIFF
--- a/crates/breez-sdk/lnurl/src/routes.rs
+++ b/crates/breez-sdk/lnurl/src/routes.rs
@@ -105,6 +105,7 @@ where
         Extension(state): Extension<State<DB>>,
     ) -> Result<Json<CheckUsernameAvailableResponse>, (StatusCode, Json<Value>)> {
         let username = sanitize_username(&identifier);
+        validate_username(&username)?;
         let user = state
             .db
             .get_user_by_name(&sanitize_domain(&state, &host)?, &username)


crates/breez-sdk/core/src/sdk/lightning_address.rs
```diff
@@ -20,7 +18,7 @@ impl BreezSdk {
             ));
         };
 
-        let username = sanitize_username(&req.username);
+        let username = validate_and_sanitize_username_for_sdk(&req.username)?;
```
dangeross: I think its clearer like this (same as in the lnurl module) and drop the _for_sdk fn
```suggestion
        let username = validate_and_sanitize_username(&req.username).map_err(SdkError::InvalidInput)?;
```

crates/breez-sdk/core/src/sdk/helpers.rs
```diff
@@ -1,23 +1,23 @@
...
 use x509_parser::parse_x509_certificate;
 
-use crate::{
-    PaymentDetails, WaitForPaymentIdentifier,
-    error::SdkError,
-    events::{EventListener, SdkEvent},
-    models::Payment,
-    persist::{CachedAccountInfo, ObjectCacheRepository, Storage},
-};
-
```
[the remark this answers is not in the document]
dangeross: Still not reverted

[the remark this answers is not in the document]
pulkit-999: actually after doing cargo fmt it pushes it to the top

crates/breez-sdk/lnurl/src/routes.rs
```diff
@@ -104,7 +104,8 @@ where
         Path(identifier): Path<String>,
         Extension(state): Extension<State<DB>>,
     ) -> Result<Json<CheckUsernameAvailableResponse>, (StatusCode, Json<Value>)> {
-        let username = sanitize_username(&identifier);
+        let username = validate_and_sanitize_username(&identifier)
+            .map_err(|e| (StatusCode::BAD_REQUEST, Json(Value::String(e))))?;
```
dangeross:
```suggestion
        validate_username(&username)?;
```

crates/breez-sdk/lnurl/src/routes.rs
```diff
@@ -837,33 +842,6 @@ fn validate_nostr_zap_request(
...
-        trace!("invalid username doesn't match regex");
-        return Err((
-            StatusCode::BAD_REQUEST,
-            Json(Value::String("invalid username".into())),
-        ));
-    }
-
-    Ok(())
-}
-
```
dangeross: Hey @pulkit-999. I'm sorry for the back-and-forth on this. 

So here it looks like there is good validation in the lnurl crate already. The regex also checks for empty strings. You can test the regex [here](https://regex101.com/r/y1mA9u/1), no match would return an error in our case.

The only change I would suggest (see suggestion) is to also validate the username in the `available` fn, where we check the username is available. We don't need any changes in breez-sdk/core because the server validates for us.

PR #597, opened 2026-02-02
Reflect deposit claim fees in payments

JssDWt:
Reflect the actual amount paid in fees in the payment for received deposits. Previously the claim fees were not reflected, i.e. they were always zero. Now the claim fees are deducted from the received amount, as expected.


danielgranhao (approved):
Other than Ross's comment, LGTM

JssDWt:
Updated our models to reflect the graphql schema. Deprecated enum variants are removed in our local representation.

PR #599, opened 2026-02-02
Fee policy support

danielgranhao:
This replaces the Drain feature with a more generic fee policy feature. Instead of selecting drain, consumers can choose whether fees are included or excluded in the amount they selected. Draining can be done by sending the balance amount with `FeesIncluded`. 

Changes:
- Remove `PayAmount` and go back to amount + token identifier in prepare requests and responses
- Add `FeePolicy` enum to requests. By default, `FeesExcluded` is used, maintaining current behavior.
- Support for draining to amountless bolt11 was added

TODO:

- [x] Docs

### DIFF
--- a/crates/breez-sdk/core/src/sdk/mod.rs
+++ b/crates/breez-sdk/core/src/sdk/mod.rs
@@ -54,14 +54,6 @@ pub(crate) struct SyncRequest {
 }
 
 impl SyncRequest {
-    pub(crate) fn new(reply: oneshot::Sender<Result<(), SdkError>>, sync_type: SyncType) -> Self {
-        Self {
-            sync_type,
-            reply: Arc::new(Mutex::new(Some(reply))),
-            force: true,
-        }
-    }
-
     pub(crate) fn full(reply: Option<oneshot::Sender<Result<(), SdkError>>>) -> Self {
         Self {
             sync_type: SyncType::Full,
--- a/crates/breez-sdk/core/src/sdk/sync.rs
+++ b/crates/breez-sdk/core/src/sdk/sync.rs
@@ -222,18 +222,11 @@ impl BreezSdk {
             return;
         }
 
-        // Just sync all lnurl metadata here, no need to be picky.
-        let (tx, rx) = oneshot::channel();
-        if let Err(e) = self
-            .sync_trigger
-            .send(SyncRequest::new(tx, SyncType::LnurlMetadata))
-        {
-            error!("Failed to trigger lnurl metadata sync: {e}");
-            return;
-        }
-
-        if let Err(e) = rx.await {
-            error!("Failed to sync lnurl metadata for invoice {}: {e}", invoice);
+        // Sync lnurl metadata directly instead of going through the sync trigger,
+        // because this function is called from the sync loop's event handler,
+        // which would deadlock waiting for itself to process the trigger.
+        if let Err(e) = self.sync_lnurl_metadata().await {
+            error!("Failed to sync lnurl metadata for invoice {invoice}: {e}");
             return;
         }
 


crates/breez-sdk/core/src/error.rs
```diff
@@ -58,11 +58,8 @@ pub enum SdkError {
     #[error("Lnurl error: {0}")]
     LnurlError(String),
 
-    #[error("Drain not supported for this payment method")]
-    DrainNotSupported,
-
     #[error("Balance {balance_sats} sats exceeds LNURL maximum sendable {max_sendable_sats} sats")]
-    DrainExceedsLnurlMax {
+    SendAllExceedsLnurlMax {
```
dangeross: Does this need renaming?

docs/breez-sdk/snippets/go/send_payment.go
```diff
@@ -267,24 +275,37 @@ func SendPaymentSpark(sdk *breez_sdk_spark.BreezSdk, prepareResponse breez_sdk_s
 	return &payment, nil
 }
 
-func PrepareSendPaymentDrain(sdk *breez_sdk_spark.BreezSdk) (*breez_sdk_spark.PrepareSendPaymentResponse, error) {
-	// ANCHOR: prepare-send-payment-drain
-	// Use PayAmountDrain to send all available funds
+func PrepareSendPaymentFeesIncluded(sdk *breez_sdk_spark.BreezSdk) (*breez_sdk_spark.PrepareSendPaymentResponse, error) {
+	// ANCHOR: prepare-send-payment-fees-included
+	// By default (FeePolicy.FeesExcluded), fees are added on top of the amount.
```
dangeross:
```suggestion
	// By default (FeePolicyFeesExcluded), fees are added on top of the amount.
```

docs/breez-sdk/snippets/go/send_payment.go
```diff
@@ -267,24 +275,37 @@ func SendPaymentSpark(sdk *breez_sdk_spark.BreezSdk, prepareResponse breez_sdk_s
...
+	// Use FeePolicy.FeesIncluded to deduct fees from the amount instead.
```
dangeross:
```suggestion
	// Use FeePolicyFeesIncluded to deduct fees from the amount instead.
```

crates/breez-sdk/core/src/sdk/lnurl.rs
```diff
@@ -24,12 +24,15 @@ impl BreezSdk {
...
-            BitcoinPayAmount::Bitcoin { amount_sats } => *amount_sats,
-            BitcoinPayAmount::Drain => {
-                return self.prepare_lnurl_pay_drain(request).await;
-            }
-        };
+        let fee_policy = request.fee_policy.unwrap_or_default();
+        let amount_sats = request.amount_sats;
+
+        // FeesIncluded uses the double-query approach
+        if fee_policy == FeePolicy::FeesIncluded {
```
roeierez: Should we validate and reject when conversion_options is set here?

danielgranhao:
@roeierez @dangeross Added [7d5bee7](https://github.com/breez/sparksdk/pull/599/commits/7d5bee7e7b700999b5efb51b1c1f5459c848d02c) which fixes a deadlock in `sync_single_lnurl_metadata`. This function is called from `handle_wallet_event`, which runs inside the sync loop's tokio::select! event branch. The old code sent a `SyncRequest` into the broadcast channel and awaited a reply — but the only consumer of that channel is the sync trigger branch of the same select! loop, which can't run while the event branch is blocked.

Test `test_08_lnurl_send_all_with_fee_overpayment` recently started to fail reliably due to this deadlock (also on `main`)

dangeross (approved):
LGTM

PR #601, opened 2026-02-03
spark-wallet: Stop background tasks on SDK disconnect

dangeross:
The BackgroundProcessor in SparkWallet (running token output optimization and event subscription) continued running after disconnect() was called. This occurred because the wallet created its own internal cancellation channel that only triggered on drop, while BreezSdk::disconnect() signaled a separate channel the wallet didn't observe. The Arc references held by background tasks prevented the wallet from being dropped.

The fix adds WalletBuilder::with_cancellation_token() to pass the SDK's shutdown signal to the wallet. When provided, background tasks observe this external token; otherwise, the original drop-based behavior is preserved for standalone usage.

This was observed using Wasm where the backgrounds tasks kept running after `sdk.disconnect(); sdk = null`. This might also happen in the UniFFI bindings until garbage collection occurs.

### DIFF
--- a/crates/breez-sdk/core/src/sdk_builder.rs
+++ b/crates/breez-sdk/core/src/sdk_builder.rs
@@ -368,8 +368,11 @@ impl SdkBuilder {
         spark_wallet_config.leaf_optimization_options.multiplicity =
             self.config.optimization_config.multiplicity;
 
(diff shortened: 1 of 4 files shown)


danielgranhao (approved):
Nice catch

PR #603, opened 2026-02-03
refactor(timelock): extract duplicate signing commitments code into helper method

pulkit-999:
- Created get_signing_commitments_for_jobs() helper method to eliminate code duplication across renew_node(), renew_refund(), and renew_zero_timelock() .
- The three renewal methods previously contained identical code blocks for fetching signing commitments . This DRY violation made maintenance difficult and increased the risk of inconsistent updates.


dangeross (approved):
LGTM

PR #604, opened 2026-02-04
Update query token txs proto

danielgranhao:
Changes:
- Updates to new query token txs proto
- Separates querying using filters from querying specific txs by hash
- Adjusts recovery integration test to cover pagination

### DIFF
--- a/crates/spark/src/services/tokens.rs
+++ b/crates/spark/src/services/tokens.rs
@@ -207,8 +207,6 @@ impl TokenService {
             ],
         };
 
-        // TODO: ask for ordering field to be added to QueryTokenTransactionsRequest
-        //  until then, PagingFilter's order is not being respected
         let order: crate::operator::rpc::spark::Order = paging.order.into();
         let response = self
             .operator_pool


.github/workflows/ci.yaml
```diff
@@ -75,8 +75,8 @@ jobs:
       FAUCET_USERNAME: ${{ secrets.FAUCET_USERNAME }}
       FAUCET_PASSWORD: ${{ secrets.FAUCET_PASSWORD }}
       FAUCET_CONCURRENCY: "2"  # Max concurrent faucet requests (rate limiting)
-      RECOVERY_TEST_MNEMONIC: ${{ secrets.RECOVERY_TEST_MNEMONIC }}
-      RECOVERY_TEST_EXPECTED_PAYMENTS: ${{ secrets.RECOVERY_TEST_EXPECTED_PAYMENTS }}
+      RECOVERY_TEST_MNEMONIC_V2: ${{ secrets.RECOVERY_TEST_MNEMONIC_V2 }}
+      RECOVERY_TEST_EXPECTED_PAYMENTS_V2: ${{ secrets.RECOVERY_TEST_EXPECTED_PAYMENTS_V2 }}
```
dangeross: I'm wondering if we need to change the env vars used in the itest also, or just the secrets vars that are passed to it?

crates/spark/src/services/tokens.rs
dangeross: Can be removed?

PR #610, opened 2026-02-06
fix sync revision number usage

JssDWt:
Fix the usage of the revision number. Because we weren't doing any updates to the same records, this was never noticed. We were supposed to send the revision number of the _previous_ version of the record with the same id. We were sending the new revision number of the record. This wasn't checked by the server, because that check is only enforced if the record already has a parent version.

Helps with #569


roeierez (approved):
LGTM.

dangeross (approved):
LGTM

roeierez (approved):

PR #614, opened 2026-02-10
token-conversions: Fix V3 pool selection

dangeross:
This PR:
- Fixes `calculate_amount_in` fn to use price-based calculation for V3 concentrated pools instead of constant product formula
- Corrects inverted price formulas when using price-based calculation
- Removes the bit-masking (multiples of 64) requirement for V3 pools
- Reduces the default max slippage bps to 10 bps, as USDB V3 pools have virtually no slippage
- Auto-populates integrator config


danielgranhao (approved):
LGTM

PR #615, opened 2026-02-10
docs: Incline fn/param/enum name preprocessor

dangeross:
Adds a preprocessor to auto-format fn/param/enums to the selected language. E.g. for `get_info` displays `GetInfo` when Go language is selected


.claude/commands/update-snippets.md
danielgranhao: Does this belong here? This command is for updating snippets. The added info is relevant for updating the docs themselves

PR #620, opened 2026-02-11
HODL invoice support

### LINKED ISSUE #520: Support HODL invoices

danielgranhao:
Closes #520

  - Add HODL invoice support: users can create Lightning invoices with an externally-provided payment hash, where the SSP holds the HTLC until the preimage is provided via `claim_htlc_payment` or the HTLC expires
  - Add `payment_hash` field to `Bolt11Invoice` receive method to opt into HODL mode                                                                                                                                                                                                                      
  - Add `htlc_details` (with `SparkHtlcStatus`) to `PaymentDetails::Lightning` for tracking HTLC state on Lightning payments
  - Add `PaymentDetailsFilter::Lightning` with `htlc_status` filtering to query HODL invoice status                                                                                                                                                                                                       
  - Expose HODL invoice creation in CLI (`--hodl` flag) and internal CLI (`lightning create-hodl-invoice`)
  - Update all storage backends (SQLite, Postgres, Node.js, IndexedDB) with new `htlc_details` column and Lightning HTLC status filtering
  - Update WASM and Flutter bindings with new fields

TODO: 

- [x] Add itest
- [x] Docs

### DIFF
--- a/crates/breez-sdk/core/src/models/adaptors.rs
+++ b/crates/breez-sdk/core/src/models/adaptors.rs
@@ -61,6 +61,17 @@ fn derive_htlc_details_from_ssp(
     })
 }
 
+/// If the HTLC details are missing a preimage, fill it in from the given fallback and update
+/// the status to [`SparkHtlcStatus::PreimageShared`] accordingly.
+fn reconcile_htlc_preimage(details: &mut SparkHtlcDetails, preimage: Option<&str>) {
+    if details.preimage.is_none() {
+        details.preimage = preimage.map(ToString::to_string);
+    }
+    if details.preimage.is_some() {
+        details.status = SparkHtlcStatus::PreimageShared;
+    }
+}
+
 impl PaymentMethod {
     fn from_transfer(transfer: &WalletTransfer) -> Self {
         match transfer.transfer_type {
@@ -80,6 +91,7 @@ impl PaymentMethod {
 }
 
 impl PaymentDetails {
+    #[allow(clippy::too_many_lines)]
     fn from_transfer(transfer: &WalletTransfer) -> Result<Option<Self>, SdkError> {
         if !transfer.is_ssp_transfer {
             // Check for Spark invoice payments
@@ -124,7 +136,12 @@ impl PaymentDetails {
                         "Invalid invoice in SspUserRequest::LightningReceiveRequest".to_string(),
                     ))?;
                 let htlc_details = if let Some(req) = &transfer.htlc_preimage_request {
-                    req.clone().try_into()?
+                    let mut details: SparkHtlcDetails = req.clone().try_into()?;
+                    reconcile_htlc_preimage(
+                        &mut details,
+                        request.lightning_receive_payment_preimage.as_deref(),
+                    );
+                    details
                 } else {
                     derive_htlc_details_from_ssp(
                         transfer,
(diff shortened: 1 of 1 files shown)


crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -375,6 +375,9 @@ pub enum PaymentDetails {
         /// The invoice destination/payee pubkey
         destination_pubkey: String,
 
+        /// The HTLC transfer details. May be absent for legacy payments.
```
dangeross: It wasn't possible before was it? Or do non-HODL lightning payments also have HTLC details?
```suggestion
        /// The HTLC transfer details.
```

crates/spark/src/ssp/graphql/models.rs
```diff
@@ -203,7 +205,7 @@ pub enum LightningReceiveRequestStatus {
     TransferCompleted,
     RefundSigningCommitmentsQueryingFailed,
     RefundSigningFailed,
-    #[serde(other, skip_serializing)]
+    #[serde(other)]
```
dangeross: Is there a reason to remove it here and not anywhere else?

crates/breez-sdk/core/src/models/mod.rs
danielgranhao: They have it now. I think I might have misunderstood it before, and that all lightning payments have always included HTLC details. Lightning payments, in both directions, rely on Spark HTLCs to settle with the SSP, so it's normal to have HTLC details.

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -1146,6 +1153,10 @@ pub enum PaymentDetailsFilter {
         /// Filter by transaction type
         tx_type: Option<TokenTransactionType>,
     },
+    Lightning {
+        /// Filter specific Spark HTLC statuses (Spark HTLCs are used to fulfil Lightning invoices)
```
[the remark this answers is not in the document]

crates/breez-sdk/core/src/persist/postgres.rs
```diff
@@ -909,6 +910,29 @@ impl Storage for PostgresStorage {
...
+                        })
+                        .collect();
+                    payment_details_clauses.push(format!(
+                        "l.htlc_details::jsonb->>'status' IN ({})",
+                        placeholders.join(", ")
+                    ));
+                    for htlc_status in htlc_statuses {
+                        params.push(Box::new(htlc_status.to_string()));
+                    }
+                }
```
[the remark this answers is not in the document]
danielgranhao: Good point. Done

crates/breez-sdk/core/src/models/mod.rs
danielgranhao: Btw @dangeross, I checked, and for old payments, there are cases we don't find the HTLC details. So the rustdoc stands. I don't love that we have to keep this field optional for this reason, but not sure what would be a better alternative.

danielgranhao: Done. Also removed preimage and hash from top level Lightning variant and now are only on HTLC details.

crates/breez-sdk/core/src/persist/postgres.rs
```diff
@@ -700,6 +700,8 @@ impl PostgresStorage {
                 "UPDATE sync_revision SET revision = 0",
                 "DELETE FROM settings WHERE key = 'sync_initial_complete'",
             ],
+            // Migration 6: Add htlc_details to lightning payments
+            &["ALTER TABLE payment_details_lightning ADD COLUMN htlc_details JSONB"],
```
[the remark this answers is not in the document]
danielgranhao: In the process of wanting to remove the no longer necessary preimage and hash columns I decided to just remove the htlc details column and put the new values on the details table. I think it resolves your concern as well.

PR #621, opened 2026-02-11
Distributed lock

dangeross:
Add distributed lock and token transfer retry for stable balance:
- Integrate distributed lock into stable balance: 
  - For payments SyncLockGuard acquires a non-exclusive "payment_conversion" lock on creation (fire-and-forget) and releases on drop when no payments remain, preventing auto-conversion races across SDK instances
  - For auto-conversion aquires an exclusive "auto_conversion" lock, which is held for a single instance. On failure to aquire, skips auto-convert
- Simplify reservation tracking from amount-based ReservationTracker to count-based LockCounter using AtomicU64
- Add retry logic in TokenService::transfer_tokens() for TRANSACTION_PREEMPTED errors — refreshes token outputs from operator and retries up to 3 times
- Add distributed lock integration tests using DataSyncFixture with TestSyncSigner

Data sync PR: https://github.com/breez/data-sync/pull/25

### DIFF
--- a/crates/flashnet/src/pool_selection.rs
+++ b/crates/flashnet/src/pool_selection.rs
@@ -54,7 +54,7 @@ pub fn select_best_pool(
         ));
     }
 
-    // Filter out non V3 oncentrated pools if any V3 concentrated pool exists
+    // Filter out non V3 concentrated pools if any V3 concentrated pool exists
     let has_v3_concentrated = viable_pools
         .iter()
         .any(|(pool, _)| pool.curve_type == Some(CurveType::V3Concentrated));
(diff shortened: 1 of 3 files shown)


crates/flashnet/src/pool_selection.rs
```diff
@@ -54,6 +54,20 @@ pub fn select_best_pool(
         ));
     }
 
+    // Filter out non V3 oncentrated pools if any V3 concentrated pool exists
```
danielgranhao:
```suggestion
    // Filter out non V3 concentrated pools if any V3 concentrated pool exists
```

PR #627, opened 2026-02-12
feat: add pubkey-based bech32 variant to LnurlInfo

erdemyerebasmaz:
## Summary
- Adds `bech32_pubkey` field to `LnurlInfo` — bech32-encoded LNURL using the pubkey-based path (`/lnurlp/{pubkey}`)
- Computed client-side from domain + identity pubkey in both register and recover flows

## Test plan
- [x] Verify `bech32_pubkey` is populated after register and recover
- [x] Verify `bech32_pubkey` decodes to a valid callback URL


kingonly:
Good point re the identity. Let's keep it with a user name then.

PR #631, opened 2026-02-13
spark: Update to hash variant v2

dangeross:
Updates to HashVariant V2 by implementing BIP-340 tagged hashing

- Add TaggedHasher utility which implements the BIP-340 tagged hashing scheme
- Update deposit proof of possession verification hash
- Update transfer signing payload hash
- Update static deposits to use RPC `generate_static_deposit_address`
- Upgrade lightning to RPC `store_preimage_share_v2` with ECIES encryption ~and signed payload~


dangeross:
The user signature is now removed from `store_preimage_share_v2`

PR #636, opened 2026-02-18
Skip serializing empty payment metadata fields

danielgranhao:
This fixes an issue that could cause `PaymentMetadata` fields to be overwritten in rtsync records. `merge()` overlays fields on top of the parent, so if we serialize None, we lose any existing value.

@dangeross spotted this issue experimentally and verified that this change fixes it.


roeierez (approved):
LGTM

dangeross (approved):
LGTM

PR #638, opened 2026-02-18
Concurrent claims

roeierez:
This PR introduces two main features to improve SDK performance in high-throughput environments:                                                                          
                                                                                                                                                                            
  - Concurrent Transfer Claims: Adds configurable parallel claiming of pending Spark transfers via the new max_concurrent_claims configuration option                       
  - SyncCoordinator: New abstraction that coalesces duplicate sync requests, reducing redundant sync operations under load                                                  
                                                                                                                                                                            
  **Changes**                                                                                                                                                                   
                                                                                                                                                                            
  Concurrent Claims                                                                                                                                                         
                                                                                                                                                                            
  - Added max_concurrent_claims configuration option to Config (default: 1)                                                                                                 
  - Claims now process in parallel using buffer_unordered when configured with values > 1                                                                                   
  - Failed claims are logged as warnings but don't block other claims (best-effort error handling)                                                                          
  - Useful for server environments with high incoming payment volume                                                                                                        
                                                                                                                                                                            
  **SyncCoordinator**                                                                                                                                                           
                                                                                                                                                                            
  - New SyncCoordinator struct replaces direct sync_trigger usage                                                                                                           
  - Coalesces multiple sync requests of the same type: if requests arrive while a sync is running, they share a single NEW sync that starts after the current one completes 
  - Different sync types are processed in order                                                                                                                             
  - Provides both trigger_sync_and_wait() and trigger_sync_no_wait() methods
  
  **Behavioral** Changes                                                                                                                                                        
                                                                                                                                                                            
  Claims now use best-effort error handling: Previously, if any single claim failed, the entire claim_pending_transfers operation would fail. Now, failed claims are logged 
  as warnings and successful claims continue. This is more resilient but callers won't see individual claim errors as exceptions.                                           
                                                                                                                                                                            
  **Configuration**                                                                                                                                                             
                                                                                                                                                                            
 ```
 let mut config = default_config(Network::Mainnet);                                                                                                                        
  // For server environments with high payment volume:                                                                                                                      
  config.max_concurrent_claims = 10;
```

Here are results for running with various configurations:

```
CONCURRENT CLAIMS BENCHMARK RESULTS
Pending transfers: 100

| Concurrency | Total Time | Avg/Claim | Throughput  |
|-------------|------------|-----------|-------------|
|           1 |    132.79s |     1.33s |       0.8/s |
|           2 |     70.31s |     703ms |       1.4/s |
|           4 |     45.12s |     451ms |       2.2/s |
|           8 |     27.19s |     271ms |       3.7/s |
|          16 |     21.90s |     218ms |       4.6/s |
```

### DIFF
--- a/crates/breez-sdk/core/src/models/mod.rs
+++ b/crates/breez-sdk/core/src/models/mod.rs
@@ -649,6 +649,18 @@ pub struct StableBalanceConfig {
 }
 
 impl Config {
+    /// Validates the configuration.
+    ///
+    /// Returns an error if any configuration values are invalid.
+    pub fn validate(&self) -> Result<(), SdkError> {
+        if self.max_concurrent_claims == 0 {
+            return Err(SdkError::InvalidInput(
+                "max_concurrent_claims must be greater than 0".to_string(),
+            ));
+        }
+        Ok(())
+    }
+
     pub(crate) fn get_all_external_input_parsers(&self) -> Vec<ExternalInputParser> {
         let mut external_input_parsers = Vec::new();
         if self.use_default_external_input_parsers {
--- a/crates/breez-sdk/core/src/sdk_builder.rs
+++ b/crates/breez-sdk/core/src/sdk_builder.rs
@@ -226,6 +226,9 @@ impl SdkBuilder {
     /// Builds the `BreezSdk` instance with the configured components.
     #[allow(clippy::too_many_lines)]
     pub async fn build(self) -> Result<BreezSdk, SdkError> {
+        // Validate configuration
+        self.config.validate()?;
+
         // Create the base signer based on the signer source
         let (signer, account_number) = match self.signer_source {
             SignerSource::Seed {

### CODE REVIEW

File: crates/breez-sdk/core/src/models/mod.rs [hunk 1]
```diff
@@ -580,6 +580,10 @@ pub struct Config {
     /// When set, received sats will be automatically converted to the specified token
     /// once the balance exceeds the threshold.
     pub stable_balance_config: Option<StableBalanceConfig>,
+    /// Maximum number of concurrent transfer claims.
+    ///
+    /// Default is 1. Increase for server environments with high incoming payment volume.
+    pub max_concurrent_claims: u32,
```
dangeross: Is there any harm doing more than 1 for end-user wallets? I guess it's just resources, but if there is a low frequency of claims, but a max currency over 1 I assume it has no real effect

roeierez: @dangeross  god point, we can change the default to 4 which is a good default also for server environment. WDYT?

dangeross: Sounds good.  We can always adjust it later

danielgranhao (commented):
One minor thing, otherwise LGTM

PR #641, opened 2026-02-19
Add get_emit_payment helper to consolidate payment event emission

dangeross:
Payment events were emitted without metadata or conversion details because each call site fetched from storage differently (or used the payment direct from a transfer). This particularly caused issues testing token conversions/stable balance where at the application level it usually reacts to events, but it was missing essential information. This PR:
- Moves `get_payment_with_conversion_details` to utils
- Adds `EventEmitter::get_emit_payment()` that re-fetches before emitting
- Replaces 6 duplicated fetch-then-emit call sites

### DIFF
--- a/crates/breez-sdk/core/src/events.rs
+++ b/crates/breez-sdk/core/src/events.rs
@@ -1,18 +1,14 @@
 use core::fmt;
 use std::{
     collections::BTreeMap,
-    sync::{
-        Arc, OnceLock,
-        atomic::{AtomicBool, AtomicU64, Ordering},
-    },
+    sync::atomic::{AtomicBool, AtomicU64, Ordering},
 };
 
 use serde::Serialize;
 use tokio::sync::{Mutex, RwLock};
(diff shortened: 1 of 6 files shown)


crates/breez-sdk/core/src/events.rs
```diff
@@ -209,6 +221,27 @@ impl EventEmitter {
         }
     }
 
+    /// Gets the payment from storage to include already stored metadata and conversion details.
+    /// Emit the appropriate event based on its status. Falls back to the provided
+    /// payment if the storage lookup fails.
+    pub(crate) async fn get_and_emit_payment(&self, payment: Payment) {
```
roeierez: Not sure EventEmitter needs to deal with fetching payments. Perhaps this function should stay out of here, fetch the payment and call emit. Still can be reused the same way as in this PR just without fetching logic inside EventEmitter.

roeierez (changes requested):

PR #642, opened 2026-02-19
Update config at runtime

dangeross:
Adds `update_config` API for partial runtime configuration updates without disconnecting
- Supports updating: sync interval, max deposit claim fee, prefer Spark over Lightning, and stable balance config
- Uses Set/Unset enums for fields where clearing has distinct semantics (stable balance, max deposit fee)
- Adds `ServiceShutdown` helper for stop-and-replace lifecycle of service-backed config (stable balance)

The primary motivation of this PR is to enable/disable the use of Stable Balance at runtime, but it also makes sense to enable updating a subset of other config fields at the same time.


crates/breez-sdk/core/src/sdk/mod.rs
```diff
@@ -91,12 +93,44 @@ impl SyncRequest {
...
+    /// Creates a new shutdown linked to the global shutdown.
+    /// Returns `(self, receiver)` — pass the receiver to the service being started.
+    pub(crate) fn new(global_shutdown: &watch::Sender<()>) -> (Self, watch::Receiver<()>) {
+        let (tx, rx) = watch::channel(());
+        let mut global_rx = global_shutdown.subscribe();
+        let tx_clone = tx.clone();
+        tokio::spawn(async move {
+            let _ = global_rx.changed().await;
+            let _ = tx_clone.send(());
+        });
```
danielgranhao: IIUC, each stable balance enable/disable cycle leaks this task. Very minor issue, but could be solved by storing the handle and aborting it on `send()` method

crates/breez-sdk/core/src/sdk/init.rs
```diff
@@ -53,18 +56,31 @@ impl BreezSdk {
...
+
+        let config_service = Arc::new(ConfigService::new(&params.config));
 
         let sdk = Self {
-            config: params.config,
+            static_config: StaticConfig {
+                network: params.config.network,
+                lnurl_domain: params.config.lnurl_domain.clone(),
+                private_enabled_default: params.config.private_enabled_default,
+            },
```
roeierez: Perhaps we can leave without this separation to static config? What prevents us from always using the config service? We control what is immutable or not anyway.

PR #644, opened 2026-02-20
get_lightning_address: wait for recovery

JssDWt:
- Sync the lightning address cache on startup so getLightningAddress returns the up-to-date address without requiring the caller to register first
  - If the background recovery hasn't run yet, getLightningAddress falls back to an inline network call
  - The cache now distinguishes "never recovered" (key absent) from "recovered, no address" (key present, value null), avoiding redundant network calls after a successful recovery that found no address


kingonly:
I don't understand the boolean tbh (same goes for sync but that a different story). quick lookup is for some kind of cache, right? So why doesn't the function goes to the cache if available, and try to fetch if it's not? We're exposing devs to stuff that they don't know of need to handle. 

Unless it works like that by default somehow.

PR #646, opened 2026-02-21
build: upgrade rusqlite to v0.32

darioAnongba:
- Bumps `rusqlite` from 0.31.0 to 0.32.1
- Bumps `rusqlite_migration` from 1.2.0 to 1.3.1 (required for rusqlite 0.32 compatibility)

### Motivation

`rusqlite 0.31.0` pulls in `libsqlite3-sys 0.28.0`. Since `libsqlite3-sys` uses `links = "sqlite3"`, Cargo enforces exactly one version of this crate in the final binary. This makes it impossible to compile any project that combines the Spark SDK with crates depending on a newer `libsqlite3-sys`, notably `sqlx-sqlite 0.8.x `(used by `sea-orm`) which requires `libsqlite3-sys ^0.30.1`.

Upgrading to `rusqlite 0.32.1` aligns on `libsqlite3-sys ^0.30.1`, resolving the conflict and allowing the Spark SDK to coexist with the broader Rust database ecosystem.

`sqlx-sqlite` next release `0.9.x` will allow cohabiting with any version between `v0.30, <v0.37` (See here: https://github.com/launchbadge/sqlx/pull/4161) so we will be able to upgrade `rusqlite` further `v0.38` once released.
Further proof that current supported version`libsqlite3-sys 0.28.0` is way too old and not supported by the Rust ecosystem anymore.

No API changes, the rusqlite surface used by the SDK is stable across `0.31→0.32`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2735,9 +2735,9 @@ dependencies = [
 
 [[package]]
 name = "libsqlite3-sys"
-version = "0.28.0"
+version = "0.30.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0c10584274047cb335c23d3e61bcef8e323adae7c5c8c760540f73610177fc3f"
+checksum = "2e99fb7a497b1e3339bc746195567ed8d3e24945ecd636e3619d20b9de9e9149"
 dependencies = [
  "cc",
  "pkg-config",
@@ -3978,9 +3978,9 @@ dependencies = [
 
 [[package]]
 name = "rusqlite"
-version = "0.31.0"
+version = "0.32.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b838eba278d213a8beaf485bd313fd580ca4505a00d5871caeb1457c55322cae"
+checksum = "7753b721174eb8ff87a9a0e799e2d7bc3749323e773db92e0984debb00019d6e"
 dependencies = [
  "bitflags 2.10.0",
  "fallible-iterator 0.3.0",
@@ -3992,9 +3992,9 @@ dependencies = [
 
 [[package]]
 name = "rusqlite_migration"
-version = "1.2.0"
+version = "1.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "55709bc01054c69e2f1cefdc886642b5e6376a8db3c86f761be0c423eebf178b"
+checksum = "923b42e802f7dc20a0a6b5e097ba7c83fe4289da07e49156fecf6af08aa9cd1c"
 dependencies = [
  "log",
  "rusqlite",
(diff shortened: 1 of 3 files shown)


roeierez (approved):
LGTM! Thanks. Let's wait for the CI to complete.

JssDWt (approved):
LGTM

benthecarman:
We had a gentlemen's agreement between all the rust bitcoin crates using sqlite (ldk-node, cdk, fedimint, etc) to use the same version so our libraries work with eachother, I had previously had made spark-sdk compatible in the past https://github.com/breez/spark-sdk/pull/260/changes/35a96ab40b9fcd661d71eefce9b301610b1149d9 and specifically added a comment to the cargo.toml on why that version was chosen. Annoying that this was ignored, will now need a fork of this to use

kingonly:
@benthecarman there's no intention to annoy you or anyone else. How exactly are we supposed to know and/or manage that? Also note this change wasn't driven by us and it's already merged. Next time keep in the loop.

benthecarman:
> How exactly are we supposed to know and/or manage that? 

I specifically left a comment next to the version to keep in sync with the rest ecosystem.

> Also note this change wasn't driven by us and it's already merged. Next time keep in the loop.

I understand.

JssDWt:
@benthecarman I do want to apologize, because I was aware of the fact rusqlite conflicts were causing errors between bitcoin libraries. I checked the other libraries through our dependency tree and was under the impression we were behind. I must have checked wrong.

I must say the sqlx incompatibility is a painful one too though.

darioAnongba:
> Annoying that this was ignored, will now need a fork of this to use

Wasn't this bound to happen at some point? You want to continue upgrading `spark-sdk`, but don't want to integrate all changes, in which case you are indeed forced to fork. Since SQLite is a peer dependency, anyone building an app with an up-to-date version of SQLite can't integrate `spark-sdk` and is forced to downgrade their version and forfeit all improvements since 2024. Cargo allows only 1 version in the dependency tree.

`sqlx-sqlite v0.9` fixes this by allowing the range `v0.30, <v0.37` (there is a reason why that's the chosen range btw), but your app will still not be compatible because you are on `0.28.0`: https://github.com/launchbadge/sqlx/blob/6956cef0bf3cea81bc0a543a53d550b4dd8d56db/sqlx-sqlite/Cargo.toml#L59

Don't want to bother anyone either but the strongest argument here is for you to upgrade. The best thing for maximum compatibility of `spark-sdk` is to upgrade to v0.9 when released.

benthecarman:
@darioAnongba i agree completely, but the ecosystem should do it together so we don't run into these incompatibilities

darioAnongba:
Agreed, but this has been merged to `main` right? Are you depending on `main` directly? This change should land into a tag/release at some point with a clear changelog message that it's a breaking change. It could also not be cherry-picked into a release at all for some time.

Under trunk based development, the latest version of the code is `main` and contains all the breaking changes.

PR #651, opened 2026-02-23
Recover already existing transfer

JssDWt:
If you create a transfer with an already existing transfer id, the SO might return an AlreadyExists status code. That code is now caught, so recreating that transfer is idempotent.

Verified the added test succeeds after the change, and fails before the change.

Fixes user-facing error:
```
error sending Spark payment: error sending Spark payment (SDK error): SdkError: SparkError: Field0=Service error: service connection error: Connection error: status: AlreadyExists, message: "failed to create transfer for transfer xxx: transfer already exists: ent: constraint failed: ERROR: duplicate key value violates unique constraint \"transfers_pkey\" (SQLSTATE 23505)", details: [x, x, x], metadata: MetadataMap { headers: {"server": "awselb/2.0", "date": "xxx", "content-type": "application/grpc", "content-length": "0"} }
```

### DIFF
--- a/crates/spark-itest/tests/local_wallet_tests.rs
+++ b/crates/spark-itest/tests/local_wallet_tests.rs
@@ -4,6 +4,7 @@ use rstest::*;
 use spark_itest::helpers::{
     WalletsFixture, deposit_to_wallet, deposit_with_amount, wait_for, wallets,
 };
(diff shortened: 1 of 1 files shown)


dangeross (approved):
LGTM

PR #658, opened 2026-02-24
Rename config to support_lnurl_verify

roeierez:
The renaming goal is to make it more clear and feature based.
Also default is false and users need to opt-in rather than opt-out. Since it sends data out of the wallet and not all apps will use it, I think it makes more sense.


dangeross (commented):
The only downside is it disables zaps by default

dangeross (approved):

danielgranhao (approved):

PR #666, opened 2026-02-25
Fix Python event listener deadlock by auto-setting asyncio event loop

### LINKED ISSUE #662: Python SDK: sync_wallet and get_info(ensure_synced=True) hang — background tasks die when uniffi drops build()'s tokio runtime
## Context

Follow-up to #652 / PR #653. Discovered while building [glow-cloud](https://github.com/kingonly/glow-cloud), a self-deployable Bitcoin/Lightning payment API wrapping the Breez Spark SDK (Python/FastAPI, deployed to Vercel serverless).

With `breez-sdk-spark` 0.10.0 and the new `builder.with_postgres_storage(config)` API, the SDK now initializes successfully from Python. However, `sync_wallet()` and `get_info(ensure_synced=True)` hang indefinitely.

## Summary

`init_and_start()` spawns background tasks (including `periodic_sync`) via `tokio::spawn` onto the ephemeral tokio runtime created by uniffi for the `build()` call. When `build()` returns, uniffi drops the runtime, aborting all spawned tasks. Subsequent calls to `sync_wallet()` try to communicate with the dead `periodic_sync` loop via broadcast/oneshot channels and hang forever.

## Environment

- `breez-sdk-spark` 0.10.0 (Python, PyPI)
- Python 3.12.3
- FastAPI + Uvicorn (ASGI)
- Reproduced on Vercel Fluid Compute (serverless)

## Steps to Reproduce

```python
import asyncio
from breez_sdk_spark import (
    SdkBuilder, Seed, default_config, Network,
    default_postgres_storage_config, GetInfoRequest,
    SyncWalletRequest,
)

async def main():
    config = default_config(network=Network.MAINNET)
    config.api_key = "YOUR_API_KEY"

    pg_config = default_postgres_storage_config(
        connection_string="postgres://user:pass@host:5432/db"
    )

    builder = SdkBuilder(config=config, seed=Seed.MNEMONIC(mnemonic="...", passphrase=None))
    await builder.with_postgres_storage(config=pg_config)
    sdk = await builder.build()  # ✅ Works

    # These work:
    info = await sdk.get_info(request=GetInfoRequest(ensure_synced=False))  # ✅
    payments = await sdk.list_payments(request=ListPaymentsRequest())       # ✅
    invoice = await sdk.receive_payment(request=...)                        # ✅

    # These hang forever:
    await sdk.get_info(request=GetInfoRequest(ensure_synced=True))   # ❌ Hangs
    await sdk.sync_wallet(request=SyncWalletRequest())               # ❌ Hangs

asyncio.run(main())
```

### 1. Background tasks spawned on an ephemeral runtime

`SdkBuilder::build()` is exported via uniffi with `async_runtime = "tokio"` ([`bindings.rs`](https://github.com/breez/spark-sdk/blob/main/crates/breez-sdk/core/src/bindings.rs)). UniFFI creates a temporary tokio runtime to poll this future.

At the end of `build()`, it calls `BreezSdk::init_and_start()` ([`init.rs`](https://github.com/breez/spark-sdk/blob/main/crates/breez-sdk/core/src/sdk/init.rs)), a **synchronous** function that calls `sdk.start()`:

```rust
pub(super) fn start(&self, initial_synced_sender: watch::Sender<bool>) {
    self.spawn_spark_private_mode_initialization();
    self.periodic_sync(initial_synced_sender);       // tokio::spawn
    self.try_recover_lightning_address();             // tokio::spawn
    self.spawn_lnurl_preimage_publisher();            // tokio::spawn
}
```

These `tokio::spawn` calls succeed because the uniffi tokio runtime is still active during `build()`. But when `build()` returns the `BreezSdk` to Python, **uniffi drops the runtime**, aborting all spawned tasks — including `periodic_sync`.

### 2. `sync_wallet` waits on a dead channel

`sync_wallet()` calls `sync_coordinator.trigger_sync_and_wait()` ([`sync_coordinator.rs`](https://github.com/breez/spark-sdk/blob/main/crates/breez-sdk/core/src/sdk/sync_coordinator.rs)):

```rust
pub async fn trigger_sync_and_wait(&self, ...) -> Result<(), SdkError> {
    let (tx, rx) = oneshot::channel();
    // ...
    self.run_sync_loop().await;
    rx.await  // ← HANGS: nobody will ever send on this channel
}
```

`run_sync_loop` sends a `SyncRequest` via `broadcast::Sender` and awaits a `reply_rx` oneshot. The receiver is the `periodic_sync` background loop — which is dead. Nobody processes the request, nobody calls `reply()`, so the `rx.await` hangs forever.

### 3. `get_info(ensure_synced=True)` waits on a watch that never updates

In [`api.rs`](https://github.com/breez/spark-sdk/blob/main/crates/breez-sdk/core/src/sdk/api.rs):

```rust
pub async fn get_info(&self, request: GetInfoRequest) -> Result<GetInfoResponse, SdkError> {
    if request.ensure_synced.unwrap_or_default() {
        self.initial_synced_watcher
            .clone()
            .changed()
            .await  // ← HANGS: periodic_sync never sends true
    }
    // ...
}
```

`initial_synced_watcher` is a `watch::Receiver<bool>` initialized to `false`. It is only set to `true` inside the `periodic_sync` loop after a full sync completes. Since that loop is dead, `changed().await` hangs forever.

### The call chain visualized

```
Python: await builder.build()
  → uniffi: creates ephemeral tokio Runtime, polls build() future
    → build() → init_and_start() → start()
      → tokio::spawn(periodic_sync)  ← spawned on ephemeral runtime
      → tokio::spawn(other_background_tasks)
    → build() returns BreezSdk
  → uniffi: drops tokio Runtime  ← ALL SPAWNED TASKS ABORTED
  → returns SDK to Python

Python: await sdk.sync_wallet(...)
  → uniffi: creates NEW ephemeral tokio Runtime
    → sync_wallet() → trigger_sync_and_wait()
      → broadcast SyncRequest to periodic_sync  ← DEAD, nobody listening
      → rx.await  ← HANGS FOREVER
```

## How This Differs from #652

| Aspect | Issue #652 | This Issue |
|--------|-----------|------------|
| **Symptom** | Panic / SIGABRT | Indefinite hang |
| **Cause** | Code running with no tokio runtime at all | Background tasks spawned on ephemeral runtime that gets dropped |
| **Where** | `create_postgres_storage()` — before `build()` | `init_and_start()` — during `build()` |
| **Fix** | Defer storage creation to inside `build()` | Ensure background tasks survive beyond `build()` |

#652 was about **no runtime**. This is about a **temporary runtime** that doesn't outlive `build()`.

## Impact

All Python SDK features that depend on background task coordination are broken:
- `sync_wallet()` — hangs
- `get_info(ensure_synced=True)` — hangs
- Wallet event processing (deposits, claims, stream events) — silently dead
- Periodic sync — never runs

Direct SDK calls (`get_info(ensure_synced=False)`, `receive_payment`, `list_payments`, `send_payment`) work fine because each gets its own ephemeral runtime per call and doesn't depend on background tasks.

## Suggested Fix

**Create a persistent tokio runtime owned by `BreezSdk`:**

Add a `tokio::runtime::Runtime` field to `BreezSdk`. Create it during `init_and_start` and use its handle to spawn background tasks:

```rust
// In BreezSdk struct:
runtime: Arc<tokio::runtime::Runtime>,

// In init_and_start:
let runtime = Arc::new(
    tokio::runtime::Builder::new_multi_thread()
        .enable_all()
        .build()
        .expect("Failed to create tokio runtime")
);

// In start() — use the owned runtime instead of tokio::spawn:
self.runtime.spawn(async move { ... });
```

This runtime lives as long as the `BreezSdk` instance. The `disconnect()` method should shut it down.

## Workaround

- Use `get_info(ensure_synced=False)` instead of `True` — works, reads from PostgreSQL storage
- Do not call `sync_wallet()` — it will hang
- Accept that wallet events (deposits, stream events) won't be processed in real-time — the SDK still works for explicit operations, but background sync is dead

danielgranhao:
## Summary                                                                                                                         
          
  - Auto-call `uniffi_set_event_loop(asyncio.get_running_loop())` before returning a `BreezSdk` instance from `connect()`, `connect_with_signer()`, and `SdkBuilder.build()`                                                                                  
  - Fixes #662: Python `EventListener.on_event` callbacks deadlock when called from Rust's TOKIO1 thread because it has no Python asyncio event loop

## Context

UniFFI 0.28.3 exports `EventListener.on_event` as an async callback interface. When the SDK emits events from the TOKIO1 background thread (e.g., `TransferClaimed` during sync), UniFFI calls `asyncio.run_coroutine_threadsafe()` which needs a registered event loop. Without one, `asyncio.get_running_loop()` raises `RuntimeError: no running event loop`, the callback never completes, and the Rust `.await` hangs forever — deadlocking `periodic_sync` and all subsequent `sync_wallet()`/`get_info(ensure_synced=True)` calls.

UniFFI provides [uniffi_set_event_loop](https://mozilla.github.io/uniffi-rs/latest/futures.html#python-uniffi_set_event_loop) for this, but users had to call it manually. Since all SDK entry points are `async`, a running event loop is always available — so the SDK can set it automatically.

Verified with glow-cloud.

### DIFF
--- a/crates/breez-sdk/bindings/langs/python/src/breez_sdk_spark/__init__.py
+++ b/crates/breez-sdk/bindings/langs/python/src/breez_sdk_spark/__init__.py
@@ -2,6 +2,7 @@ from breez_sdk_spark.breez_sdk_spark_bindings import *
 from breez_sdk_spark.breez_sdk_spark import *
 
 import asyncio
+import functools
 from breez_sdk_spark.breez_sdk_spark import (
     connect as _original_connect,
     connect_with_signer as _original_connect_with_signer,
@@ -14,11 +15,13 @@ def _ensure_event_loop():
     uniffi_set_event_loop(asyncio.get_running_loop())
 
 
+@functools.wraps(_original_connect)
 async def connect(*args, **kwargs):
     _ensure_event_loop()
     return await _original_connect(*args, **kwargs)
 
 
+@functools.wraps(_original_connect_with_signer)
 async def connect_with_signer(*args, **kwargs):
     _ensure_event_loop()
     return await _original_connect_with_signer(*args, **kwargs)


JssDWt (commented):
Looking good as a general solution. I think the function signatures need something for IDE support / type checking, like functools.wraps or ParamSpec? So that the user gets intellisense for these functions.

dangeross (approved):
LGTM

PR #670, opened 2026-02-26
Detect lightning address changes in real time

### LINKED ISSUE #648: Allow detect lightning address changes in runtime
Allow one instance detect lightning address changes in another instance immediately.

danielgranhao:
Closes #648

The idea is to use rtsync to deliver a simple "notification". Upon receiving it, instances recover the address from the server, keeping it as the single source of truth.

### DIFF
--- a/crates/breez-sdk/core/src/realtime_sync/storage.rs
+++ b/crates/breez-sdk/core/src/realtime_sync/storage.rs
@@ -269,7 +269,7 @@ impl SyncedStorage {
                     .await
             }
             RecordType::LightningAddress => {
-                return Ok(self.handle_lightning_address_change().await);
+                return Ok(self.handle_lightning_address_change());
             }
         }?;
         Ok(RecordOutcome::Completed)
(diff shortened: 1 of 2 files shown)


docs/breez-sdk/snippets/rust/src/getting_started.rs
```diff
@@ -85,6 +85,11 @@ impl EventListener for SdkEventListener {
             SdkEvent::Optimization { optimization_event } => {
                 // An optimization event occurred
             }
+            SdkEvent::LightningAddressChanged {
+                lightning_address,
+            } => {
+                // The lightning address has changed
+            }
```
dangeross: Should we add a note to the receive_lnurl_pay docs about handling the LightningAddressChanged event?

crates/breez-sdk/core/src/realtime_sync/storage.rs
```diff
@@ -341,17 +354,87 @@ impl SyncedStorage {
...
+
+        let cache = ObjectCacheRepository::new(Arc::clone(&self.inner));
+        let old = cache
+            .fetch_lightning_address()
+            .await
+            .ok()
+            .flatten()
+            .flatten();
+
+        let resp = match client.recover_lightning_address().await {
```
JssDWt: I'm looking at whether we can remove the network request from the pull_sync_once_local loop.

PR #688, opened 2026-03-03
lnurl postgres tests

JssDWt:
The last migration added in #687 had an invalid version number (same as the previous number 20260303). That is fixed in this PR. 

In order to ensure there are no bugs on postgres alone, the LNURL server is now also tested with the postgres backend in the LNURL flow integration tests.


dangeross (approved):
LGTM

PR #775, opened 2026-04-03
Send all with stable balance

dangeross:
Send-all with stable balance works by combining token conversion with `FeesIncluded` fee policy. The user provides their full token balance as the amount, sets a `token_identifier`, and specifies `ConversionOptions` with `ToBitcoin` conversion type along with `FeePolicy::FeesIncluded`. During prepare, the SDK estimates the total sats available by converting all tokens to sats and adding any existing sat balance, then deducts fees. At send time, the SDK re-evaluates whether this is still a send-all based on current balance, if so converts the full token amount using `AmountIn` (rather than `MinAmountOut`), then queries the actual post-conversion sat balance and sends the entire balance minus fees - effectively draining the wallet completely

**Breaking change:**
- PrepareLnurlPayRequest
  - Replaces `amount_sats` with `amount`
  - Adds `token_identifier` 

Closes #770

### DIFF
--- a/crates/breez-sdk/cli/src/command/mod.rs
+++ b/crates/breez-sdk/cli/src/command/mod.rs
@@ -659,7 +659,7 @@ pub(crate) async fn execute_command(
                     };
                 println!(
                     "Estimated conversion of {} {} with a {} {} fee",
-                    conversion_estimate.amount, units, conversion_estimate.fee, units
+                    conversion_estimate.amount_out, units, conversion_estimate.fee, units
                 );
                 let line = rl
                     .readline_with_initial("Do you want to continue (y/n): ", ("y", ""))?
@@ -713,14 +713,15 @@ pub(crate) async fn execute_command(
                     let max_sendable = pay_request.max_sendable / 1000;
                     let prompt =
                         format!("Amount to pay (min {min_sendable} sat, max {max_sendable} sat): ");
-                    let amount_sats = rl.readline(&prompt)?.parse::<u64>()?;
+                    let amount = rl.readline(&prompt)?.parse::<u128>()?;
 
                     let prepare_response = sdk
                         .prepare_lnurl_pay(PrepareLnurlPayRequest {
-                            amount_sats,
+                            amount,
                             comment,
                             pay_request,
                             validate_success_action_url: validate_success_url,
+                            token_identifier: None,
                             conversion_options,
                             fee_policy,
                         })
@@ -729,7 +730,7 @@ pub(crate) async fn execute_command(
                     if let Some(conversion_estimate) = &prepare_response.conversion_estimate {
                         println!(
                             "Estimated conversion of {} token base units with a {} token base units fee",
-                            conversion_estimate.amount, conversion_estimate.fee
+                            conversion_estimate.amount_out, conversion_estimate.fee
                         );
                         let line = rl
                             .readline_with_initial("Do you want to continue (y/n): ", ("y", ""))?
--- a/crates/breez-sdk/cli/src/command/mod.rs
+++ b/crates/breez-sdk/cli/src/command/mod.rs
@@ -188,6 +188,10 @@ pub enum Command {
         #[arg(short = 'i', long)]
         idempotency_key: Option<String>,
 
+        /// If provided, the amount can be provide in the token units.
+        #[arg(short = 't', long)]
+        token_identifier: Option<String>,
+
         // If provided, the payment will include a token conversion step, converting from the
         // specified token to Bitcoin to fulfill the payment.
         #[arg(long = "from-token")]
@@ -658,8 +662,8 @@ pub(crate) async fn execute_command(
                         "token base units"
                     };
                 println!(
-                    "Estimated conversion of {} {} with a {} {} fee",
-                    conversion_estimate.amount_out, units, conversion_estimate.fee, units
+                    "Estimated conversion of {} {} with a {} token base units fee",
+                    conversion_estimate.amount_out, units, conversion_estimate.fee
                 );
                 let line = rl
                     .readline_with_initial("Do you want to continue (y/n): ", ("y", ""))?
@@ -687,6 +691,7 @@ pub(crate) async fn execute_command(
             comment,
             validate_success_url,
             idempotency_key,
+            token_identifier,
             convert_from_token_identifier,
             convert_max_slippage_bps: max_slippage_bps,
             fees_included,
@@ -711,8 +716,13 @@ pub(crate) async fn execute_command(
                 | InputType::LnurlPay(pay_request) => {
                     let min_sendable = pay_request.min_sendable.div_ceil(1000);
                     let max_sendable = pay_request.max_sendable / 1000;
-                    let prompt =
-                        format!("Amount to pay (min {min_sendable} sat, max {max_sendable} sat): ");
+                    let prompt = if token_identifier.is_none() {
+                        format!("Amount to pay (min {min_sendable} sat, max {max_sendable} sat): ")
+                    } else {
+                        format!(
+                            "Amount to pay (min {min_sendable} sat, max {max_sendable} sat) in token units: "
+                        )
+                    };
                     let amount = rl.readline(&prompt)?.parse::<u128>()?;
 
                     let prepare_response = sdk
@@ -721,16 +731,23 @@ pub(crate) async fn execute_command(
                             comment,
                             pay_request,
                             validate_success_action_url: validate_success_url,
-                            token_identifier: None,
+                            token_identifier,
                             conversion_options,
                             fee_policy,
                         })
                         .await?;
 
                     if let Some(conversion_estimate) = &prepare_response.conversion_estimate {
+                        let units = if conversion_estimate.options.conversion_type
+                            == ConversionType::FromBitcoin
+                        {
+                            "token base units"
+                        } else {
+                            "sats"
+                        };
                         println!(
-                            "Estimated conversion of {} token base units with a {} token base units fee",
-                            conversion_estimate.amount_out, conversion_estimate.fee
+                            "Estimated conversion of {} {} with a {} token base units fee",
+                            conversion_estimate.amount_out, units, conversion_estimate.fee
                         );
                         let line = rl
                             .readline_with_initial("Do you want to continue (y/n): ", ("y", ""))?
--- a/crates/breez-sdk/cli/src/command/mod.rs
+++ b/crates/breez-sdk/cli/src/command/mod.rs
@@ -188,7 +188,7 @@ pub enum Command {
         #[arg(short = 'i', long)]
         idempotency_key: Option<String>,
 
-        /// If provided, the amount can be provide in the token units.
+        /// If provided, the amount will be denominated in token base units.
         #[arg(short = 't', long)]
         token_identifier: Option<String>,
 
@@ -655,15 +655,19 @@ pub(crate) async fn execute_command(
             };
 
             if let Some(conversion_estimate) = &prepare_response.conversion_estimate {
-                let units =
+                let (in_units, out_units) =
                     if conversion_estimate.options.conversion_type == ConversionType::FromBitcoin {
-                        "sats"
+                        ("sats", "token base units")
                     } else {
-                        "token base units"
+                        ("token base units", "sats")
                     };
                 println!(
-                    "Estimated conversion of {} {} with a {} token base units fee",
-                    conversion_estimate.amount_out, units, conversion_estimate.fee
+                    "Estimated conversion from {} {} to {} {} with a {} token base units fee",
+                    conversion_estimate.amount_in,
+                    in_units,
+                    conversion_estimate.amount_out,
+                    out_units,
+                    conversion_estimate.fee
                 );
                 let line = rl
                     .readline_with_initial("Do you want to continue (y/n): ", ("y", ""))?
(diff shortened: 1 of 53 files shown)


crates/breez-sdk/core/src/sdk/payments.rs
```diff
@@ -1380,31 +1553,303 @@ impl BreezSdk {
...
+                    return Err(SdkError::Generic(
+                        "Request token identifier must match conversion options".to_string(),
+                    ));
+                }
+                let token_balances = self.spark_wallet.get_token_balances().await?;
+                let token_balance = token_balances.get(token_id).map_or(0, |tb| tb.balance);
+                amount == token_balance && fee_policy == FeePolicy::FeesIncluded
+            }
+            None => false,
+        };
```
danielgranhao, line 1665: IIUC, we'll send all even when stable balance is disabled. If it is disabled, the payment shouldn't spend from existing sats balance, right?

crates/breez-sdk/core/src/sdk/lnurl.rs
```diff
@@ -152,6 +203,15 @@ impl BreezSdk {
             None
         };
 
+        // For send-all with conversion, pass through FeesIncluded so the conversion
+        // flow queries actual balance post-conversion. Otherwise use FeesExcluded.
+        let has_conversion = request.prepare_response.conversion_estimate.is_some();
```
danielgranhao, line 188: can be removed, already computed above

crates/breez-sdk/core/src/sdk/lnurl.rs
```diff
@@ -132,11 +178,16 @@ impl BreezSdk {
...
+            // With conversion, skip this check — the slippage buffer deducted at prepare
+            // time means the actual post-conversion balance may exceed the invoice amount,
+            // and we want to overpay the invoice to send everything.
+            if !has_conversion {
+                let max_allowed_overpayment = current_fee.max(1);
+                if overpayment > max_allowed_overpayment {
+                    return Err(SdkError::Generic(format!(
+                        "Fee overpayment ({overpayment} sats) exceeds allowed maximum ({max_allowed_overpayment} sats)"
+                    )));
+                }
```
danielgranhao: It concerns me a bit that we end up allowing for an overpayment as big as the slippage buffer. IIUC, assuming constant conversion rates, this will always result in a fee overpayment to the SSP given by the slippage tolerance. Is this correct?

crates/breez-sdk/core/src/sdk/lnurl.rs
```diff
@@ -20,26 +20,70 @@ use super::{BreezSdk, helpers::process_success_action};
...
+            let amount_after_slippage = estimated_sats.saturating_sub(slippage_buffer);
+            trace!(
+                "prepare_lnurl: token_amount={}, estimated_sats={}, slippage_bps={}, slippage_buffer={}, amount_for_invoice={}",
+                request.amount,
+                estimated_sats,
+                max_slippage_bps,
+                slippage_buffer,
+                amount_after_slippage
+            );
+            (amount_after_slippage, FeePolicy::FeesIncluded)
```
danielgranhao: When `is_token_conversion && token_identifier.is_some()`, fee_policy is always overridden to `FeesIncluded`, even if the user passed `FeesExcluded` or default. This applies to partial token amounts too, not just send-all. Maybe `FeesExcluded` doesn't make so much sense, but if that's the case, should we explicitly disallow it and return error?

crates/breez-sdk/core/src/sdk/payments.rs
```diff
@@ -1380,31 +1553,303 @@ impl BreezSdk {
...
+            )
+            .await?;
+
+        let total_u64: u64 = estimated_sats.try_into()?;
+        if total_u64 <= lightning_fee_sats {
+            return Err(SdkError::InvalidInput(
+                "Amount too small to cover fees".to_string(),
+            ));
+        }
+
```
danielgranhao, line 1828: Probably I'm missing something, but shouldn't this check for bolt11 invoice amount + fees instead of just fees?

crates/breez-sdk/core/src/sdk/payments.rs
dangeross: Nice catch. Fixed

crates/breez-sdk/core/src/sdk/lnurl.rs
dangeross: Nice catch, I assumed overpayment went to the recipient not the SSP. Changed to just use the standard fees included flow. In tests fees are about x1.5 the estimate, where x0.5 goes to the SSP

crates/breez-sdk/core/src/sdk/lnurl.rs
dangeross: Fixed to error out if FeesExcluded is explicitly set

crates/breez-sdk/core/src/sdk/lnurl.rs
```diff
@@ -20,26 +20,80 @@ use super::{BreezSdk, helpers::process_success_action};
...
+                    request.conversion_options.as_ref(),
+                    fee_policy,
+                )
+                .await?
+            }
+            None => false,
+        };
+
+        // token_identifier with conversion is only valid for send-all
+        if request.token_identifier.is_some() && !is_send_all_with_conversion {
```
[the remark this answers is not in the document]
roeierez: That's great Ross!

crates/breez-sdk/cli/src/command/mod.rs
```diff
@@ -188,6 +188,10 @@ pub enum Command {
         #[arg(short = 'i', long)]
         idempotency_key: Option<String>,
 
+        /// If provided, the amount can be provide in the token units.
```
danielgranhao:
```suggestion
        /// If provided, the amount will be denominated in token base units.
```

roeierez (approved):
LGTM

crates/breez-sdk/core/src/sdk/lnurl.rs
```diff
@@ -152,6 +181,14 @@ impl BreezSdk {
             None
         };
 
+        // For send-all with conversion, pass through FeesIncluded so the conversion
+        // flow queries actual balance post-conversion. Otherwise use FeesExcluded.
+        let internal_fee_policy = if is_fees_included && has_conversion {
+            FeePolicy::FeesIncluded
+        } else {
+            FeePolicy::FeesExcluded
+        };
```
danielgranhao: The comment doesn't fully match the code (mentions send-all but that is not checked). 

Also, I'm finding it a bit hard to reason about this. Wondering if there's something we can do to simplify this. I'm ok with leaving it for a separate PR though.

crates/breez-sdk/core/src/sdk/lnurl.rs
```diff
@@ -378,17 +415,24 @@ impl BreezSdk {
...
     pub(super) async fn prepare_lnurl_pay_fees_included(
         &self,
         request: PrepareLnurlPayRequest,
-        amount_sats: u64,
+        amount: u128,
+        conversion_estimate: Option<crate::ConversionEstimate>,
     ) -> Result<PrepareLnurlPayResponse, SdkError> {
+        let amount_sats: u64 = amount
+            .try_into()
+            .map_err(|_| SdkError::InvalidInput("Amount too large".to_string()))?;
```
danielgranhao: If `amount` is always sats, I think the method signature should reflect that and be kept as `amount_sats: u64`

crates/breez-sdk/core/src/sdk/lnurl.rs
danielgranhao, line 547: I noticed now that this would now return a sat amount for a prepare request that asked for tokens to be sent. Should we replace `amount_sats` with `amount` that is the same as the requested one?

crates/breez-sdk/core/src/sdk/payments.rs
```diff
@@ -1380,31 +1553,318 @@ impl BreezSdk {
...
+        Ok(Some(PrepareSendPaymentResponse {
+            payment_method: SendPaymentMethod::Bolt11Invoice {
+                invoice_details: invoice.clone(),
+                spark_transfer_fee_sats,
+                lightning_fee_sats,
+            },
+            amount: estimated_sats,
```
danielgranhao, line 1835: The amount should be the requested token amount otherwise we break the contract of `amount` denomination being determined by the `token_identifier` field

crates/breez-sdk/core/src/sdk/payments.rs
```diff
@@ -1380,31 +1553,318 @@ impl BreezSdk {
...
+                "Amount is below the minimum of {dust_limit_sats} sats required for this address after lowest fees of {min_fee_sats} sats"
+            )));
+        }
+
+        Ok(Some(PrepareSendPaymentResponse {
+            payment_method: SendPaymentMethod::BitcoinAddress {
+                address: withdrawal_address.clone(),
+                fee_quote,
+            },
+            amount: estimated_sats,
```
danielgranhao, line 1900: same here

crates/breez-sdk/core/src/sdk/payments.rs
danielgranhao, line 174: same here: we are returning amount in sats even when input was tokens

docs/breez-sdk/snippets/csharp/SendPayment.cs
```diff
@@ -138,6 +138,54 @@ async Task PrepareSendPaymentFeesIncluded(BreezSdk sdk)
...
         }
 
+        async Task PrepareSendPaymentSendAll(BreezSdk sdk)
+        {
+            // ANCHOR: prepare-send-payment-send-all
+            // To send the entire token balance plus any remaining sats,
+            // provide the full token balance as the amount with ToBitcoin
+            // conversion options and FeesIncluded. The SDK converts all
+            // tokens to sats, combines with existing sat balance, and
+            // deducts fees — draining the wallet completely.
```
danielgranhao: I think this is missing that stable balance must be enabled for the remaining sats to be sent as well

crates/breez-sdk/core/src/sdk/payments.rs
danielgranhao: I see that you did it intentionally. Can't we keep the amount the same as the prepare request and consumers check the sat amount from the conversion estimate?

crates/breez-sdk/core/src/sdk/payments.rs
```diff
@@ -1380,31 +1553,318 @@ impl BreezSdk {
...
+                let token_balance = token_balances.get(token_id).map_or(0, |tb| tb.balance);
+                let has_active_stable_token = match &self.stable_balance {
+                    Some(sb) => sb.get_active_token_identifier().await.as_ref() == Some(token_id),
+                    None => false,
+                };
+                amount == token_balance
+                    && fee_policy == FeePolicy::FeesIncluded
+                    && has_active_stable_token
+            }
+            None => false,
```
danielgranhao, line 1664: At send time this re-reads the token balance and may result in a send all being silently degraded to a normal send. Unless I'm missing something, this goes agaisnt the PR description: "re-validates the token balance hasn't changed"

danielgranhao (commented):
A few more comments :)

crates/breez-sdk/cli/src/command/mod.rs
dangeross: Fixed

crates/breez-sdk/core/src/sdk/lnurl.rs
dangeross: Clarified the comment slightly

crates/breez-sdk/core/src/sdk/lnurl.rs
dangeross: Right, reverted

crates/breez-sdk/core/src/sdk/payments.rs
dangeross: You're right. But I think its needed when sending all token + sats balance. I've decided to amend the contract, which I actually think clarifies behavior, that the response `amount` is **always** the amount to be sent. This is also reflected in the response `token_identifier`. If the user needs to see the conversion details (amount in/out) they can see that in the `conversion_estimate`. 

For example, send all (balance: 250 sats, 1000000 USDB):
- Request: amount = 1000000, token_identifier = USDB, conversion_options = ToBitcoin, fee_policy = FeesIncluded
- Response: amount 1250, token_identifier = None, conversion_estimate = {amount_in = 1000000, amount_out = 1000}

I think this needs your re-approval @roeierez

crates/breez-sdk/core/src/sdk/lnurl.rs
dangeross: Amount to be sent is always sats (see https://github.com/breez/spark-sdk/pull/775#discussion_r3063394982)

crates/breez-sdk/core/src/sdk/payments.rs
dangeross: See https://github.com/breez/spark-sdk/pull/775#discussion_r3063394982

crates/breez-sdk/core/src/sdk/payments.rs
dangeross: See https://github.com/breez/spark-sdk/pull/775#discussion_r3063394982

docs/breez-sdk/snippets/csharp/SendPayment.cs
dangeross: Fixed

crates/breez-sdk/core/src/sdk/payments.rs
dangeross: Clarified the PR description. It behaves like a normal send all, if the balance increases between prepare and send, only the prepared amount is sent (no longer a send-all)

danielgranhao (approved):
LGTM

PR #776, opened 2026-04-03
Expose webhook

### LINKED ISSUE #49: Webhook support for received payments

JssDWt:
Summary                                                                                         
                                                                                                  
  - Notify external systems when an LNURL invoice is paid via a configurable per-domain webhook   
  - Webhook URLs are configured in the domain_webhooks table (manually populated)                 
  - The webhook delivery system (webhooks/ module) is generic and reusable for future webhook     
  types — it only knows about identifiers, URLs, and JSON payloads                                
  - Payload uses the { "template": "...", "data": { ... } } envelope, same as other Breez webhooks
  - amount_sat is nullable because the SSP webhook doesn't always include htlc_amount (e.g. older 
  SSP versions, or non-standard payment flows). Not ideal, but I didn't find a way around it.     
                                                                                                  
  Delivery guarantees                                                                             
                                                                                                  
  - Deliveries are persisted to the database before being sent, so nothing is lost on crash       
  - Failed deliveries are retried with exponential backoff (30s base, 1.5x multiplier, no cap)
  - Delivery records are kept for a configurable TTL (default 90 days) for debugging, then cleaned
   up hourly                                                                                      
                                                                                                  
  Throttling                                                                                      
                                                                                               
  - Per-URL concurrency limit (20 in-flight requests per URL)
  - The claim query returns at most one pending delivery per unique URL per poll cycle, so a
  slow/down domain cannot starve deliveries to other domains                                      
  - If a URL's concurrency limit is reached, excess deliveries are unclaimed and retried next
  cycle                                                                                           
                                                                                               
  Other changes                                                                                   
                                                                                               
  - Renamed newly_paid table to pending_zap_receipts to clarify its purpose                       
  - Moved zap receipt publishing logic from background.rs into zap.rs
  - Added domain and amount_received_sat columns to the invoices table                            
  - Parse htlc_amount from the SSP webhook payload 

Fixes #49


crates/breez-sdk/lnurl/src/postgresql/repository.rs
```diff
@@ -546,4 +562,181 @@ impl crate::repository::LnurlRepository for LnurlRepository {
...
+                 SELECT id, ROW_NUMBER() OVER (PARTITION BY url ORDER BY next_retry_at ASC) AS rn
+                 FROM webhook_deliveries
+                 WHERE next_retry_at <= $1
+                   AND succeeded_at IS NULL
+                   AND COALESCE(claimed_at, 0) < $3
+             )
+             UPDATE webhook_deliveries
+             SET claimed_at = $2
+             WHERE id IN (SELECT id FROM candidates WHERE rn = 1)
+             RETURNING id, identifier, url, payload, created_at, retry_count, next_retry_at",
```
roeierez, line 673: Seems like in take_pending_zap_receipts we use a different strategy (simpler) without the nested query. Do you think it fits here also?

JssDWt: In this webhook design I wanted to ensure that different domains don't affect eachother. So a slow webhook recipient does not affect a fast webhook recipient. Also I limit the amount of in-flight webhooks to 20 currently, per domain. 

Now the problem is which webhook deliveries do you pick up? 

Situation:
- domain is saturated (20 requests outstanding)
- 20 new webhooks for that domain pending in the db as well.

If you pick up, say, 4 webhook deliveries at a time, and the ordering is the same, you would keep picking up the same webhook deliveries for that saturated domain. Since you won't be able to process them, and you will keep picking up the same, other domains will be affected (they don't be picked up). This query picks up 1 webhook delivery _per domain_ at a time. Another option could have been random ordering, but that doesn't index well.

We didn't have this problem in the zap receipt processing, because these limits were not applied there (yet?).

crates/breez-sdk/lnurl/src/webhooks/background.rs
```diff
@@ -0,0 +1,633 @@
...
+        trigger_rx,
+    ));
+    tokio::spawn(webhook_cleanup_processor(db, webhook_delivery_ttl_days));
+}
+
+/// Calculate the next retry delay with exponential backoff.
+#[allow(clippy::cast_precision_loss, clippy::cast_possible_truncation)]
+pub(crate) fn next_retry_delay(retry_count: i32) -> i64 {
+    (BASE_RETRY_DELAY_MS as f64 * RETRY_MULTIPLIER.powi(retry_count)) as i64
+}
```
dangeross, line 50: I get we want to be flexible with delivering webhook notifications, but should we add an upper limit in case we get zombied webhook deliveries? Something reasonable like a month?

crates/breez-sdk/lnurl/src/postgresql/repository.rs
```diff
@@ -546,4 +562,181 @@ impl crate::repository::LnurlRepository for LnurlRepository {
...
+    ) -> Result<Vec<WebhookDelivery>, WebhookRepositoryError> {
+        let now = now_millis();
+        let stale_threshold = now.saturating_sub(300_000); // 5 minutes
+        let rows = sqlx::query(
+            "WITH candidates AS (
+                 SELECT id, ROW_NUMBER() OVER (PARTITION BY url ORDER BY next_retry_at ASC) AS rn
+                 FROM webhook_deliveries
+                 WHERE next_retry_at <= $1
+                   AND succeeded_at IS NULL
+                   AND COALESCE(claimed_at, 0) < $3
```
dangeross: Should we for update skip locked here, else seems redundant?

crates/breez-sdk/lnurl/src/webhooks/background.rs
JssDWt: They implicitly get an upper limit from the `webhook_delivery_ttl_days`. After 90 days the webhook is deleted by the cleanup. Is that ok?

dangeross: Ah, cool beans. Missed that

crates/breez-sdk/lnurl/src/postgresql/repository.rs
JssDWt: It's not redundant, because the claimed_at lock should live _after_ this query. But you're right about the fact that multiple instances could pick up the same rows here. For update skip locked here would lock all rows though, which is not what we want. I'll cook on this.

JssDWt: I think it would become this:

```
  UPDATE webhook_deliveries                                                                       
  SET claimed_at = $2
  WHERE id IN (                                                                                   
      SELECT d.id                                                                               
      FROM (                                                                                      
          SELECT DISTINCT url
          FROM webhook_deliveries                                                                 
          WHERE next_retry_at <= $1                                                             
            AND succeeded_at IS NULL
            AND COALESCE(claimed_at, 0) < $3
      ) urls
      CROSS JOIN LATERAL (                                                                        
          SELECT id
          FROM webhook_deliveries                                                                 
          WHERE url = urls.url                                                                  
            AND next_retry_at <= $1
            AND succeeded_at IS NULL
            AND COALESCE(claimed_at, 0) < $3
          ORDER BY next_retry_at ASC                                                              
          FOR UPDATE SKIP LOCKED
          LIMIT 1                                                                                 
      ) d                                                                                       
  )
  RETURNING id, identifier, url, payload, created_at, retry_count, next_retry_at
```

JssDWt: Updated

docs/breez-sdk/src/guide/lnurl_webhooks.md
```diff
@@ -0,0 +1,60 @@
...
+
+![Webhook flow](images/lnurl_webhook_flow.svg)
+
+As an example, if you want to send push notifications to your users, you could run a Notification Delivery Service (NDS) that receives the webhook and forwards a push notification to the user's device:
+
+![NDS push notification flow](images/lnurl_webhook_nds.svg)
+
+## Getting started
+
+To start receiving webhooks, provide your webhook endpoint URL to Breez. Breez will configure it for your domain so that all Lightning Address payments on that domain trigger a POST request to your endpoint.
```
danielgranhao: Should we be specific and suggest sending an email to contact@breez.technology like we do for lnurl domains?

docs/breez-sdk/src/guide/lnurl_webhooks.md
danielgranhao, line 1: Maybe nice to link to this page from `receive_lnurl_pay.md`

docs/breez-sdk/src/SUMMARY.md
```diff
@@ -55,6 +54,12 @@
 
 ---
 
+# Lightning Address webhooks
+
+- [Lightning Address payment notifications](guide/lnurl_webhooks.md)
+
```
danielgranhao: Could this be placed under/subpage of `- [Receiving payments using LNURL-Pay/Lightning address](guide/receive_lnurl_pay.md)`? If this was already discussed and decision was to put it here, fine by me

danielgranhao (commented):
Looking good!

docs/breez-sdk/src/guide/lnurl_webhooks.md
```diff
@@ -0,0 +1,60 @@
...
+
+When one of your users receives a payment to their Lightning Address, Breez can send a webhook to your server. Payments are received automatically without any user interaction — the webhook simply lets you know it happened. For example, you could send the user a push notification, update a balance in your backend, trigger a fulfillment flow, or log the event for analytics.
+
+## How it works
+
+Your users' Lightning Addresses are served by the Breez LNURL server. When a payment comes in, the LNURL server sends a webhook to your server.
+
+![Webhook flow](images/lnurl_webhook_flow.svg)
+
+As an example, if you want to send push notifications to your users, you could run a Notification Delivery Service (NDS) that receives the webhook and forwards a push notification to the user's device:
```
dangeross, line 11: Should we link to our notify service? or best not because it doesn't yet handle this template type

docs/breez-sdk/src/guide/lnurl_webhooks.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Payload
+
+All payloads use a `{ "template": "...", "data": { ... } }` envelope. Currently the only template is `payment_received`:
+
+```json
+{
+  "template": "payment_received",
+  "data": {
+    "paymentHash": "abc123...",
```
dangeross: I was checking the notify service and it accepts "payment_received" template, but the "paymentHash" needs to be snake case (this is just for information)

docs/breez-sdk/src/guide/lnurl_webhooks.md
JssDWt: I thought not, because it doesn't handle this template type?

docs/breez-sdk/src/guide/lnurl_webhooks.md
JssDWt: Changed the template to `spark_payment_received` and the fields to snake_case.

dangeross (approved):
LGTM

danielgranhao (approved):
LGTM

roeierez (approved):

PR #778, opened 2026-04-04
add Unknown fallback for backend enum parsing

pulkit-999:
fix: Added Unknown fallback for backend enum parsing

- Added Unknown variant to SparkHtlcStatus, TokenTransactionType and ConversionStatus
- Replaced FromStr with From<&str> for safe parsing
- Replaced .parse() usages with Enum::from(...)
- Updated FromSql implementations to avoid failures on unknown values
- Added tracing warnings for unrecognized values

This PR closes the issue #773.

### DIFF
--- a/crates/breez-sdk/core/src/models/mod.rs
+++ b/crates/breez-sdk/core/src/models/mod.rs
@@ -415,6 +415,7 @@ pub enum TokenTransactionType {
     Transfer,
     Mint,
     Burn,
+    Unknown,
 }
 
 impl fmt::Display for TokenTransactionType {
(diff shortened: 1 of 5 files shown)


danielgranhao:
Thanks for the PR @pulkit-999. On review, we realized #773 was mis-scoped and this change doesn't actually close the forward-compat hole it targets.

The enums touched here (`SparkHtlcStatus`, `TokenTransactionType`, `ConversionStatus`) are never parsed from backend responses at runtime — their `FromStr` impls only exist for round-tripping values through our own local SQLite/Postgres database. The values themselves are derived internally from typed proto enums (`TransferStatus`, `PreimageRequestStatus`, …) or from internal SDK logic, so the added `Unknown` variants are unreachable in practice.

The actual leak is one layer up, at the prost boundary: prost's enum accessors (e.g. `transfer.status()`) silently collapse unknown variants to variant 0 via `try_from(i32).unwrap_or_default()`. For our enums, variant 0 is usually a plausible-but-wrong state (`TransferStatus::SenderInitiated`, `PreimageRequestStatus::WaitingForPreimage`), so a new backend variant gets silently mis-categorized before it ever reaches the types this PR modifies.

Closing this in favor of #791, which reframes the fix at the prost boundary. Thanks again for taking a look.

PR #779, opened 2026-04-06
Lib: Add partner ID per gRPC request

hydra-yse:
This PR:
- Changes the `spark` crate to allow passing custom headers to `Session`
- Adds a wrapper around `SessionManager` in the SDK that allows for fetching of a partner_id JWT

### DIFF
--- a/crates/breez-sdk/core/src/sdk/sync.rs
+++ b/crates/breez-sdk/core/src/sdk/sync.rs
@@ -5,6 +5,10 @@ use std::sync::Arc;
 use tokio::sync::watch;
 use tracing::{Instrument, debug, error, info, trace, warn};
 
+use crate::Network;
+use crate::session_manager::{
+    BreezSessionManager, KEY_BREEZ_JWT, calculate_expiry, is_jwt_expired, jwt_exp,
+};
 use crate::{
     DepositInfo, InputType, MaxFee, PaymentDetails, PaymentType,
     error::SdkError,
@@ -27,7 +31,33 @@ use super::{
     parse_input,
 };
 
+const JWT_REFRESH_RETRY_SECS: u64 = 10;
+
 impl BreezSdk {
+    async fn jwt_interval(&self, refresh_counter: u8) -> Duration {
+        let mut token = self.session_manager.token.read().await.clone();
+
+        if token.is_none() {
+            token = self
+                .storage
+                .get_cached_item(KEY_BREEZ_JWT.to_string())
+                .await
+                .ok()
+                .flatten();
+            if let Some(token) = &token
+                && !is_jwt_expired(token)
+            {
+                *self.session_manager.token.write().await = Some(token.clone());
+            }
+        }
+
+        Duration::from_secs(match (token, refresh_counter) {
+            (None, counter) if counter < 3 => 0,
+            (None, _) => JWT_REFRESH_RETRY_SECS,
+            (Some(token), _) => jwt_exp(&token).map_or(JWT_REFRESH_RETRY_SECS, calculate_expiry),
+        })
+    }
+
     pub(super) fn periodic_sync(&self, initial_synced_sender: watch::Sender<bool>) {
         let sdk = self.clone();
         let mut shutdown_receiver = sdk.shutdown_sender.subscribe();
@@ -35,6 +65,10 @@ impl BreezSdk {
         let sync_coordinator = sdk.sync_coordinator.clone();
         let mut sync_trigger_receiver = sdk.sync_coordinator.subscribe();
         let mut last_sync_time = SystemTime::now();
+        let (should_refresh_jwt, mut refresh_counter) = (
+            matches!(sdk.config.network, Network::Mainnet) && sdk.config.api_key.is_some(),
+            0,
+        );
 
         let sync_interval = u64::from(self.config.sync_interval_secs);
         let span = tracing::Span::current();
@@ -91,6 +125,28 @@ impl BreezSdk {
                             last_sync_time = SystemTime::now();
                         }
                     }
+
+                    () = tokio::time::sleep(sdk.jwt_interval(refresh_counter).await), if should_refresh_jwt => {
+                        refresh_counter = refresh_counter.saturating_add(1);
+                        let api_key = sdk.config.api_key.as_ref().unwrap();
+                        let token = match BreezSessionManager::new_jwt(api_key).await {
+                            Ok(token) => token,
+                            Err(err) => {
+                                warn!("Could not fetch new JWT: {err}");
+                                continue;
+                            }
+                        };
+                        *sdk.session_manager.token.write().await = Some(token.clone());
+                        if let Err(err) = sdk
+                            .storage
+                            .set_cached_item(KEY_BREEZ_JWT.to_string(), token)
+                            .await
+                        {
+                            warn!("Could not persist JWT: {err}");
+                        }
+                        refresh_counter = 0;
+                    }
+
                     // Ensure we sync at least the configured interval
                     () = tokio::time::sleep(Duration::from_secs(10)) => {
                         let now = SystemTime::now();
--- a/crates/breez-sdk/core/src/sdk/sync.rs
+++ b/crates/breez-sdk/core/src/sdk/sync.rs
@@ -1,8 +1,6 @@
 use platform_utils::time::{Duration, Instant, SystemTime};
-use platform_utils::{DefaultHttpClient, HttpClient as _, tokio};
-use serde::Deserialize;
+use platform_utils::tokio;
 use spark_wallet::WalletEvent;
-use std::collections::HashMap;
 use std::sync::Arc;
 use tokio::sync::watch;
 use tracing::{Instrument, debug, error, info, trace, warn};
@@ -13,7 +11,6 @@ use super::{
     parse_input,
 };
 use crate::Network;
-use crate::session_manager::{KEY_BREEZ_JWT, calculate_expiry, is_jwt_expired, jwt_exp};
 use crate::{
     DepositInfo, InputType, MaxFee, PaymentDetails, PaymentType,
     error::SdkError,
@@ -29,58 +26,8 @@ use crate::{
         utxo_fetcher::DetailedUtxo,
     },
 };
-use breez_sdk_common::breez_server::PRODUCTION_BREEZSERVER_URL;
-
-const JWT_REFRESH_RETRY_SECS: u64 = 10;
-
-#[derive(Deserialize)]
-struct JwtResponse {
-    token: String,
-}
 
 impl BreezSdk {
-    pub(crate) async fn new_jwt(api_key: &str) -> Result<String, SdkError> {
-        let client = DefaultHttpClient::new(None);
-        let mut headers = HashMap::new();
-        headers.insert("authorization".to_string(), format!("Bearer {api_key}"));
-        let res = client
-            .get(
-                format!("{PRODUCTION_BREEZSERVER_URL}/api/jwt"),
-                Some(headers),
-            )
-            .await
-            .map_err(|err| SdkError::Generic(format!("Could not retrieve JWT token: {err}")))?;
-
-        let JwtResponse { token } = serde_json::from_str(&res.body).map_err(|err| {
-            SdkError::Generic(format!("Could not parse JWT token response: {err}"))
-        })?;
-        Ok(token)
-    }
-
-    async fn jwt_interval(&self, refresh_counter: u8) -> Duration {
-        let mut token = self.session_manager.get_token().await;
-
-        if token.is_none() {
-            token = self
-                .storage
-                .get_cached_item(KEY_BREEZ_JWT.to_string())
-                .await
-                .ok()
-                .flatten();
-            if let Some(token) = &token
-                && !is_jwt_expired(token)
-            {
-                self.session_manager.set_token(token.clone()).await;
-            }
-        }
-
-        Duration::from_secs(match (token, refresh_counter) {
-            (None, counter) if counter < 3 => 0,
-            (None, _) => JWT_REFRESH_RETRY_SECS,
-            (Some(token), _) => jwt_exp(&token).map_or(JWT_REFRESH_RETRY_SECS, calculate_expiry),
-        })
-    }
-
     pub(super) fn periodic_sync(&self, initial_synced_sender: watch::Sender<bool>) {
         let sdk = self.clone();
         let mut shutdown_receiver = sdk.shutdown_sender.subscribe();
@@ -163,7 +110,7 @@ impl BreezSdk {
                         sdk.session_manager.set_token(token.clone()).await;
                         if let Err(err) = sdk
                             .storage
-                            .set_cached_item(KEY_BREEZ_JWT.to_string(), token)
+                            .set_cached_item(jwt::KEY_BREEZ_JWT.to_string(), token)
                             .await
                         {
                             warn!("Could not persist JWT: {err}");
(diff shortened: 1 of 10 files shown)


crates/breez-sdk/core/src/session_manager.rs
```diff
@@ -0,0 +1,147 @@
...
+
+#[macros::async_trait]
+impl SessionManager for BreezSessionManager {
+    async fn get_session(
+        &self,
+        service_identity_key: &PublicKey,
+    ) -> Result<Session, SessionManagerError> {
+        let mut session = self.inner.get_session(service_identity_key).await?;
+        let mut headers = HashMap::new();
+        match self.get_or_set_partner_id().await {
```
roeierez: This will lock on every payment so it is better to check existing header and expiration in the current session before attempting to fetch it.

roeierez (changes requested):

roeierez (changes requested):

roeierez (changes requested):

crates/breez-sdk/core/src/session_manager.rs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+fn is_jwt_expired(token: &str) -> bool {
+    let Some(exp) = jwt_exp(token) else {
+        return true;
+    };
+    let Ok(duration) = SystemTime::now().duration_since(UNIX_EPOCH) else {
+        return true;
+    };
+    exp.saturating_add(EXPIRY_GRACE_PERIOD_SECS) <= duration.as_secs()
```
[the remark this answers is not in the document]
danielgranhao: Maybe it's worth adding a couple of unit tests so that things like these would be caught. At least to this method and `jwt_exp` looks like it makes sense.

roeierez: Shouldn't this be saturating_sub ?

roeierez (changes requested):

crates/breez-sdk/core/src/session_manager.rs
```diff
@@ -0,0 +1,184 @@
...
+                .refresh_spawn_count
+                .fetch_update(Ordering::Release, Ordering::Acquire, |current| {
+                    (current < MAX_REFRESH_SPAWNS).then_some(current.saturating_add(1))
+                })
+                .is_ok();
+            if can_spawn {
+                let partner_id = self.partner_id.clone();
+                let storage = self.storage.clone();
+                let spawn_count = self.refresh_spawn_count.clone();
+                spawn(async move {
```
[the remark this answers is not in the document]
hydra-yse: We don't want to spawn threads indefinitely though. Even if they all run to completion relatively quickly. This way we can limit the amount of retries

crates/breez-sdk/core/src/session_manager.rs
```diff
@@ -0,0 +1,184 @@
...
+                    let mut headers = HashMap::new();
+                    headers.insert("authorization".to_string(), format!("Bearer {api_key}"));
+                    let res = match client
+                        .get(format!("{BREEZ_SERVER_URL}/api/jwt"), Some(headers))
+                        .await
+                    {
+                        Ok(res) => res,
+                        Err(err) => {
+                            warn!("Could not retrieve JWT token: {err}");
+                            return;
```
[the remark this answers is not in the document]
hydra-yse: Answering for both you and @danielgranhao. That is intended, we don't want to retry indefinitely. If we fail the request three times in a row, that means we can't fetch the parent_id for that session

roeierez:
@hydra-yse we also need to add the header to the so calls (currently it is integrated in the ssp calls).

danielgranhao (commented):
Just one more thing, otherwise LGTM

crates/breez-sdk/core/src/session_manager.rs
roeierez: I don't see this concern resolved

crates/breez-sdk/core/src/session_manager.rs
```diff
@@ -0,0 +1,147 @@
...
+        if let Some(partner_id) = lock.as_ref()
+            && !is_jwt_expired(partner_id)
+        {
+            return Ok(Some(partner_id.clone()));
+        }
+
+        let client = DefaultHttpClient::new(None);
+        let mut headers = HashMap::new();
+        headers.insert("authorization".to_string(), format!("Bearer {api_key}"));
+        let res = client
```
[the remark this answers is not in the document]
roeierez: I am nor a fun of this network request during lock. If breez server is down it will affect critical payment path.
I think a better approach would be make refresh the session in the background (perhaps as part of the sync?) and here use a trivial approach of taking existing partner id.

crates/breez-sdk/core/src/session_manager.rs
```diff
@@ -0,0 +1,228 @@
...
+    let Some(exp) = jwt_exp(token) else {
+        return true;
+    };
+    let Ok(now) = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .map(|t| t.as_secs())
+    else {
+        return true;
+    };
+    exp.saturating_sub(EXPIRY_GRACE_PERIOD_SECS) <= now
```
JssDWt: Checking sub vs add: With 'grace period' we mean the period _before_ expiring when we already consider the token expired?

JssDWt (commented):
Looking good. Just the locking and cache left imo.

crates/breez-sdk/core/src/session_manager.rs
hydra-yse: Yep 👍

roeierez (changes requested):

crates/breez-sdk/core/src/session_manager.rs
```diff
@@ -0,0 +1,305 @@
...
+                            warn!("Could not fetch new JWT: {err}");
+                            None
+                        }
+                    };
+                    // Drain channel
+                    while let Ok(RefreshRequest { reply: sender }) = refresh_rx.try_recv() {
+                        if let Some(token) = &token {
+                            send_reply(sender, token.clone());
+                        }
+                    }
```
hydra-yse: Only opinionated choice I have here is whether we should drain the channel independently from whether or not the server call succeeds. We may be able to fit multiple network requests within the 2-second `REFRESH_TIMEOUT` so we could try multiple times (sequentially) if necessary, and only drain on success/n-failures.

crates/breez-sdk/core/src/session_manager.rs
```diff
@@ -0,0 +1,305 @@
...
+impl BreezSessionManager {
+    pub(crate) fn new(
+        inner: Arc<dyn SessionManager>,
+        config: Config,
+        storage: Arc<dyn Storage>,
+        shutdown_rx: watch::Receiver<()>,
+    ) -> Self {
+        let (refresh_tx, refresh_rx) = mpsc::channel::<RefreshRequest>(40);
+        if let Some(api_key) = &config.api_key {
+            spawn(Self::jwt_refresh_loop(
```
roeierez: did you consider moving the interval check to periodic_sync where where we already have an event loop and a task spawned?
I think we can pass the BreezSessionManager there and it will make sure to invoke new_jwt before expiration and setting it. 
I think it will make it simpler without any extra channel communication mechanism. Also the get_session becomes a lot simpler as no need to refresh the jwt there.

roeierez (changes requested):

crates/breez-sdk/core/src/session_manager.rs
hydra-yse: Check 044ddf9a62e6

crates/breez-sdk/core/src/session_manager.rs
```diff
@@ -0,0 +1,260 @@
...
+            return;
+        }
+
+        let fetch_partner_id = async || -> Result<String, SessionManagerError> {
+            self.storage
+                .get_cached_item(KEY_BREEZ_JWT.to_string())
+                .await
+                .map_err(|err| SessionManagerError::Generic(format!("Could not read JWT: {err}")))?
+                .ok_or(SessionManagerError::Generic("JWT not found".to_string()))
+        };
```
roeierez: This will hit the storage on every request to ssp or so. While sqlite is OK on other storage implementations like postgres it is not efficient. 
I think the session manager should be dumb just taking the token it has in memory and set it as header.
The sync logic that refreshes the token should handle persisting and setting this memory field inside the session manager.

crates/breez-sdk/core/src/sdk/sync.rs
```diff
@@ -91,6 +97,26 @@ impl BreezSdk {
                             last_sync_time = SystemTime::now();
                         }
                     }
+
+                    () = sdk.session_manager.jwt_refresh_interval(is_first_jwt_refresh), if should_refresh_jwt => {
```
roeierez: I don't think we need the jwt_refresh_interval on the session manager. The refresh logic better fits here IMO.
I think we can simplify calculating the interval to refresh by:
1. getting the in memory value from the session manager, if doesn't exist fetch from the persistent storage
2. if token exists, Calculate expiration and return the expiration minus some threshold (5 minutes?) If token doesn't exist return 0 (refresh immediately).

roeierez (changes requested):

crates/breez-sdk/core/src/sdk/sync.rs
```diff
@@ -27,14 +31,44 @@ use super::{
     parse_input,
 };
 
+const JWT_REFRESH_RETRY_SECS: u64 = 10;
+
 impl BreezSdk {
+    async fn jwt_interval(&self, refresh_counter: u8) -> Duration {
+        let mut token = self.session_manager.token.read().await.clone();
```
roeierez: Better to expose get_token/set_token API instead of exposing the lock?

crates/breez-sdk/core/src/session_manager.rs
```diff
@@ -0,0 +1,160 @@
...
+
+impl BreezSessionManager {
+    pub(crate) fn new(inner: Arc<dyn SessionManager>) -> Self {
+        Self {
+            inner,
+            token: RwLock::new(None),
+        }
+    }
+
+    pub(crate) async fn new_jwt(api_key: &str) -> Result<String, SessionManagerError> {
```
roeierez: Should we move this function also to the sync layer?

roeierez (changes requested):
Just a couple more comments, almost there.

crates/breez-sdk/core/src/session_manager.rs
```diff
@@ -0,0 +1,140 @@
...
+    calculate_expiry(exp) == 0
+}
+
+pub(crate) fn jwt_exp(token: &str) -> Option<u64> {
+    let payload_b64 = token.split('.').nth(1)?;
+    let decoded = URL_SAFE_NO_PAD.decode(payload_b64).ok()?;
+    let payload = std::str::from_utf8(&decoded).ok()?;
+    let Jwt { exp } = serde_json::from_str(payload).ok()?;
+    Some(exp)
+}
```
roeierez: These are only used in sync so perhaps move them there as well?

crates/breez-sdk/core/src/session_manager.rs
```diff
@@ -0,0 +1,140 @@
...
+        service_identity_key: &PublicKey,
+    ) -> Result<Session, SessionManagerError> {
+        let mut session = self.inner.get_session(service_identity_key).await?;
+        if !session.headers.contains_key(PARTNER_ID_HEADER)
+            && let Some(token) = self.token.read().await.as_ref()
+        {
+            session
+                .headers
+                .insert(PARTNER_ID_HEADER.to_string(), token.clone());
+        }
```
JssDWt: Is the `partner_id` header the place for this token? I would expect a JWT token on an `Authorization` header. In a `partner_id` header I would expect the partner id. Was this discussed with Spark?

crates/breez-sdk/core/src/sdk/sync.rs
```diff
@@ -20,21 +29,69 @@ use crate::{
...
+                .flatten();
+            if let Some(token) = &token
+                && !is_jwt_expired(token)
+            {
+                self.session_manager.set_token(token.clone()).await;
+            }
+        }
+
+        Duration::from_secs(match (token, refresh_counter) {
+            (None, counter) if counter < 3 => 0,
```
JssDWt: A duration of zero means that the sync loop select will immediately fall into the jwt case? Will it keep falling into that case over and over again?

crates/breez-sdk/core/src/sdk/sync.rs
```diff
@@ -91,6 +148,29 @@ impl BreezSdk {
                             last_sync_time = SystemTime::now();
                         }
                     }
+
+                    () = tokio::time::sleep(sdk.jwt_interval(refresh_counter).await), if should_refresh_jwt => {
```
JssDWt: I feel that this select arm could be a little bit lighter still.  Because that jwt_interval function gets evaluated on every loop, and it may hit the database.

crates/breez-sdk/core/src/session_manager.rs
```diff
@@ -0,0 +1,140 @@
...
+}
+
+#[macros::async_trait]
+impl SessionManager for BreezSessionManager {
+    async fn get_session(
+        &self,
+        service_identity_key: &PublicKey,
+    ) -> Result<Session, SessionManagerError> {
+        let mut session = self.inner.get_session(service_identity_key).await?;
+        if !session.headers.contains_key(PARTNER_ID_HEADER)
```
roeierez: It seems this will not update a token if expired as it only updates if the header doesn't exist.

crates/breez-sdk/core/src/session_manager.rs
```diff
@@ -0,0 +1,140 @@
...
+            && let Some(token) = self.token.read().await.as_ref()
+        {
+            session
+                .headers
+                .insert(PARTNER_ID_HEADER.to_string(), token.clone());
```
JssDWt, line 43: Inside `get_session_interceptor` a call to `get_authenticated_session` is made, passing the session including headers we create here. But if the auth expired, a new session is created, overwriting the old old, and the new one doesn't have headers. We could add them in `set_session`, which is called afterwards. Otherwise every time re-auth happens, the header is not set in the first request.

JssDWt (commented):
Things have simplified a lot since I last looked at this! I like it! There's still some comments:

crates/breez-sdk/core/src/sdk/sync.rs
roeierez: Seems like it will enter here every interval even if we don't have API key (non mainnet)?
Can we change that (perhaps return long time to sleep from jwt_interval in this case) ?

roeierez (changes requested):

crates/breez-sdk/core/src/sdk/sync.rs
hydra-yse: Yep, up to 3 times in a row (also depends on how the runtime schedules tasks). This shouldn't be a problem since we set the counter for fails, so it will fallback to a 5 second delay if we can't fetch the JWT for 3 times in a row

crates/breez-sdk/core/src/sdk/sync.rs
hydra-yse: It will only enter that arm when on Mainnet + with API key. Non-mainnet case is skipped entirely, with or without API key. Do we need to support that?

hydra-yse: Ideally, we only hit db at startup, on first fetch, then following ones are taken from memory. Did you have anything in mind to simplify?

**Edit:** Maybe we can introduce a timeout for the db call. Wdyt?

crates/breez-sdk/core/src/session_manager.rs
hydra-yse: Agreed, nice catch! 74004ca65e96

crates/breez-sdk/core/src/session_manager.rs
hydra-yse: Done 4c02e920ac20

crates/breez-sdk/core/src/session_manager.rs
hydra-yse: Done b940b89b2cce

crates/breez-sdk/core/src/session_manager.rs
hydra-yse: Thought the same, but we already have an Authorization header to authenticate with the SO. cc @roeierez do you confirm we have the correct format?

crates/breez-sdk/core/src/sdk/sync.rs
JssDWt: I think I would feel more relaxed about this if the jwt refresh had its own loop in its own task. Then it doesn't really matter whether it's slow or not, because it's not interfering with other sync.

roeierez: How is it skipped? The call to jwt_interval is invoked every time (3 times in a row on startup as it returns 0)

hydra-yse: We moved it into the sync on purpose, so we could save on spawning new tasks, despite the delay introduced. The goal was to make it light enough so that it doesn't interfere with the sync process at all.
@roeierez wdyt?

crates/breez-sdk/core/src/sdk/sync.rs
JssDWt: What is the advantage of doing this 3 times in a row? Is this a solution for a race? Can you point me to the cause of that race?

crates/breez-sdk/core/src/sdk/sync.rs
hydra-yse: the if should_refresh_jwt condition after the match arm disables it completely, including the sdk.jwt_interval() call

roeierez: I think jwt_interval is called and only then should_refresh_jwt isn't it?

roeierez: I was trying to avoid indeed a separate task but I agree we should make sure it doesn't slow down the sync. We can do a separate task if it is not possible using the sync loop

hydra-yse: Ok, it is indeed called: https://docs.rs/tokio/latest/tokio/macro.select.html
I'll add an instant return in that case
Done b2851a8

hydra-yse: I've moved the database fetching to the beginning of the spawned task, out of the interval calculation: 09c6055

crates/breez-sdk/core/src/session_manager.rs
roeierez: I think spark defined partner_id but we will ask them to make sure.

crates/breez-sdk/core/src/sdk/sync.rs
```diff
@@ -632,3 +660,153 @@ impl BreezSdk {
...
+            }
+        }
+
+        pub(super) async fn jwt_interval(&self, should_refresh_jwt: bool, refresh_counter: u8) {
+            if !should_refresh_jwt {
+                std::future::pending::<()>().await;
+            }
+            let token = self.session_manager.get_token().await;
+            let duration = Duration::from_secs(match (token, refresh_counter) {
+                (None, counter) if counter < 3 => 0,
```
roeierez: The counter mechanism is a bit scary in a way that it may run 3 requests in a row without any backoff. 
Perhaps just use the JWT_REFRESH_RETRY_SECS as long as we don't have a token?

roeierez (commented):
Looks good, One comment.

crates/breez-sdk/core/src/sdk/sync.rs
hydra-yse: Done 26f4043

crates/breez-sdk/core/src/sdk/sync.rs
hydra-yse: Changed in 26f4043

roeierez (approved):
LGTM

crates/breez-sdk/core/src/sdk/sync.rs
```diff
@@ -91,6 +91,30 @@ impl BreezSdk {
                             last_sync_time = SystemTime::now();
                         }
                     }
+
+                    // Only executes on mainnet with API key
+                    () = sdk.jwt_refresh_interval() => {
+                        let Some(api_key) = sdk.config.api_key.as_ref() else {
+                            warn!("Could not refresh JWT: no API key provided");
+                            return;
```
JssDWt: 👀 No api key means exit the function?

JssDWt (commented):
One last thingie maybe?

crates/breez-sdk/core/src/sdk/sync.rs
hydra-yse: Should be `unreachable`, but I've reorganized + switched to `continue` so that worst-case, we enter this arm for no reason every 5 seconds. This should never happen unless there's a bug in tokio.

JssDWt (approved):
LGTM

PR #781, opened 2026-04-07
feat: add built-in PRF provider support for passkeys

erdemyerebasmaz:
Built-in PRF providers and a unified `PasskeyClient` API so integrators can enable passkey login (biometric wallet access, no seed phrase) with minimal platform-specific code. Wallet keys are derived deterministically from a passkey via the WebAuthn PRF extension and are never stored.

## Summary

**Client API (`PasskeyClient`)**
- `register` / `sign_in` to create or unlock a wallet from a passkey
- `connect_with_passkey` for one-call onboarding on iOS/Android: silent sign-in for returning users, automatic fall-through to register for new ones
- `labels` to derive multiple wallets from a single passkey, discoverable through Nostr relays
- `check_availability` to gate passkey UI and surface domain-association problems in one call
- Every flow returns a `PasskeyCredential` (credential ID always, plus user handle / AAGUID / backup flag on registration) for pinning returning users and blocking duplicate registrations

**Built-in PRF providers**
- Web: standalone JS implementation using the WebAuthn PRF extension
- iOS/macOS and Android: native implementations using AuthenticationServices and Credential Manager respectively
- React Native and Flutter: platform channel wrappers that delegate to the native iOS/Android implementations above
- Kotlin KMP: repackages the Android Credential Manager implementation for KMP targets
- Custom `PrfProvider` for hardware keys, FIDO2, or file-backed seeds (the Rust CLI ships YubiKey / FIDO2 / file examples)

**Platform support**
- Default RP ID `keys.breez.technology` for cross-platform credential sharing
- Android 9+ via Play Services, Android 14+ with any compatible authenticator
- iOS 18+ / macOS 15+ with an ObjC bridge: Apple marks PRF types `NS_REFINED_FOR_SWIFT` on Xcode 26, requiring an ObjC helper to access them
- Dart exceptions propagate as typed errors across the flutter_rust_bridge FFI boundary instead of causing panics

## Documentation
- Passkey login guide: Setup (hosting the Web / Android / iOS config), Onboarding (register / sign-in / unified flows), Credential metadata, Managing labels, and PRF providers
- UX guide with recommended onboarding patterns, prompt counts, and error-recovery flows
- Compiled, multi-language snippets (Rust, TypeScript, Swift, Kotlin, Dart, Python, Go, C#, React Native)

## Known issues
- Passkey support depends on the device's authenticator (e.g. Apple Passwords, Google Password Manager). Devices without a compatible authenticator (e.g. GrapheneOS) may not support passkeys.
- Using the default `keys.breez.technology` RP ID requires registering your app with the team. To self-host, serve your own `assetlinks.json` / `apple-app-site-association` (see the Setup guide).


packages/react-native/src/passkey-prf-provider.ts
dangeross: Shouldn't it be PasskeyPrfProvider.ts?

dangeross (commented):
Looking good. Are you planning to add support to the android and kotlin-multiplatform bindings also?

crates/breez-sdk/bindings/langs/swift/Sources/BreezSdkSpark/PlatformPasskeyPrfProvider.swift
```diff
@@ -0,0 +1,282 @@
...
+    /// - Throws: `PasskeyPrfError` if the user cancels or PRF is not supported by the authenticator.
+    public func createPasskey() async throws {
+        try await registerCredential()
+    }
+
+    /// Check if a PRF-capable passkey is available on this device.
+    ///
+    /// - Returns: `true` if the platform supports passkeys with the PRF extension.
+    public func isPrfAvailable() async throws -> Bool {
+        return true // iOS 18+ always supports platform passkeys with PRF
```
roeierez: Is this always true? Shouldn't users need biometric setup as a requirement?

crates/breez-sdk/wasm/js/passkey-prf-provider/index.js
```diff
@@ -0,0 +1,280 @@
...
+            }
+            throw error;
+        }
+    }
+
+    /**
+     * Check if a PRF-capable passkey is available on this device.
+     *
+     * @returns {Promise<boolean>} true if WebAuthn with PRF extension is likely supported.
+     */
```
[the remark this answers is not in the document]
erdemyerebasmaz: [fixup(wasm): move stranded isPrfAvailable JSDoc to the correct method](https://github.com/breez/spark-sdk/pull/781/commits/1f39f2fb0705490a4f9e8b68cb14500c5f5a2575)

packages/react-native/src/passkey-prf-provider.ts
erdemyerebasmaz: Renamed b60c7dd8a47f8ebe3241bf086a0693593d9793fe

packages/react-native/src/passkey-prf-provider.ts
```diff
@@ -0,0 +1,155 @@
...
+   *
+   * @param salt - The salt string to use for PRF evaluation.
+   * @returns The 32-byte PRF output.
+   * @throws If authentication fails, PRF is not supported, or the user cancels.
+   */
+  async derivePrfSeed(salt: string): Promise<Uint8Array> {
+    if (!BreezSdkSparkPasskey) {
+      throw new Error(
+        'BreezSdkSparkPasskey native module not found. ' +
+        'Ensure the react-native package is properly linked.'
```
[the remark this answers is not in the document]
erdemyerebasmaz: Added a more informative error based on platform & OS version 2ec41619384a19fac59a46c65e392cda83ee63f4

packages/wasm/package.json
```diff
@@ -21,6 +21,10 @@
       "./deno": "./deno/breez_sdk_spark_wasm.js",
       "./nodejs": "./nodejs/index.js",
       "./web": "./web/index.js",
+      "./passkey-prf-provider": {
+        "import": "./web/passkey-prf-provider/index.js",
+        "default": "./web/passkey-prf-provider/index.js"
+      },
```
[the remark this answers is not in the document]
erdemyerebasmaz: Bundled on 5060b6eefb5065f476da4249368f6e44f933e06e

crates/breez-sdk/bindings/langs/swift/Sources/BreezSdkSpark/PlatformPasskeyPrfProvider.swift
erdemyerebasmaz: >Is this always true?

No, it's not always true, it's bound by OS version:
```swift
@available(iOS 18.0, macOS 15.0, *)
public class PlatformPasskeyPrfProvider: PasskeyPrfProvider { … }
```

> Shouldn't users need biometric setup as a requirement?

This is an availability check, not usability. It's not PRF providers responsibility to check if the device is ready. OS handles this when auth prompt is requested, either fails or offers alternate auth methods other than biometrics. Clarified it on doc comment e55689bb97f23242cdf150896482456b4dc6dd88.

crates/breez-sdk/bindings/langs/swift/Sources/BreezSdkSpark/PasskeyProvider.swift
```diff
@@ -0,0 +1,282 @@
...
+/// - Associated Domains entitlement: `webcredentials:<rpId>`
+/// - The domain's `apple-app-site-association` must list your app
+///
+/// Example:
+/// ```swift
+/// let prfProvider = PlatformPasskeyPrfProvider()
+/// let passkey = Passkey(prfProvider: prfProvider, relayConfig: nil)
+/// let wallet = try await passkey.getWallet(walletName: "personal")
+/// ```
+@available(iOS 18.0, macOS 15.0, *)
```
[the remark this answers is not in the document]
dangeross: Ok. So before 18.0 the PlatformPasskeyPrfProvider is just not available to use

packages/react-native/android/src/main/java/com/breeztech/breezsdkspark/BreezSdkSparkReactNativePackage.kt
```diff
@@ -10,10 +10,10 @@ import java.util.HashMap
...
 class BreezSdkSparkReactNativePackage : TurboReactPackage() {
   override fun getModule(name: String, reactContext: ReactApplicationContext): NativeModule? {
-    return if (name == BreezSdkSparkReactNativeModule.NAME) {
-      BreezSdkSparkReactNativeModule(reactContext)
-    } else {
-      null
+    return when (name) {
+      BreezSdkSparkReactNativeModule.NAME -> BreezSdkSparkReactNativeModule(reactContext)
+      BreezSdkSparkPasskeyModule.NAME -> BreezSdkSparkPasskeyModule(reactContext)
+      else -> null
```
dangeross, line 16: Is there going to be issues because this is a generated file?

packages/react-native/android/build.gradle
```diff
@@ -133,6 +133,9 @@ dependencies {
   //noinspection GradleDynamicVersion
   implementation "com.facebook.react:react-native:+"
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+  implementation "androidx.credentials:credentials:1.3.0"
+  implementation "androidx.credentials:credentials-play-services-auth:1.3.0"
+  implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:1.8.0"
 }
```
dangeross, line 140: Also a generated file, maybe we have to look at patching these instead

packages/react-native/src/index.tsx
```diff
@@ -15,8 +15,8 @@ export * from './generated/breez_sdk_spark';
 export * from './generated/breez_sdk_spark_bindings';
 
 // Export the platform passkey PRF provider.
-export { PasskeyPrfProvider } from './passkey-prf-provider';
-export type { PasskeyPrfProviderOptions } from './passkey-prf-provider';
+export { PasskeyPrfProvider } from './PasskeyPrfProvider';
+export type { PasskeyPrfProviderOptions } from './PasskeyPrfProvider';
```
dangeross: Also generated

packages/react-native/android/src/main/java/com/breeztech/breezsdkspark/BreezSdkSparkReactNativePackage.kt
erdemyerebasmaz: Created a post-script and a guard mechanism for these edits on generated files d173e7fdb13903676be6339cb385c471922156ee dce1a1a6b29e2269848a9490ac0a2e3d165a923e
and added a section on CLAUDE.md for it to better handle changes made on generated files 238a9551f146d519057322530108c7de01f79439

It can be improved later on by looking at options to resolve issues on codegen level.

packages/react-native/package.json
```diff
@@ -53,11 +53,12 @@
     "prepare": "bob build && yarn plugin:build",
     "release": "release-it --only-version",
     "typecheck": "tsc",
-    "ubrn:android": "ubrn build android --release --config ubrn.config.yaml --and-generate",
+    "ubrn:android": "ubrn build android --release --config ubrn.config.yaml --and-generate && yarn post-ubrn",
     "ubrn:build": "yarn ubrn:ios && yarn ubrn:android",
-    "ubrn:checkout": "ubrn checkout --config ubrn.config.yaml",
+    "ubrn:checkout": "ubrn checkout --config ubrn.config.yaml && yarn post-ubrn",
```
dangeross: Not sure if it should be run here, as it would be pre-applying the git patch file

packages/react-native/package.json
```diff
@@ -53,11 +53,12 @@
     "prepare": "bob build && yarn plugin:build",
     "release": "release-it --only-version",
     "typecheck": "tsc",
-    "ubrn:android": "ubrn build android --release --config ubrn.config.yaml --and-generate",
+    "ubrn:android": "ubrn build android --release --config ubrn.config.yaml --and-generate && yarn post-ubrn",
     "ubrn:build": "yarn ubrn:ios && yarn ubrn:android",
```
dangeross, line 67: I don't mind too much if running "post-ubrn" twice is ok, but "ubrn:build" is the script used during publish and will run "post-ubrn" twice

erdemyerebasmaz: I agree that it's redundant but I find it necessary as both scripts regen `index.tsx`, this is the least regression prone approach imo but we could move `post-ubrn` to `ubrn:build` from `android` & `ios` as currently `build` is the only one used in workflows.

erdemyerebasmaz: Addressed in a75b59daf546a75288a9c630b9066e7245640171

packages/react-native/package.json
erdemyerebasmaz: Addressed in a75b59daf546a75288a9c630b9066e7245640171

docs/breez-sdk/snippets-processor/src/main.rs
```diff
@@ -183,7 +183,7 @@ impl SnippetsProcessor {
         vec![
             ("Rust", "rust", format!("snippets/rust/src/{}.rs", file_base)),
             ("Swift", "swift", format!("snippets/swift/BreezSdkSnippets/Sources/{}.swift", capitalize_first(file_base))),
-            ("Kotlin", "kotlin", format!("snippets/kotlin_mpp_lib/shared/src/commonMain/kotlin/com/example/kotlinmpplib/{}.kt", capitalize_first(file_base))),
+            ("Kotlin", "kotlin", format!("snippets/kotlin_mpp_lib/shared/src/androidMain/kotlin/com/example/kotlinmpplib/{}.kt", capitalize_first(file_base))),
```
dangeross: Doesn't this only apply to Passkey.kt?

docs/breez-sdk/snippets/wasm/passkey-prf-provider.d.ts
dangeross, line 1: Why is it needed? are they also generated in the wasm package?

docs/breez-sdk/snippets-processor/src/main.rs
erdemyerebasmaz: Yes, changed so that Kotlin lookups try `commonMain` first, then fall back to `androidMain`. That keeps every existing snippet working and resolves `Passkey.kt` from `androidMain`.

[fix(docs): try Kotlin commonMain first, fall back to androidMain](https://github.com/breez/spark-sdk/pull/781/commits/8cbef633b5598d656122355cc119dde4b00ef947)

docs/breez-sdk/snippets/wasm/passkey-prf-provider.d.ts
erdemyerebasmaz: It was a workaround for tsx type resolution, but it's no longer needed. Added types condition to the `./passkey-prf-provider` subpath export so consumers get types automatically.

[fix(docs,wasm): expose types for passkey-prf-provider, drop shim](https://github.com/breez/spark-sdk/pull/781/commits/5d13af1ce1341536fc2ad76359018b2646c20d44)

packages/react-native/ios/BreezSdkSparkPasskey.swift
```diff
@@ -0,0 +1,294 @@
...
+            return try await assertionWithPrf(saltData: saltData, rpId: rpId)
+        }
+    }
+
+    private func assertionWithPrf(saltData: Data, rpId: String) async throws -> Data {
+        let provider = ASAuthorizationPlatformPublicKeyCredentialProvider(relyingPartyIdentifier: rpId)
+        let challenge = randomBytes(count: 32)
+        let request = provider.createCredentialAssertionRequest(challenge: challenge)
+
+        let prfInput = ASAuthorizationPublicKeyCredentialPRFAssertionInput(
```
dangeross: This constructs `ASAuthorizationPublicKeyCredentialPRFAssertionInput` / `ASAuthorizationPublicKeyCredentialPRFValues` directly in Swift, and the delegate reads `credential.prf` directly (lines 188, 243, 256). Those are the same PRF APIs that the Swift binding (`PasskeyProvider.swift`) and the Flutter iOS shim (`BreezSdkSparkPasskeyPlugin.swift`) deliberately route through the `PasskeyPRFHelper` ObjC bridge, with the rationale that on Xcode 26 these types are `NS_REFINED_FOR_SWIFT` and their initializers/accessors are no longer reachable from Swift (`PasskeyPRFHelper.h` header comment).

No PRF helper ships under `packages/react-native/ios/`, and the podspec compiles `ios/*.swift`, so an integrator building the RN pod on Xcode 26 would hit this file.

It is not caught in CI: the `react-native` PR job runs on `ubuntu-latest` (tsc only), and the iOS jobs that do use Xcode (`build-bindings-ios.yml`, `publish-react-native.yml`) only cross-compile the Rust UniFFI framework via `cargo build` / xcframework packaging. Nothing runs `pod install` + `xcodebuild` over `ios/*.swift`, so the TurboModule is compiled only in an integrator's app build.

erdemyerebasmaz: Good catch, and thanks for the CI analysis. This is fixed on #909: the RN iOS module was ported to the shared `PasskeyAssertionCore` (52f11cd6b), which routes all PRF construction and extraction through the `PasskeyPRFHelper` ObjC bridge (`PasskeyPRFHelper.h`/`.m` now ship under `packages/react-native/ios/`), the same path the Swift binding and the Flutter shim use. `BreezSdkSparkPasskey.swift` is now a thin RCT bridge that delegates to the core, with no direct use of the `NS_REFINED_FOR_SWIFT` `ASAuthorizationPublicKeyCredentialPRF*` types.

The CI-coverage gap you flagged is separate and still valid: nothing runs `pod install` + `xcodebuild` over `ios/*.swift`, so a future regression here would not be caught. Worth a follow-up to add an Xcode pod-build job.

PR #785, opened 2026-04-08
Add retry logic with exponential backoff to faucet requests

danielgranhao:
## Summary
Add automatic retry logic with exponential backoff to faucet funding requests in both breez-sdk and spark-itest integration tests. This improves reliability when dealing with transient faucet failures.

## Key Changes
- Added `Duration` import for timeout handling
- Added `warn` to tracing imports for retry logging
- Implemented retry mechanism with up to 3 retries (4 total attempts)
- Exponential backoff strategy: 2^attempt seconds between retries (2s, 4s, 8s)
- Extracted core funding logic into new `try_fund_address()` private method
- Main `fund_address()` method now handles retry loop and returns last error if all attempts fail
- Added warning logs indicating retry attempts with backoff duration

## Implementation Details
- Retries use exponential backoff: `Duration::from_secs(2u64.pow(attempt))`
- Backoff only applied after first attempt (attempt > 0)
- Logs include attempt count and total retries for visibility
- Changes applied consistently to both `crates/breez-sdk/breez-itest/src/faucet.rs` and `crates/spark-itest/src/faucet.rs`

### DIFF
--- a/crates/breez-sdk/breez-itest/src/faucet.rs
+++ b/crates/breez-sdk/breez-itest/src/faucet.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 use std::sync::LazyLock;
+use std::time::Duration;
 
 use anyhow::{Context, Result, bail};
 use platform_utils::{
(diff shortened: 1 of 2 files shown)


dangeross (approved):
LGTM

PR #786, opened 2026-04-08
Add user agent support to rtsync and breez server clients

danielgranhao:
## Summary
This PR adds support for custom User-Agent headers in gRPC client communications across both native and WebAssembly transports.

## Key Changes
- **Transport layer enhancements**:
  - Refactored `transport_wasm.rs` from a type alias to a proper `Transport` struct that wraps `tonic_web_wasm_client::Client` and implements `tower_service::Service`
  - Added User-Agent header injection in the WASM transport's `call` method
  - Updated `transport.rs` to accept and apply user agent configuration to the endpoint

- **Client API updates**:
  - Modified `GrpcClient::new()` to accept an optional `user_agent` parameter in both native and WASM implementations
  - Updated `BreezSyncerClient::new()` to accept and forward the user agent parameter to `GrpcClient`
  - Added `user_agent` field to `RealTimeSyncParams` struct

- **SDK integration**:
  - Updated `SdkBuilder` to pass the user agent to real-time sync initialization
  - Updated all test calls to `BreezSyncerClient::new()` to include the new user agent parameter

- **Dependencies**:
  - Added `http` and `tower-service` crate dependencies for WASM target to support the new Transport implementation

## Implementation Details
The WASM transport implementation uses a custom `Transport` struct that intercepts requests and injects the User-Agent header before delegating to the underlying `tonic_web_wasm_client::Client`. The native transport leverages tonic's built-in `user_agent()` endpoint configuration method. This approach maintains consistency across both platforms while respecting their respective capabilities.

https://claude.ai/code/session_01GR3QbjiAz5WvY2hKoaUrRB

### DIFF
--- a/crates/breez-sdk/common/src/grpc/transport_wasm.rs
+++ b/crates/breez-sdk/common/src/grpc/transport_wasm.rs
@@ -21,8 +21,14 @@ impl Service<http::Request<tonic::body::BoxBody>> for Transport {
     }
 
     fn call(&mut self, mut req: http::Request<tonic::body::BoxBody>) -> Self::Future {
+        // Set both `User-Agent` and `X-User-Agent`. Chrome silently drops any
+        // script-set `User-Agent` on Fetch requests (crbug.com/571722), so we
+        // also send `X-User-Agent` to ensure the value reaches the server
+        // cross-browser. Firefox and Safari honor `User-Agent`.
         req.headers_mut()
             .insert("User-Agent", self.user_agent.clone());
+        req.headers_mut()
+            .insert("X-User-Agent", self.user_agent.clone());
         self.inner.call(req)
     }
 }
--- a/crates/breez-sdk/common/src/grpc/transport_wasm.rs
+++ b/crates/breez-sdk/common/src/grpc/transport_wasm.rs
@@ -1,6 +1,31 @@
-use anyhow::Result;
+use std::task::{Context, Poll};
 
-pub type Transport = tonic_web_wasm_client::Client;
+use anyhow::Result;
+use http::HeaderValue;
+use tower_service::Service;
+
+#[derive(Clone)]
+pub struct Transport {
+    inner: tonic_web_wasm_client::Client,
+    user_agent: HeaderValue,
+}
+
+impl Service<http::Request<tonic::body::BoxBody>> for Transport {
+    type Response = http::Response<tonic_web_wasm_client::ResponseBody>;
+    type Error = tonic_web_wasm_client::Error;
+    type Future =
+        <tonic_web_wasm_client::Client as Service<http::Request<tonic::body::BoxBody>>>::Future;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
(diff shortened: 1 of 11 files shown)


crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -482,8 +482,15 @@ impl SdkBuilder {
             }
         };
 
+        let user_agent = format!(
+            "{}/{}",
+            crate::built_info::PKG_NAME,
+            crate::built_info::GIT_VERSION.unwrap_or(crate::built_info::PKG_VERSION),
+        );
+        info!("Building SparkWallet with user agent: {}", user_agent);
```
JssDWt: nit: technically building breez server now. Let's remove SparkWallet from the comment, and call it 'sdk' or something.

crates/breez-sdk/common/src/grpc/transport_wasm.rs
```diff
@@ -1,16 +1,45 @@
...
+        <tonic_web_wasm_client::Client as Service<http::Request<tonic::body::BoxBody>>>::Future;
 
-pub type Transport = tonic_web_wasm_client::Client;
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, mut req: http::Request<tonic::body::BoxBody>) -> Self::Future {
+        req.headers_mut()
+            .insert("User-Agent", self.user_agent.clone());
```
JssDWt, line 29: Noting:
- Chrome will silently drop this header
- On Firefox, this will trigger a http preflight requests where User-Agent (or *) needs to be in the allowed headers list. This is worth some testing I think.

crates/breez-sdk/core/src/sdk_builder.rs
danielgranhao: Done

crates/breez-sdk/common/src/grpc/transport_wasm.rs
danielgranhao: > Chrome will silently drop this header

This was the existing pattern from `crates/spark/src/operator/rpc/transport/grpc_client_wasm.rs`. But yes, seems like you are right on this [chrome issue](https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_request_header). I've adjusted both clients to also set x-user-agent (as done by e.g. the [grpc/grpc-web JS client](https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/grpcwebclientbase.js)).

> On Firefox, this will trigger a http preflight requests where User-Agent (or *) needs to be in the allowed headers list. This is worth some testing I think.

Here we shouldn't have an issue. I've checked both the Breez server and data-sync, and both allow it.

danielgranhao: https://github.com/breez/data-sync/pull/31

JssDWt (approved):
LGTM

hydra-yse (approved):
utACK 👍

dangeross (approved):
LGTM

PR #790, opened 2026-04-09
SSR Support for WASM Package

dangeross:
This PR:
- Adds a Node.js ESM wrapper in the `/nodejs` subpackage
- Generates an SSR subpackage at `/ssr`, stubbing the real web module for node.js server-side rendering
- Updates docs, README and adds SSR example

### DIFF
--- a/crates/xtask/src/package.rs
+++ b/crates/xtask/src/package.rs
@@ -202,32 +202,32 @@ fn package_wasm_target(
     Ok(())
 }
 
-/// Parsed exports from a wasm-bindgen generated JS file.
+/// Parsed exports from a wasm-bindgen generated `.d.ts` file.
 struct WasmExports {
     functions: Vec<String>,
     classes: Vec<String>,
-    reexports: Vec<String>,
 }
 
-/// Parse exported symbols from a wasm-bindgen generated ESM `.js` file.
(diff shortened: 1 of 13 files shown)


danielgranhao (approved):

crates/xtask/src/package.rs
```diff
@@ -198,6 +202,248 @@ fn package_wasm_target(
...
+        classes,
+        reexports,
+    })
+}
+
+/// Parse exported symbols from a wasm-bindgen generated CJS `.js` file (Node.js target).
+///
+/// Recognises the pattern: `module.exports.NAME = ...`
+/// Skips internal wasm-bindgen symbols (`__wbindgen_*`, `__wbg_*`, `__wasm`).
+fn parse_nodejs_exports(js_path: &Path) -> Result<Vec<String>> {
```
JssDWt: These parsers have some brittleness obviously. Perhaps we can combine them into a single one by parsing `breez_sdk_spark_wasm.d.ts`? Both the web and the node.js have this file that contains the same exports. We could use a single parser then. I was hoping we could use some wasm_bindgen hook, but it seems that's not the case.

JssDWt (approved):

PR #795, opened 2026-04-11
Add unilateral exit support to the SDK

JssDWt:
## Summary

- Adds SDK support for preparing and executing a unilateral exit: automatic leaf selection, CPFP fee-bumping PSBTs, an adaptive sweep transaction, and a detection walk that handles outputs already spent on-chain by any known variant (CPFP, direct-tx, direct refund, direct-from-cpfp refund, or future additions). Includes a comprehensive guide and snippets.

## Design decisions

- **Automatic leaf selection** — `prepare_unilateral_exit` greedily picks profitable leaves from the wallet: a leaf is included when its value strictly exceeds the marginal cost of exiting it (CPFP fees for any not-yet-covered ancestor + refund, plus the incremental sweep input fee). Shared ancestors across selected leaves benefit from reduced marginal cost. `list_leaves` is still available for developers who want to inspect leaves independently.
- **Signing** — developers pass a `CpfpSigner`. Bundled `SingleKeySigner` (via the `single_key_cpfp_signer` factory) covers the common case; integrators with a different key-management story implement the trait themselves.
- **Output-spent detection** — the SDK walks each leaf's exit-path outputs via a new `BitcoinChainService::get_outspend` method. Already-spent steps are returned with `cpfp_tx_hex = None` (matching the docs' skip contract) instead of being dropped from the response. Known spenders (`node_tx`, `direct_tx`, `refund_tx`, `direct_from_cpfp_refund_tx`, `direct_refund_tx`) are hydrated from the `TreeNode`; unknown spenders are fetched via `get_transaction_hex` so future protocol additions are handled as long as they pay the leaf's P2TR.
- **Adaptive sweep** — the sweep rebuilds around whichever confirmed spender credits each leaf's P2TR, falling back to the CPFP `refund_tx` default when nothing has landed. Signing uses the leaf's derived P2TR key, which is invariant across refund variants.
- **Fee calculation uses weight units throughout** — fees are computed as `ceil(fee_rate * package_weight / 4)` rather than converting to vbytes first, avoiding rounding errors that caused "min relay fee not met" rejections.
- **Package-level fee rate** — CPFP child transaction fees cover the full parent+child package weight, since the parent transactions have zero fees (ephemeral anchors).
- **Tap tweak in the signer** — `sign_hash_schnorr_with_tweak` always applies the BIP341 tap tweak because the secret key is only accessible inside the signer. The caller passes `Option<TapNodeHash>` (`None` for key-path-only spends) matching the `Address::p2tr` API.
- **Sweep transaction built by the SDK** — the SDK constructs and signs the sweep transaction internally using the leaf-derived keys, so the caller only needs to broadcast it. The caller is responsible for the destination address and dust checks.
- **CPFP change chains to the same address** — each CPFP fee-bump sends change back to the external UTXO's address, and the next CPFP spends from that change output. The remaining value after the last CPFP is the caller's to spend.
- **CSV timelock exposed on the response** — each transaction pair includes `csv_timelock_blocks` (max block-based CSV across all inputs, per BIP68) so the caller knows which transactions require waiting and for how long, without having to parse raw transactions.
- **RBF** — CPFP children signal RBF. To bump stuck packages, call `prepare_unilateral_exit` again with a higher `fee_rate_sat_per_vbyte` and rebroadcast; the caller picks any convenient UTXOs for the new attempt.
- **Private-CA operator pools** — `SparkSigningOperator` gained an optional `ca_cert: Option<Vec<u8>>` field so integrators connecting to a privately-signed operator pool (e.g. local test fixtures) can pin the CA.

## Breaking change

`BitcoinChainService::get_outspend` is a new required trait method (backed by Esplora's `/tx/:txid/outspend/:vout`). Integrator-supplied `BitcoinChainService` impls will need to add it before upgrading.


crates/breez-sdk/cli/src/command/mod.rs
```diff
@@ -1042,3 +1150,122 @@ pub(crate) fn print_value<T: serde::Serialize>(value: &T) -> Result<(), serde_js
 fn serialize<T: serde::Serialize>(value: &T) -> Result<String, serde_json::Error> {
     serde_json::to_string_pretty(value)
 }
+
+/// Sign a CPFP PSBT with the given private key, handling both segwit v0 (ECDSA) and
+/// taproot (Schnorr) inputs, plus ephemeral anchor inputs.
+fn sign_cpfp_psbt(
```
roeierez: This seems a lot of work needed to be done by the user. Did you consider allowing users to pass a signer for the unilateral exit? And providing a default implementation for that signer from signing key?
Seems like if we have such trait we can move all this logic into the SDK and return signed transactions ready to broadcast?

docs/breez-sdk/src/guide/unilateral_exit.md
```diff
@@ -0,0 +1,223 @@
...
+1. **List your leaves** to decide which ones to exit
+2. **Prepare the unilateral exit** — the SDK builds all transactions and CPFP PSBTs
+3. **Sign the CPFP PSBTs** with your external UTXO's private key
+4. **Broadcast the packages** sequentially — each parent+child pair, waiting for confirmation between each
+5. **Wait for timelocks** on the leaf and refund transactions
+6. **Broadcast the sweep transaction** to collect your funds
+
+## Step 1: List your leaves
+
+Before starting a unilateral exit, inspect the leaves in your wallet to decide which ones to recover. Use {{#name list_leaves}} with a minimum value filter to exclude leaves that would cost more in fees than they are worth.
```
roeierez: Can the SDK calculate this cost internally and decide which leaves should be included (in the prepare step)?
It will remove a lot of work from the users which also won't need to handle low level leaves. Perhaps it will allow us to remove the leaves endpoint?

roeierez (commented):
Very nice Jesse! I think worth elaborating a bit about UX, that's where my comments went.

docs/breez-sdk/src/guide/unilateral_exit.md
JssDWt: I find it hard to make the decision which leaves are worth unilaterally exiting, because the cutoff point is not at zero. One person may find it feasible to exit if 80% of their funds are eaten by fees, the other might not. Also the resulting utxo has a cost to spend again. I agree though this interface is complicated for the user. Maybe we can provide some tweakable parameters to control which leaves are feasible to exit?

crates/breez-sdk/cli/src/command/mod.rs
JssDWt: I did not consider passing a signer there. That is a great idea.

docs/breez-sdk/src/guide/unilateral_exit.md
roeierez: Yes I think if some info is needed from the user to take a decision it is a good approach to let us take the decision inside the SDK.
Question: Why would a user would want to leave leaves out of the unilateral exit in case some fund can be taken? Do users have other choice?  I mean excluding them won't make them usable.

JssDWt:
- The unilateral exit api now works by passing an external signer for the CPFP input.
- We now autoselect the leaves to exit based on onchain profitability. Assuming that you would want to get as much funds out of this as possible.
- We now check which transactions have already confirmed, so we don't build an invalid chain of CPFP transactions.
- We now signal for RBF in all transactions

crates/spark/src/services/unilateral_exit.rs
```diff
@@ -246,77 +516,72 @@ fn create_tx_cpfp_psbt(
...
         version: Version::non_standard(3),
         lock_time: LockTime::ZERO,
-        input: inputs,
+        input: tx_inputs,
         output: vec![TxOut {
             value: Amount::from_sat(adjusted_output_value),
-            script_pubkey: output_script_pubkey,
+            script_pubkey: change_script_pubkey.clone(),
         }],
     };
```
dangeross, line 652: Is there a chance here the fee bump tx decreases to an amount that is dust?

docs/breez-sdk/src/guide/unilateral_exit.md
```diff
@@ -0,0 +1,179 @@
...
+
+### CPFP inputs
+
+The CPFP inputs must be UTXOs you control. The CPFP fee-bump chain produces change outputs that go back to the **same address** as the first CPFP input. This address is reused for every change output in the chain, so make sure you use an address you are comfortable reusing (e.g., a dedicated address for this purpose).
+
+The total value of the CPFP inputs must be sufficient to cover all CPFP fees for the selected leaves. If the inputs don't have enough value, the call returns an error.
+
+### Signer
+
+The SDK provides a default `SingleKeySigner` that handles P2WPKH and P2TR inputs from a single private key. If your CPFP inputs require a different signing scheme (e.g., multisig, hardware wallet, or you don't want to pass raw key material into the SDK), you can implement the `CpfpSigner` trait yourself. The trait receives a PSBT (Partially Signed Bitcoin Transaction) and must return it with the external inputs signed.
```
dangeross: Perhaps also add a note about the lack of a Flutter trait

Also maybe "interface" is more generic
```suggestion
The SDK provides a default `SingleKeySigner` that handles P2WPKH and P2TR inputs from a single private key. If your CPFP inputs require a different signing scheme (e.g., multisig, hardware wallet, or you don't want to pass raw key material into the SDK), you can implement the `CpfpSigner` interface yourself. The interface receives a PSBT (Partially Signed Bitcoin Transaction) and must return it with the external inputs signed.
```

docs/breez-sdk/src/guide/unilateral_exit.md
```diff
@@ -0,0 +1,179 @@
...
+Each entry in {{#name leaves}} contains:
+- **{{#name leaf_id}}**: The leaf's unique identifier
+- **{{#name value}}**: The leaf's value in satoshis
+- **{{#name estimated_cost}}**: The estimated marginal cost to exit this leaf (CPFP chain fees + sweep input fee). The actual profit per leaf is `value - estimated_cost`.
+- **{{#name transactions}}**: Ordered list of transactions to broadcast for this leaf (see [Step 2](#step-2-broadcast-the-transaction-packages))
+
+Each transaction in {{#name transactions}} contains:
+- **{{#name node_id}}**: The tree node ID this transaction belongs to
+- **{{#name tx_hex}}**: The pre-signed Spark transaction (node TX, leaf TX, or refund TX)
+- **{{#name cpfp_tx_hex}}**: The signed CPFP fee-bump transaction. `None` if this node is already confirmed on-chain — skip it, there is nothing to broadcast.
```
dangeross: nit: `None` is language specific, maybe "Not set if..."

dangeross (commented):
Great so far (50% through)

crates/breez-sdk/core/src/signer/single_key_signer.rs
```diff
@@ -0,0 +1,122 @@
...
+            let mut witness = Witness::new();
+            witness.push(signature.to_vec());
+            witness.push(pk.to_bytes());
+            psbt.inputs[i].final_script_witness = Some(witness);
+            psbt.inputs[i].partial_sigs.clear();
+        }
+
+        // Sign and finalize taproot inputs
+        if !taproot_indices.is_empty() {
+            let keypair = bitcoin::key::Keypair::from_secret_key(&secp, &self.secret_key);
```
danielgranhao: I think this P2TR branch signs with the untweaked key, which would fail script verification at broadcast. The scriptPubKey is built treating the pubkey as the internal key (committing to the tweaked output key), so the signer needs to apply the BIP341 tweak before signing — same as what `DefaultSigner::sign_hash_schnorr_with_tweak` already does for Spark's own P2TR spending paths.

```rust
use bitcoin::key::TapTweak;
let keypair = bitcoin::key::Keypair::from_secret_key(&secp, &self.secret_key)
    .tap_tweak(&secp, None)
    .to_keypair();
```

Could we add an itest that broadcasts a P2TR-funded CPFP to make sure this kind of thing surfaces?

crates/internal/src/command/withdraw.rs
```diff
@@ -319,17 +366,41 @@ fn sign_psbt(psbt: &mut Psbt, signing_key: &PrivateKey) -> Result<(), SparkWalle
...
-    for (i, pubkey, signature) in signatures.into_iter() {
+
+    // Apply ECDSA signatures
+    for (i, pubkey, signature) in ecdsa_signatures {
         psbt.inputs[i].partial_sigs.insert(pubkey, signature);
     }
+
+    // Sign taproot inputs with Schnorr key-path spend
+    if !taproot_indices.is_empty() {
+        let keypair = bitcoin::key::Keypair::from_secret_key(&secp, &signing_key.inner);
```
danielgranhao: Same thing here

danielgranhao (commented):
haven't reviewed in detail yet, just raising something detected by Claude

crates/spark/src/services/unilateral_exit.rs
JssDWt: Added a check for dust. Also changed the selection logic to just error when there isn't enough funds in the CPFP output, rather than skipping selecting leaves.

crates/breez-sdk/core/src/signer/single_key_signer.rs
JssDWt: Added that test and a whole bunch more to ensure the transactions are OK.

crates/breez-sdk/core/src/sdk/api.rs
```diff
@@ -97,6 +97,180 @@ impl BreezSdk {
...
+            .inputs
+            .into_iter()
+            .map(|input| convert_cpfp_input(input, btc_network))
+            .collect::<Result<Vec<_>, SdkError>>()?;
+
+        // Pass 1: auto-select leaves and build CPFP chains assuming nothing is
+        // confirmed.
+        let exit_result = self
+            .spark_wallet
+            .unilateral_exit_autoselect(request.fee_rate, inputs.clone(), destination.clone())
```
roeierez: I wonder if after checking for confirmed transactions (nodes) it may change the way we select profitable nodes and we should run the selection again. Not sure it is very important though.

crates/breez-sdk/core/src/sdk/api.rs
```diff
@@ -97,6 +97,180 @@ impl BreezSdk {
...
+    /// Prepares a unilateral exit by automatically selecting profitable leaves,
+    /// building all necessary transactions, and signing the CPFP fee-bump
+    /// transactions using the provided signer.
+    ///
+    /// Uses a two-pass approach: first auto-selects leaves and builds CPFP
+    /// chains assuming all ancestors are unconfirmed, then checks the chain
+    /// service for already-confirmed ancestors. If any are found, rebuilds
+    /// the CPFP chain skipping confirmed nodes.
+    #[allow(clippy::too_many_lines)]
+    pub async fn prepare_unilateral_exit(
```
roeierez: Would be nice to extract the unilateral exit code to its own file. Will be consistent with the current convention.

crates/breez-sdk/core/src/sdk/api.rs
```diff
@@ -97,6 +97,180 @@ impl BreezSdk {
...
+                    selected_ids,
+                    inputs,
+                    Some(exit_result.prefetched_nodes),
+                    &confirmed_node_ids,
+                )
+                .await?
+        };
+
+        // Build a lookup for selected leaf metadata by leaf ID.
+        let selected_by_id: std::collections::HashMap<String, &spark_wallet::SelectedLeaf> =
```
roeierez: NIT: I think we can use TreeNodeID as the hashmap key?

roeierez (commented):
Looks Good!

docs/breez-sdk/snippets/swift/BreezSdkSnippets/Sources/GettingStarted.swift
```diff
@@ -80,9 +80,6 @@ class SdkEventListener: EventListener {
         case .lightningAddressChanged(let lightningAddress):
             // The lightning address has changed
             let _ = lightningAddress
-        default:
-            // Handle any future event types
-            break
```
danielgranhao, line 85: nit: is this removal intentional? Seems unrelated to PR. In any case, if we want to remove it, probably best to remove similar "catch-all" logic from other languages like python, go, etc..

docs/breez-sdk/src/guide/unilateral_exit.md
```diff
@@ -0,0 +1,187 @@
...
+- **{{#name node_id}}**: The tree node ID this transaction belongs to
+- **{{#name tx_hex}}**: The pre-signed Spark transaction (node TX, leaf TX, or refund TX)
+- **{{#name cpfp_tx_hex}}**: The signed CPFP fee-bump transaction. Not set if this node is already confirmed on-chain — skip it, there is nothing to broadcast.
+- **{{#name csv_timelock_blocks}}**: If present, the number of blocks you must wait after the *previous* transaction confirms before broadcasting this one
+
+## Step 2: Broadcast the transaction packages
+
+You are responsible for broadcasting the transactions yourself. For each transaction in a leaf's {{#name transactions}} list:
+
+1. **If {{#name cpfp_tx_hex}} is not set**: skip this entry — the transaction is already confirmed on-chain.
```
danielgranhao: Is this up to date? Maybe I'm missing something, but it seems to me that only place where we set `cpfp_tx_hex` is `sdk/unilateral_exit.rs:152` (`cpfp_tx_hex: Some(serialize_hex(&signed_tx))`) so it is always `Some` and when transaction is already confirmed we are not including it in the transactions list

crates/breez-sdk/core/src/sdk/unilateral_exit.rs
```diff
@@ -0,0 +1,542 @@
...
+) -> (HashSet<spark_wallet::TreeNodeId>, Vec<String>) {
+    let mut confirmed_node_ids: HashSet<spark_wallet::TreeNodeId> = HashSet::new();
+    let mut known_unconfirmed: HashSet<spark_wallet::TreeNodeId> = HashSet::new();
+    let mut unverified_node_ids: Vec<String> = Vec::new();
+
+    for leaf_psbts in leaves {
+        for tc in &leaf_psbts.tx_cpfp_psbts {
+            if confirmed_node_ids.contains(&tc.node_id) {
+                continue;
+            }
```
danielgranhao: Won't this skip checking the refund tx if the node tx was found confirmed, causing the case where the node tx is confirmed and the refund tx is unconfirmed to not be correctly handled? I could be missing something of course.

danielgranhao (commented):
It's looking really good!

docs/breez-sdk/src/guide/unilateral_exit.md
JssDWt: The code is now aligned with the documentation there.

docs/breez-sdk/snippets/swift/BreezSdkSnippets/Sources/GettingStarted.swift
JssDWt: That was not intentional

roeierez (approved):

JssDWt:
Rebased

crates/breez-sdk/wasm/src/signer.rs
danielgranhao, line 994: `signHashSchnorrWithTweak` missing here

crates/spark-wallet/src/wallet.rs
```diff
@@ -1260,22 +1268,271 @@ impl SparkWallet {
...
+        // Build inputs and prevouts
+        let mut inputs = Vec::with_capacity(refund_outputs.len());
+        let mut prev_outputs = Vec::with_capacity(refund_outputs.len());
+        let mut total_value: u64 = 0;
+
+        for refund_output in &refund_outputs {
+            inputs.push(TxIn {
+                previous_output: refund_output.outpoint,
+                ..Default::default()
+            });
```
danielgranhao, line 1466: Sweep tx should also signal RBF?

danielgranhao (changes requested):
2 more small things, otherwise LGTM

JssDWt:
superseded by #992, because this PR got behind by quite a lot, and had several issues still, related to chained timelocks, RBF, discoverability of which exit path was taken, etc.

Issue #791, opened 2026-04-09
Handle unknown proto enum variants at the prost boundary


danielgranhao:
## Problem

Adding a proto enum variant is wire-compatible per the spec, so the backend ships new variants without coordination. prost's enum accessors silently collapse unknown values to variant 0 (`try_from(i32).unwrap_or_default()`), which for our enums is a plausible-but-wrong state (`TransferStatus::SenderInitiated`, `PreimageRequestStatus::WaitingForPreimage`, …). Result: silent data corruption downstream.

Supersedes #773 — the enums listed there are derived from internal state, not parsed from backend responses, so `Unknown` fallbacks on them wouldn't fix the real leak.

## Fix

At the prost boundary, replace accessor calls like `transfer.status()` with explicit `TransferStatus::try_from(transfer.status)`. On `Err`, log warn and map to an `Unknown` variant.

We then need to carefully review each use of all enums and decide whether unknown is acceptable or should result in some error.

PR #796, opened 2026-04-11
fix: handle unknown proto enum variants at prost boundary

pulkit-999:
Replace prost enum accessors (e.g. transfer.status()) with explicit try_from on the raw i32 field. Unknown values are now logged and mapped to an `Unknown(i32)` variant instead of silently defaulting to variant 0, which caused plausible-but-wrong states.

Affected enums:
- TransferStatus
- TransferType
- PreimageRequestStatus
- TokenTransactionStatus

 Partial fix for #791 ,business logic still needs review to decide whether `Unknown` should be an error or a fallback in each context.


kingonly:
We're not interested in these LLMs PRs unless you actually encountered this issue yourself and proposing a fix that helps you in your own project.

PR #797, opened 2026-04-11
Split DepositAddress into SingleUseDepositAddress and StaticDepositAddress

JssDWt:
Touches spark-wallet only.

- Return the full `SingleUseDepositAddress` struct (including keys and `leaf_id`) from `generate_deposit_address` instead of just the address string
- Since non-static deposit code paths always need a `leaf_id`, make it non-optional by splitting into `SingleUseDepositAddress` (with required `leaf_id`) and `StaticDepositAddress` (without)


roeierez:
@JssDWt can you elaborate on what was the motivation for this?

JssDWt:
@roeierez The motivation is I want to use the keys + leaf id from generate_deposit_address in another project. Also making the leaf id non-optional is useful for my usecase.

roeierez:
Thanks. I agree was just curious if some kind if an issue triggered this

roeierez (approved):
LGTM.
Nit: perhaps renaming DepsitAddress to SingleUseDepostiAddtess makes sense now?

dangeross (approved):
LGTM

PR #798, opened 2026-04-12
ship dSYMs with breez_sdk_sparkFFI.xcframework

erdemyerebasmaz:
Xcode archive validation rejected apps linking the Swift package with
"The archive did not include a dSYM for the breez_sdk_sparkFFI.framework
with the UUIDs [...]" because the XCFramework was assembled by copying
raw dylibs into a committed framework skeleton with no debug symbols
anywhere in the bundle.

- Build the iOS/Darwin bindings with CARGO_PROFILE_RELEASE_DEBUG=full so
  the dylibs carry a debug map that dsymutil can consume.
- Run dsymutil per target and stage a breez_sdk_sparkFFI.framework.dSYM
  next to each dylib; strip -S the shipped dylib afterwards so LC_UUID
  is preserved but debug info is dropped from the binary.
- In the merge jobs, lipo the per-arch DWARF Mach-Os into a single
  universal dSYM for the ios-arm64_x86_64-simulator and
  macos-arm64_x86_64 slices.
- Download the three per-slice dSYM artifacts in publish-swift and
  embed them under <slice>/dSYMs/breez_sdk_sparkFFI.framework.dSYM, and
  advertise them via DebugSymbolsPath=dSYMs in the xcframework plist.

### DIFF
--- a/.github/workflows/build-bindings-darwin.yml
+++ b/.github/workflows/build-bindings-darwin.yml
@@ -52,19 +52,41 @@ jobs:
 
       - name: Build bindings
         working-directory: crates/breez-sdk/bindings
+        env:
+          CARGO_PROFILE_RELEASE_DEBUG: "full"
         run: cargo lipo --release --targets ${{ matrix.target }}
-      
+
+      - name: Generate dSYM and strip dylib
+        run: |
+          cd target/${{ matrix.target }}/release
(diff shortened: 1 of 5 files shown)


dangeross (approved):
LGTM

roeierez (approved):
LGTM!

dangeross (approved):

PR #799, opened 2026-04-13
Cross-Chain send

### LINKED ISSUE #634: Support USDT/USDC
Allow the user to send and receive stablecoins from their BTC or stable balance.
Let's use this https://docs.flashnet.xyz/products/orchestration/overview
and then maybe add https://docs.garden.finance/ 
I propose to support the following to begin with:
* USDT/USDC Solana
* USDT Tron
* USDT/USDC Ethereum 

If we can generalize, even better.

dangeross:
Adds cross-chain send: send BTC or Spark tokens from a Spark wallet to addresses on Arbitrum, Base, Solana, Tron, and other supported chains. Two providers are wired in behind a unified API.

Closes #634 

### Breaking Change
- Changes the `payment_request` in `PrepareSendPaymentRequest` to a enum variant `PaymentRequest`.

### Providers

- **Orchestra** (Flashnet) — bridge/swap routes from Spark (BTC sats or Spark tokens) to external chains. Polls in-flight orders via Orchestra's `/status` API and reconciles the destination-chain delivered amount onto the payment row.
- **Boltz** — Lightning reverse-swap from sats to USDC/USDT on Arbitrum / Base / Solana / Tron. The SDK pays a hold invoice via the shared `LightningSender`; boltz-client drives the swap to completion over its WebSocket and emits silent metadata updates.

### API

- `BreezSdk::get_cross_chain_routes(filter)` returns the available `CrossChainRoutePair`s tagged with their provider. The same call deduplicates per-source-asset variants (e.g. one Tron/USDT route that accepts both BTC and USDB sources) into a single pair with `supported_sources: Vec<SourceAsset>`.
- `PaymentRequest::CrossChain { address, route, max_slippage_bps }` carries the user's selection into `prepare_send_payment`.
- `SendPaymentMethod::CrossChainAddress { .. }` is the prepared shape; pass it back to `send_payment` to execute. The prepared row carries `expires_at`, the provider-internal context (`CrossChainProviderContext::{Orchestra, Boltz}`), `amount_in`, `estimated_out`, and `fee_amount` — enough to resume the send without re-quoting.
- `ConversionInfo::Orchestra { .. }` / `ConversionInfo::Boltz { .. }` are persisted on the payment row's metadata so the cross-chain leg can be surfaced and tracked alongside the local Spark transfer.

### Configuration

- New `Config::cross_chain_config: Option<CrossChainConfig>` with `default_slippage_bps: Option<u32>`.
- `default_config(Mainnet)` opts in by default; `default_config(Regtest)` leaves it unset.
- `Config::validate()` rejects `cross_chain_config` on non-mainnet networks (cross-chain providers only support mainnet today).
- `validate_server_mode` rejects `cross_chain_config` when `background_tasks_enabled` is false: both providers depend on background work (Orchestra's monitor task, Boltz's WebSocket subscription) that the server runtime doesn't host.

### Validation

- **Address-family vs route-chain check** at prepare time: a hand-built `PaymentRequest::CrossChain` pairing e.g. a Solana address with a Base route fails fast with a clear message rather than relying on the provider to reject the submit. Inferred from `CrossChainRoutePair::contract_address`; skipped for native-asset routes (the provider validates).
- **Slippage range bounds**: `MIN_CROSS_CHAIN_SLIPPAGE_BPS = 10`, `MAX = 500`, default `100`. Validated at config + per-request.
- **Quote expiry validation** at send time — each provider parses its own format (Boltz: Unix epoch seconds; Orchestra: RFC3339) and surfaces `Cross-chain quote has expired. Please re-prepare.` rather than letting the provider reject the submit server-side.

### Retry safety (idempotency)

- The user's `SendPaymentRequest::idempotency_key` is now accepted for BTC-source cross-chain sends. The top-level `get_payment_by_id(idempotency_key)` lookup in `orchestrate_send` short-circuits a retry that lands after the first send already persisted a payment row.
- When the caller doesn't supply a key, each provider derives a deterministic `TransferId` from its quote/swap id (`derive_btc_leg_transfer_id` → UUIDv5 from `cross_chain:{provider}:{id}`, same shape as `stable_balance::per_receive_transfer_id`). A re-sent `prepared` shape produces the same `TransferId`, so `spark_wallet::transfer(..., transfer_id)` deduplicates at the Spark protocol layer — no second deposit fires even if the first send's persistence step never completed.
- **Token-source cross-chain sends** (e.g. USDB → external chain) carry no retry guarantee: `spark_wallet::transfer_tokens` doesn't accept an idempotency token at the protocol layer, mirroring the existing contract for direct token sends. The top-level dispatcher rejects `idempotency_key` for these with the standard "Idempotency key is not supported for token payments" message; callers retrying on transient errors can double-deposit. Documented on the `CrossChainService::send` trait and on `OrchestraClient::transfer_to_deposit`.

### Resilience

- **Orchestra post-submit timeout**: when the local payment row hasn't synced within 30 s, return a `Payment` built from the deposit `AssetTransfer` (with the Orchestra `ConversionInfo` attached) instead of erroring. The local Spark/Token transfer is settled by then; cross-chain pending state lives on `conversion_info.status`.
- **Boltz timeout**: same shape — return the pre-built pending `Payment` if the post-LN-pay storage poll times out.

### Errors

- New `From<boltz_client::BoltzError> for SdkError` classifies into `NetworkError` / `StorageError` / `InvalidInput` / `Generic`. Drops the local `boltz_err_to_sdk` wrapper; cross-chain provider call sites consolidate to `?`.
- `CrossChainProviders::get` returns `InvalidInput` on missing provider (was `Generic`); error message uses the provider's `Display` impl rather than `Debug`.
- User-facing error messages avoid Rust field names and debug-formatted enums (e.g. `cross_chain_config is only supported on mainnet (got Regtest)` → `Cross-chain sends are only available on Mainnet, not on Regtest.`).

### Internals

- `cross_chain` module with `CrossChainService` trait, `OrchestraService`, `BoltzService`, `BoltzStorageAdapter`, `BoltzSdkEventListener`.
- `AssetTransfer` hoisted from `flashnet::amm::models` to top-level `flashnet::models` — shared between AMM and Orchestra. `flashnet::OrchestraClient::transfer_to_deposit` returns the full `AssetTransfer` instead of stringifying to a hash, matching the AMM `transfer_asset` shape.
- `fee_overpayment` kernel moved to `utils::fees` so the Bolt11, LNURL-pay, and Boltz cross-chain send paths share the same overpayment reconciliation rather than duplicating it inline.
- New `payment_from_asset_transfer` in `utils::conversions` — shared between the AMM swap path and the Orchestra send fallback.
- boltz-client bumped to rev `7367b536` for USDC destinations + a quieter WS message parse log.

### Tests

- ~60 new unit tests across `cross_chain/{orchestra,boltz,boltz_event_listener,boltz_storage_adapter,mod}.rs` and `sdk/payments/prepare/cross_chain.rs`, including the `derive_btc_leg_transfer_id` helper (caller-key precedence, deterministic derivation, namespace separation between providers).
- Storage backend parity: shared test suite adds Boltz `ConversionInfo` insert/update fixtures, run against sqlite, postgres, mysql, web, node-sqlite, node-postgres, node-mysql.
- `ConversionInfo` type-discriminator migration regression tests on sqlite + postgres + mysql, plus a realtime-sync `test_incoming_payment_metadata_upgrades_pre_migration_conversion_info` covering the lenient deserializer's upgrade path for incoming sync records.
- `#[macros::test_all]` for native + wasm pure-logic tests; `#[cfg(feature = "sqlite")]` + plain `#[tokio::test]` for storage-touching tests.
- New itests: `lightning_send_server_mode`, `conversion_send_server_mode`.
- All 404 SDK unit tests + 82 integration tests pass.

### Storage migrations

Per-backend type-discriminator backfill migration on the `conversion_info` column so existing payment rows survive the move from `ConversionInfo` (struct) to `ConversionInfo` (tagged enum: `Amm` / `Orchestra` / `Boltz`). Applied to sqlite, postgres, mysql, web (IndexedDB), node-sqlite, node-postgres, node-mysql. The migration is what protects direct DB reads — the strict tagged-enum deserializer on `from_json_string_opt` would otherwise fail on untagged rows. Incoming pre-migration sync records are upgraded separately by `deserialize_conversion_info_with_default_type` and re-serialized on insert.


crates/breez-sdk/common/src/input/cross_chain.rs
```diff
@@ -0,0 +1,570 @@
...
+
+    None
+}
+
+// ---------------------------------------------------------------------------
+// URI parsing
+// ---------------------------------------------------------------------------
+
+/// Parsed representation of a canonical cross-chain URI of the form
+/// `<scheme>:<address>?chain=<chain>&asset=<asset>&amount=<u128>&label=<text>`.
```
danielgranhao: Is this a common standard scheme?

If it is just an SDK-level scheme, could you elaborate what's the value it provides over using types?

crates/breez-sdk/core/src/sdk/api.rs
```diff
@@ -70,7 +70,57 @@ impl BreezSdk {
...
+                Err(err)
+            }
+        }
+    }
+
+    /// Returns the available `{chain, asset}` routes for a given address family.
+    ///
+    /// The UI calls this after `parse()` returns `CrossChainAddress` to populate
+    /// the chain/asset picker. Optionally filter by a specific asset or provider.
+    pub async fn get_cross_chain_routes(
```
danielgranhao: If consumers should always call this after parse, why not do it automatically and include routes in parse result?

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -1300,9 +1332,29 @@ pub enum OnchainConfirmationSpeed {
...
 }
 
+/// The payment destination. Either a raw string (bolt11, spark address, BIP-21,
+/// cross-chain URI, etc.) that is parsed internally, or a structured
+/// cross-chain destination with explicit chain + asset selection.
+#[derive(Debug, Clone, Serialize)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
+pub enum PaymentRequest {
+    /// Existing string-based input (bolt11, spark address, BIP-21, cross-chain URI, etc.)
+    Raw(String),
```
danielgranhao: This is a detail that we can iron out later, but I don't love `Raw` because it suggests “unparsed blob” while IMO the real destinction is:
- string-only request: the string fully identifies the payment target
- string + extra disambiguation: the address alone is insufficient, so chain/asset/provider are needed

What about `Direct`?

crates/breez-sdk/core/src/sdk/api.rs
```diff
@@ -70,7 +70,57 @@ impl BreezSdk {
...
+        &self,
+        family: crate::CrossChainAddressFamily,
+        asset: Option<String>,
+        provider: Option<crate::cross_chain::CrossChainProvider>,
+    ) -> Result<Vec<crate::CrossChainRoutePair>, SdkError> {
```
danielgranhao: Maybe CrossChainRoutePair should include which provider fulfils it? Otherwise later when paying how do consumers know which providers they can choose?

crates/breez-sdk/core/src/cross_chain/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+        family: CrossChainAddressFamily,
+        asset: Option<&str>,
+    ) -> Result<Vec<CrossChainRoutePair>, SdkError>;
+
+    /// Fetch a quote for a cross-chain send.
+    async fn prepare(
+        &self,
+        recipient_address: &str,
+        dest_chain: &str,
+        dest_asset: &str,
```
danielgranhao: could be replaced by `route: &CrossChainRoutePair`?

crates/breez-sdk/core/src/cross_chain/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+        self.0.iter()
+    }
+}
+
+/// Data stashed on the prepared send payment so the provider can resume
+/// the send stage without re-quoting.
+#[derive(Debug, Clone)]
+pub(crate) struct CrossChainPrepared {
+    pub quote_id: String,
+    pub deposit_address: String,
```
danielgranhao: This is specific to Orchestra. For boltz provider we must pay a BOLT11 invoice. Maybe an enum? We can also figure it out later when integrating Boltz

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -1300,9 +1332,29 @@ pub enum OnchainConfirmationSpeed {
...
+    /// Cross-chain send with explicit chain/asset selection.
+    /// Amount comes from `PrepareSendPaymentRequest.amount`, not here.
+    CrossChain {
+        address: String,
+        chain: String,
+        asset: String,
+        /// Optional cross-chain provider selection. When set, only the specified
+        /// provider is used. When `None`, the SDK selects the best available provider.
+        provider: Option<crate::cross_chain::CrossChainProvider>,
```
danielgranhao: I think it would be good to add the provider to `CrossChainRoutePair` as per the other comment, and then replace these with `CrossChainRoutePair`. Consumers would call prepare with a PaymentRequest::CrossChain directly with one of the `CrossChainRoutePair` obtained from `get_cross_chain_routes`.

crates/breez-sdk/core/src/sdk/api.rs
roeierez: @danielgranhao I think it is a good separation of concerns not to involve routes as part of the parse actually. parse can also be used regardless

crates/breez-sdk/common/src/input/cross_chain.rs
dangeross: It's kind of an artifact of an old design. But loosely based on [EIP-681](https://eips.ethereum.org/EIPS/eip-681), which is based on BIP-21. EIP-681 doesn't have the chain/asset info, but it can contain the contract address for ERC-20 tokens. 

The [EIP-681](https://eips.ethereum.org/EIPS/eip-681) URI format parsing might be the way to go. If you compare it to the routes JSON from Orchestra, we could match the contract_address/chain_id to the chain/asset. 

Solana and Tron also use similar schemas:
- Solana: solana:mvines9iiHiQTysrwkJjGf2gb9Ex9jXJX8ns3qwf2kN?amount=0.01&spl-token=EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v
- Tron: tron:TXYZ...abc?amount=100&token=TRC20_CONTRACT_ADDRESS

crates/flashnet/src/config.rs
```diff
@@ -14,18 +15,56 @@ pub struct IntegratorConfig {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("OrchestraConfig")
+            .field("base_url", &self.base_url)
+            .field("api_key", &"<redacted>")
+            .finish()
+    }
+}
+
+const ORCHESTRA_API_KEY: &str = "fnp_7NLgx9aK-57OAhnRbpAm-VUbpYxReW8f5eDWddRDpm8";
+const ORCHESTRA_BASE_URL: &str = "https://orchestration.flashnet.xyz";
```
roeierez: Since we are using our own keys I think they should be dynamic (returned from BreezServer). Otherwise now way to rotate keys, mitigate abuse, etc...
We will also be able to create keys per partner this way.

crates/breez-sdk/core/src/sdk/mod.rs
```diff
@@ -201,6 +210,7 @@ pub fn default_config(network: Network) -> Config {
         stable_balance_config: None,
         max_concurrent_claims: 4,
         spark_config: None,
+        cross_chain_enabled: matches!(network, Network::Mainnet),
```
roeierez: Not sure I would enable it by default. It has some overhead (web sockets for example) which I think better to let users opt in for this.

crates/breez-sdk/core/src/cross_chain/orchestra.rs
```diff
@@ -0,0 +1,516 @@
...
+/// The canonical Spark source chain string used by Orchestra.
+const SPARK_SOURCE_CHAIN: &str = "spark";
+
+fn parse_amount(value: &str, field: &str) -> Result<u128, SdkError> {
+    value
+        .parse::<u128>()
+        .map_err(|e| SdkError::Generic(format!("Orchestra returned invalid {field}: {e}")))
+}
+
+const DEFAULT_SLIPPAGE_BPS: u32 = 50;
```
roeierez: Should we allow users configure this in the sdk config?

crates/breez-sdk/core/src/cross_chain/boltz_storage_adapter.rs
```diff
@@ -0,0 +1,352 @@
...
+    async fn increment_key_index(&self) -> Result<u32, BoltzError> {
+        let _guard = self.key_index_mutex.lock().await;
+        let key = key_index_cache_key(&self.instance_id);
+        let current = self
+            .storage
+            .get_cached_item(key.clone())
+            .await
+            .map_err(|e| BoltzError::Store(format!("Failed to read key index: {e}")))?
+            .and_then(|v| v.parse::<u32>().ok())
+            .unwrap_or(0);
```
roeierez: What is increment_key_index used for? Asking to understand if there are consequences of zeroing it in some corrupted data scenario or a bug.

crates/breez-sdk/core/src/cross_chain/boltz_storage_adapter.rs
```diff
@@ -0,0 +1,352 @@
...
+                self.save_active_ids(&ids).await?;
+            }
+        }
+        Ok(())
+    }
+
+    async fn update_swap(&self, swap: &BoltzSwap) -> Result<(), BoltzError> {
+        let serialized = serde_json::to_string(swap)
+            .map_err(|e| BoltzError::Store(format!("Failed to serialize swap: {e}")))?;
+        self.storage
```
roeierez: MemoryBoltzStorage (https://github.com/breez/boltz-client/blob/main/crates/lib/src/store.rs#L52) skips the update in case it doesn't exist, here we create it?

crates/breez-sdk/core/src/cross_chain/boltz_storage_adapter.rs
```diff
@@ -0,0 +1,352 @@
...
+    instance_id: String,
+    key_index_mutex: Arc<Mutex<()>>,
+    /// Serializes read-modify-write of `boltz_active_swap_ids` across
+    /// concurrent `insert_swap`/`update_swap` calls. Without it, two
+    /// concurrent swap transitions can race on the load → mutate → save
+    /// cycle and lose one another's update.
+    active_ids_mutex: Arc<Mutex<()>>,
+}
+
+impl BoltzStorageAdapter {
```
roeierez, line 57: Did you consider implementing a swaps table rather than use the key value cache? somehow it seems to me it will be clearer and easier to manage (also atomically).

crates/breez-sdk/core/src/cross_chain/orchestra.rs
```diff
@@ -0,0 +1,1428 @@
...
+        let (monitor_trigger, _) = broadcast::channel(10);
+
+        let service = Self {
+            client,
+            spark_wallet,
+            storage,
+            monitor_trigger: monitor_trigger.clone(),
+        };
+        info!("Orchestra service initialized");
+        service.spawn_monitor(shutdown_receiver, &monitor_trigger);
```
roeierez, line 127: It seems that this should be running on client mode only? On server mode users would probably use the webhooks feature?
I don't think we have an immediate demand/use case to run the cross chain on server mode (can be deferred for now) but good to keep that in mind and stay aligned with our guidelines.

crates/breez-sdk/core/src/cross_chain/orchestra.rs
```diff
@@ -0,0 +1,1428 @@
...
+        // Step 1: Spark transfer to the Orchestra deposit address.
+        let asset_transfer = self
+            .client
+            .transfer_to_deposit(
+                deposit_address,
+                prepared.amount_in,
+                prepared.token_identifier.as_deref(),
+                transfer_id,
+            )
+            .await?;
```
roeierez, line 627: What is the mechanism to recover if we crash after this send?

roeierez (commented):
Looks great! Things look in place, I went over the common logic and orchestra specifics, still need to go over boltz provider specifics.

crates/breez-sdk/core/src/sdk/lightning_sender.rs
```diff
@@ -0,0 +1,176 @@
...
+
+use super::{SyncCoordinator, SyncType};
+
+/// Reusable helper that owns the dependencies needed to pay a BOLT11
+/// invoice, persist the resulting [`Payment`] row, and reconcile its status
+/// with the SSP via background polling.
+///
+/// Held behind `Arc` and shared between `BreezSdk` and any cross-chain
+/// provider that pays LN invoices (currently: Boltz reverse swap).
+pub(crate) struct LightningSender {
```
roeierez, line 30: Do you think we can use this LightningSender also for the bolt11 send in the sdk API? Seems like we duplicated the logic from there?

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -1255,3 +1360,99 @@ mod tests {
...
+        let handle: BoltzInstanceHandle = serde_json::from_str(&raw)
+            .map_err(|e| SdkError::Generic(format!("Corrupted Boltz instance handle: {e}")))?;
+        return Ok(handle);
+    }
+
+    let mut seed = [0u8; 32];
+    thread_rng().fill_bytes(&mut seed);
+    let handle = BoltzInstanceHandle {
+        instance_id: uuid::Uuid::new_v4().to_string(),
+        seed_hex: hex::encode(seed),
```
roeierez: Encrypt it?

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -1255,3 +1360,99 @@ mod tests {
...
+struct BoltzInstanceHandle {
+    instance_id: String,
+    seed_hex: String,
+}
+
+/// Initializes the Boltz reverse-swap cross-chain provider: loads or creates
+/// the local instance seed, constructs the inner `BoltzClient`, registers the
+/// event listener, resumes any active swaps, and returns an SDK-side wrapper
+/// ready to be inserted into the provider registry.
+async fn build_boltz_service(
```
roeierez: Perhaps this method belongs to the boltz provider?

crates/breez-sdk/core/src/cross_chain/boltz_event_listener.rs
```diff
@@ -0,0 +1,337 @@
...
+            .insert_payment_metadata(payment_id, updated)
+            .await
+            .map_err(|e| format!("persist degraded-flag update: {e}"))?;
+        Ok(())
+    }
+}
+
+#[macros::async_trait]
+impl BoltzEventListener for BoltzSdkEventListener {
+    async fn on_event(&self, event: BoltzSwapEvent) {
```
roeierez, line 176: Question is how reliable this mechanism of relying solely on events (web sockets?) to progress the swap status.
Do we have another mechanism to progress on both positive (success swap ) and negative (refund) cases?

crates/breez-sdk/core/src/cross_chain/orchestra.rs
dangeross: Its only available in client mode:
- default_config(): https://github.com/breez/spark-sdk/pull/799/changes#diff-0460beaf5752085f2fef4ddc3c6e99b97357457af23242f144955e3d9e84e117R222-R223
- default_server_config(): https://github.com/breez/spark-sdk/pull/799/changes#diff-0460beaf5752085f2fef4ddc3c6e99b97357457af23242f144955e3d9e84e117R267
- validate_server_mode(): https://github.com/breez/spark-sdk/pull/799/changes#diff-65c90d6c61087efceb3a1c997ba02b5986bdbefc3fa3b74159b3937cfabbe083R557-R561

crates/breez-sdk/core/src/cross_chain/orchestra.rs
dangeross: Flashnet auto-detects deposits to the address and will fulfill/refund. We could some type of cache using the idempotency_key to store the quote id ahead of the transfer. Wdyt?

roeierez: I am not sure it is necessary. Do we identify these flashnet refunds as failed payments/refunds?
Just want to know how the payments list would look like for a failed/refunded payment.

dangeross: In general after calling `submit_spark()` the payment is monitored until terminal then marked `Completed`/`Refunded`. If the `submit_spark()` request failed its marked `RefundNeeded` and monitored

But back to the original question, at this point if there was a crash, there would be no conversion metadata associated to the payment

roeierez: Thanks Ross. I think it is OK as is currently.

crates/breez-sdk/core/src/cross_chain/boltz_storage_adapter.rs
danielgranhao: The key index is used to derive a swap's preimage, so it's important that this is durable. Zeroing results in preimage reuse, which at best results in errors when trying to create a swap (honest boltz backend) or at worst results in fund loss (malicious or boltz). 

I see that here we proceed with 0 if we fail to parse. That should be a loud error, will fix.

crates/breez-sdk/core/src/cross_chain/boltz_storage_adapter.rs
danielgranhao: The MemoryBoltzStorage is mainly a test helper and not necessarily supposed to be a reference. In practice, both behaviors are fine. But the question actually reveals that having separate insert and update methods is unnecessary. I'll update the trait to have a single upsert method.

crates/breez-sdk/core/src/cross_chain/orchestra.rs
danielgranhao: it is configurable now

crates/breez-sdk/core/src/sdk/mod.rs
danielgranhao: Disabled it

crates/breez-sdk/core/src/cross_chain/boltz_storage_adapter.rs
danielgranhao: Done

crates/breez-sdk/core/src/cross_chain/boltz_storage_adapter.rs
danielgranhao: Done

crates/breez-sdk/core/src/cross_chain/orchestra.rs
danielgranhao: Added a test to ensure server mode rejects cross chain

crates/breez-sdk/core/src/sdk/lightning_sender.rs
danielgranhao: LightningSender was created exactly for that purpose! Must have been an uncaught rebase issue that reverted it. Nice catch. Fixed

crates/breez-sdk/core/src/cross_chain/boltz_event_listener.rs
danielgranhao: It isn't solely events. This listener only mirrors state into payment metadata. The swaps themselves are progressed by the boltz client, which relies on web sockets but also has a background poll loop on top that covers terminal cases: delivery confirmation (Settling → Completed) and timeout-gated refund detection (Claiming → Failed). On restart, resume_swaps() reloads active swaps and resubscribes.

crates/breez-sdk/core/src/sdk_builder.rs
danielgranhao: Done

crates/breez-sdk/core/src/sdk_builder.rs
danielgranhao: Done

crates/flashnet/src/config.rs
danielgranhao: Done

crates/breez-sdk/core/src/cross_chain/boltz_storage_adapter.rs
danielgranhao: I went for the KV cache to keep it simple and avoid changing all storages. But I agree it would be nice to avoid the mutex. I'll migrate to a swaps table in a later PR

crates/breez-sdk/core/src/cross_chain/boltz_event_listener.rs
danielgranhao: Actually, you were right, we can get into sync issues if, for any reason, we fail to update the storage when handling a terminal event here. I saw this in practice due to a race bug that I fixed in [5c8b23d](https://github.com/breez/spark-sdk/pull/799/commits/5c8b23db69bcf145fdb6ff89e44cd75215859fbc). But even after the fix, it can always happen that we may fail to update swap, so the same commit also adds a reconciliation step at startup that checks all payments with pending boltz swaps to see if they may have terminated.

crates/breez-sdk/core/src/cross_chain/boltz.rs
```diff
@@ -0,0 +1,1107 @@
...
+const SEND_POLL_TIMEOUT_SECS: u64 = 60;
+
+/// Hardened derivation index reserved for encrypting the Boltz instance handle
+/// at rest. `1112430164` == ASCII "BOLT", distinct from the session store's
+/// "SESN" path, `RTSyncSigner`'s indices, and the `KeySet` master keys, so this
+/// scope can never collide with another subsystem deriving from the same
+/// identity master key. No per-network variant is needed: the signer's
+/// `identity_master_key` is already derived under a network-specific account
+/// number, so mainnet and regtest yield distinct encryption keys regardless.
+const BOLTZ_INSTANCE_ENCRYPTION_PATH: &str = "m/1112430164'/0'/0'/0/0";
```
dangeross: `1112430164` is `0x424E5654` = "BNVT", not "BOLT" (that would be `1112493140`). Harmless as a derivation index, but this value now derives the at-rest encryption key for the Boltz seed, so if someone later "corrects" the literal to the real "BOLT" value to match this comment, every existing encrypted handle stops decrypting (we discard and regenerate, dropping any swap in flight on that device).

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -861,6 +892,116 @@ async fn build_stable_balance(
...
+/// Total attempts to fetch the orchestra config from Breez server before
+/// giving up for this session.
+const ORCHESTRA_FETCH_MAX_ATTEMPTS: u32 = 3;
+/// First backoff between orchestra config fetch retries; doubles each attempt.
+const ORCHESTRA_FETCH_BASE_BACKOFF: Duration = Duration::from_millis(256);
+
+/// Fetches the orchestra config from Breez server with a bounded retry.
+///
+/// Only connectivity failures are retried (with a short doubling backoff): a
+/// definitive empty response from the server is a final answer and is not
```
dangeross: `fetch_orchestra_config` collapses every gRPC error (including a definitive 4xx like auth) into `ServiceConnectivityError::Other`, so the `Err` arm below cannot tell a transient failure from a permanent one and retries all of them 3x with backoff. Only the `Ok(None)` empty-config case short-circuits. Minor given it is bounded and an optional provider, but "only connectivity failures are retried" promises a distinction the code does not make. Narrow the wording, or actually classify the error?

crates/breez-sdk/core/src/cross_chain/boltz.rs
danielgranhao: Nice that you caught this. I remember thinking I should verify the Claude-generated ASCII as it's prone to make mistakes and then I forgot. FIxed

crates/breez-sdk/core/src/sdk_builder.rs
danielgranhao: Narrowed the wording

crates/breez-sdk/core/src/cross_chain/boltz.rs
dangeross: It's funny that Claude found the error though 😅

crates/breez-sdk/core/src/cross_chain/orchestra.rs
```diff
@@ -0,0 +1,1741 @@
...
+            metadata,
+            &self.spark_wallet,
+            &self.storage,
+            true,
+        )
+        .await
+        .unwrap_or_else(|e| {
+            error!("Failed to persist Orchestra metadata for payment {spark_tx_hash}: {e:?}");
+            spark_tx_hash
+        });
```
danielgranhao: If this fails, can't we end up dropping the `ConversionInfo` without any reconciliation path? If so, in the best case we lose metadata, in the worst, we lose information about a possible `RefundNeeded`. Maybe we need a follow-up similar to #927, but for orchestra conversion.

danielgranhao (approved):

crates/breez-sdk/core/src/cross_chain/orchestra.rs
dangeross: I don't know of a similar catch-all fallback path for orchestra, they seem to always handle the refunding automatically

PR #800, opened 2026-04-14
lnurl server: auto-refresh allowed domains

JssDWt:
Automatically refresh allowed domains every minute. So the server doesn't have to be restarted when a new domain is added.

### DIFF
--- a/crates/breez-sdk/lnurl/src/domains.rs
+++ b/crates/breez-sdk/lnurl/src/domains.rs
@@ -1,0 +1,65 @@
+use std::{collections::HashSet, sync::Arc, time::Duration};
+
+use tokio::sync::RwLock;
+use tracing::{debug, error, info};
+
+use crate::repository::{LnurlRepository, LnurlRepositoryError};
+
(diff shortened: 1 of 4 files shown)


crates/breez-sdk/lnurl/src/domains.rs
```diff
@@ -0,0 +1,67 @@
...
+    let mut sorted: Vec<&str> = initial.iter().map(String::as_str).collect();
+    sorted.sort_unstable();
+    info!("loaded allowed domains: {}", sorted.join(", "));
+    let domains = Arc::new(RwLock::new(initial));
+
+    let db_clone = db.clone();
+    let domains_clone = Arc::clone(&domains);
+    tokio::spawn(async move {
+        debug!("Allowed domains refresher started");
+        loop {
```
dangeross, line 27: Do we need to handle shutdown here?

JssDWt: Not really, it'll get dropped when the server stops because of the tokio runtime shutting down.

roeierez (approved):

dangeross (approved):

PR #805, opened 2026-04-16
LNURL server: sign webhooks

JssDWt:
- Sign outgoing webhooks (X-Breez-Signature header)
- Route deliveries by domain instead of URL — there's only one webhook endpoint per domain, so storing the URL as routing key doesn't make sense. The URL is now resolved from domain_webhooks at send time. We still persist it on each attempt so we can see where it was sent for debugging
- domain_webhooks is now fully owned by the webhook domain. The LNURL side just enqueues a delivery for a domain and doesn't care about URLs or secrets
- Webhook configs are cached in memory and refreshed every 60s

### DIFF
--- a/crates/breez-sdk/lnurl/migrations/postgres/20260416120000_webhook_signing.sql
+++ b/crates/breez-sdk/lnurl/migrations/postgres/20260416120000_webhook_signing.sql
@@ -1,0 +1,18 @@
+-- Per-domain signing secret for outgoing webhooks.
+ALTER TABLE domain_webhooks ADD COLUMN webhook_secret TEXT NOT NULL;
+
+-- Route webhook deliveries by domain instead of URL.
+-- The URL is resolved at send time from domain_webhooks and stored for audit.
+ALTER TABLE webhook_deliveries ADD COLUMN domain TEXT NOT NULL DEFAULT 'unknown';
+ALTER TABLE webhook_deliveries ALTER COLUMN url DROP NOT NULL;
+
+-- Replace old (identifier, url) uniqueness with domain-based constraint.
+ALTER TABLE webhook_deliveries DROP CONSTRAINT webhook_deliveries_identifier_url_key;
+CREATE UNIQUE INDEX idx_webhook_deliveries_identifier_domain
+    ON webhook_deliveries (identifier, domain);
+
+-- Update the pending index to partition by domain instead of url.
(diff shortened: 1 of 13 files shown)


crates/breez-sdk/lnurl/migrations/postgres/20260416120000_webhook_signing.sql
```diff
@@ -0,0 +1,17 @@
+-- Per-domain signing secret for outgoing webhooks.
+ALTER TABLE domain_webhooks ADD COLUMN webhook_secret TEXT NOT NULL;
```
dangeross, line 2: Do we need to set defaults? You mentioned webhooks exist already

crates/breez-sdk/lnurl/migrations/sqlite/20260416120000_webhook_signing.sql
```diff
@@ -0,0 +1,41 @@
...
+    url TEXT,
+    payload TEXT NOT NULL,
+    created_at BIGINT NOT NULL,
+    succeeded_at BIGINT,
+    retry_count INTEGER NOT NULL DEFAULT 0,
+    next_retry_at BIGINT NOT NULL,
+    claimed_at BIGINT,
+    last_error_status_code INTEGER,
+    last_error_body TEXT,
+    UNIQUE (identifier, url),
```
dangeross: Do we need this constraint now we have the `(identifier, domain)` one, especially now that url can be NULL

crates/breez-sdk/lnurl/migrations/postgres/20260416120000_webhook_signing.sql
JssDWt: There are currently no rows in this table. Just 2 rows in the webhook_deliveries table.

crates/breez-sdk/lnurl/migrations/sqlite/20260416120000_webhook_signing.sql
JssDWt: Very good catch. That constraint was acutally still there on postgres too. I simplified the sqlite migration too, because there are no existing databases there anyway.

dangeross (approved):
LGTM

danielgranhao (approved):
LGTM

PR #808, opened 2026-04-17
Core: Remove JWT from SSP requests

hydra-yse:
Following the offline dicussions, JWT usage on the SSP is not planned, so we remove it in order not to affect WASM builds (some calls fail with CORS errors)

### DIFF
--- a/crates/breez-sdk/core/src/sdk/sync.rs
+++ b/crates/breez-sdk/core/src/sdk/sync.rs
@@ -692,6 +692,10 @@ mod jwt {
     }
 
     impl BreezSdk {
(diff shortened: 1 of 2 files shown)


JssDWt (approved):
Looks fine. If ever we need different headers to go to the SSP, we should split it up into multiple session managers probably.

roeierez (approved):
LGTM

roeierez:
> Looks fine. If ever we need different headers to go to the SSP, we should split it up into multiple session managers probably.

Yeh, or use different fields in side the session (so_headers, ssp_headers).

dangeross (approved):

PR #811, opened 2026-04-20
fix: balance doesn't update immediately on receive (multi-instance)

danielgranhao:
This fixes 3 issues that could cause balance to lag increasing after receiving a payment when multiple instances are running concurrently:
- Claim race caused the losing instance to only insert leaves into store on the next background full sync, causing a lag of up to sync period  -> fixed by querying for the transfer on TransferAlreadyClaimed error
- With the fix above, instances reliably race for optimizing leaves, causing balance to wobble due to the losing instance refreshing leaves immediately -> fixed by removing the refresh on optimization failure. There's no need to refresh leaves immediately. 
- On browser instances sharing an indexed db storage, the sync skip mechanism could cause an instance to never emit `Synced` even, which can cause the UI to never get updated -> fixed by emitting `Synced` when syncing is skipped

### DIFF
--- a/crates/breez-sdk/core/src/sdk/sync.rs
+++ b/crates/breez-sdk/core/src/sdk/sync.rs
@@ -284,11 +284,9 @@ impl BreezSdk {
             && now.saturating_sub(last) < sync_interval_secs
         {
             debug!("sync_wallet_internal: Synced recently, skipping");
-            // Another instance sharing our storage synced recently, so shared
-            // caches are fresh — tell consumers so their UIs refresh. Without
-            // this, an instance that keeps skipping (e.g. a browser tab when
-            // another tab is the periodic-sync winner) would never emit a
-            // `Synced` event.
+            // When another instance shares our storage and keeps winning the sync
+            // race, we would otherwise never emit a Synced event. Emit it here so
+            // consumers are still notified that storage is up to date.
             self.event_emitter.emit(&SdkEvent::Synced).await;
             return Ok(());
         }
--- a/crates/breez-sdk/core/src/sdk/sync.rs
+++ b/crates/breez-sdk/core/src/sdk/sync.rs
@@ -284,6 +284,12 @@ impl BreezSdk {
             && now.saturating_sub(last) < sync_interval_secs
         {
             debug!("sync_wallet_internal: Synced recently, skipping");
+            // Another instance sharing our storage synced recently, so shared
+            // caches are fresh — tell consumers so their UIs refresh. Without
+            // this, an instance that keeps skipping (e.g. a browser tab when
+            // another tab is the periodic-sync winner) would never emit a
+            // `Synced` event.
+            self.event_emitter.emit(&SdkEvent::Synced).await;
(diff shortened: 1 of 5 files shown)


crates/spark/src/services/transfer.rs
```diff
@@ -727,8 +739,20 @@ impl TransferService {
...
-                    return Err(ServiceError::TransferAlreadyClaimed);
+                    // Another instance of this wallet claimed the transfer concurrently.
+                    // Re-query the transfer from the coordinator; if it's now completed,
+                    // return its finalized leaves so the caller can update its local tree
+                    // instead of surfacing the error.
+                    debug!(
+                        "Transfer {} already claimed by another instance; fetching finalized transfer",
+                        transfer.id
+                    );
+                    match self.query_transfer(&transfer.id).await? {
```
roeierez: It seems that the `claim_transfer_sign_refunds` which is called inside `claim_transfer_with_leaves` already queries for the transfer before it returns `TransferAlreadyClaimed` error. Perhaps we can pass it as part of the error instead of querying again?

roeierez (changes requested):

crates/spark/src/services/transfer.rs
danielgranhao: Good catch! Fixed. Went for a slight refactor to get rid of `TransferAlreadyClaimed` error and treat this as a successful outcome.

roeierez (approved):

danielgranhao:
@roeierez Re-triggered review request because I added a new commit that also handles "already claimed" case for `finalize_node_signatures`. I noticed we were failing there on first claim attempt.

crates/spark/src/services/transfer.rs
roeierez: @danielgranhao  Now that we have two points to check I wonder if we should only propagate the error and the natural place to query the transfer is in the outer layer (caller).

danielgranhao: Makes total sense. Done

roeierez (approved):
Looks Good!

crates/breez-sdk/core/src/sdk/sync.rs
```diff
@@ -284,6 +284,12 @@ impl BreezSdk {
             && now.saturating_sub(last) < sync_interval_secs
         {
             debug!("sync_wallet_internal: Synced recently, skipping");
+            // Another instance sharing our storage synced recently, so shared
+            // caches are fresh — tell consumers so their UIs refresh. Without
+            // this, an instance that keeps skipping (e.g. a browser tab when
+            // another tab is the periodic-sync winner) would never emit a
+            // `Synced` event.
```
JssDWt: nit: I find this comment a bit confusing
```suggestion
            // When another instance shares our storage and keeps winning the sync
            // race, we would otherwise never emit a Synced event. Emit it here so
            // consumers are still notified that storage is up to date.
```

JssDWt (approved):
Looking good. Just a small nit.

PR #816, opened 2026-04-22
Cross-chain: Multi-step conversion details

dangeross:
This PR:
- Replaces the flat `from`/`to` fields on `ConversionDetails` with an ordered `conversions: Vec<Conversion>` array, where each element represents one conversion step (AMM, Orchestra, or Boltz) with its own provider, status, from/to sides, and fees
- Introduces `ConversionProvider`, `ConversionSide`, and `Conversion` types, each side carries `chain`, `asset`, `amount`, `fee`, and `decimals`

**Enrichment logic**
- `get_payment_with_conversion_details` and `list_payments` now build conversions from both child payments (AMM) and the parent's own `ConversionInfo` (cross-chain)
- Conversions are ordered directionally: sends = `[AMM, cross-chain]`, receives = `[cross-chain, AMM]`
- Cross-chain payments without pre-set `conversion_details` (e.g. Orchestra token sends) now get `ConversionDetails` populated at retrieval time
- Handles auto-conversion and self-transfer cases where only one child payment is linked

**ConversionInfo cleanup**
- Renamed `destination_*` fields to direction-neutral names (`chain`, `asset`, `recipient_address`, `asset_decimals`)
- Added `delivered_amount` to Orchestra variant, monitoring now freezes `estimated_out` at prepare time and sets `delivered_amount` from the status response, matching Boltz behavior
- Orchestra monitor also updates `fee` from the status response's settled `fee_amount`
- Added `asset_decimals` to both cross-chain variants, populated from the route at prepare time
- Added lenient `PaymentMetadata` deserializer that backfills `"type": "amm"` for pre-migration sync records missing the tag


crates/breez-sdk/core/src/utils/payments.rs
```diff
@@ -50,37 +54,122 @@ pub(crate) async fn get_payment_and_emit_event(
...
+        } else {
+            // Cross-chain send without pre-set conversion_details — derive status from info
+            let status = extract_conversion_info(payment.details.clone())
+                .map_or(ConversionStatus::Completed, |info| info.status().clone());
+            payment.conversion_details = Some(crate::models::ConversionDetails {
+                status,
+                conversions,
+            });
+        }
+    }
```
danielgranhao: Looks like it duplicates logic from lines 496–525 of sdk/payments.rs. Can we extract it?

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -255,104 +255,221 @@ impl Payment {
...
+                amount: source_payment.amount,
+                fee: 0,
+                decimals: from_decimals,
+            },
+            to: ConversionSide {
+                chain: chain.clone(),
+                asset: asset.clone(),
+                amount: delivered_amount.unwrap_or(*estimated_out),
+                fee: fee.unwrap_or(0),
+                decimals: Some(asset_decimals.unwrap_or(DEFAULT_CROSS_CHAIN_DECIMALS)),
```
danielgranhao: Do we need this "migration" logic? Seems unnecessary given it's an unreleased feature

crates/breez-sdk/core/src/utils/payments.rs
```diff
@@ -50,37 +54,122 @@ pub(crate) async fn get_payment_and_emit_event(
...
+        map.get(&payment.id).cloned()
+    } else {
+        None
+    };
+
+    let conversions = build_conversions(payment, child_payments.as_deref());
+
+    if !conversions.is_empty() {
+        if let Some(ref mut cd) = payment.conversion_details {
+            cd.conversions = conversions;
```
danielgranhao: Here we leave the status unchanged. For a payment with AMM + cross chain, status may already be completed despite cross chain part being pending, so the top level status lies. 

I think we need to recompute status here from the worst per-step status.

danielgranhao: I think this also affects the duplicated logic in sdk/payments.rs

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -255,104 +255,221 @@ impl Payment {
...
 #[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
-pub struct ConversionStep {
-    /// The underlying payment id of the conversion step
-    pub payment_id: String,
-    /// Payment amount in satoshis or token base units
+pub struct ConversionSide {
+    /// The chain or network (e.g. "spark", "lightning", "solana", "base")
+    pub chain: String,
+    /// The asset ticker (e.g. "BTC", "USDB", "USDC", "USDT")
+    pub asset: String,
```
danielgranhao: Should we also include a hard identifier? Ticker may not be specific enough. I think it would be good to include something like `identifier` or similar that is None for sats, but otherwise is either a contract address or a spark token identifier.

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -255,104 +255,221 @@ impl Payment {
...
+/// One side (source or destination) of a conversion.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
-pub struct ConversionStep {
-    /// The underlying payment id of the conversion step
-    pub payment_id: String,
-    /// Payment amount in satoshis or token base units
+pub struct ConversionSide {
+    /// The chain or network (e.g. "spark", "lightning", "solana", "base")
+    pub chain: String,
```
danielgranhao: Maybe also some kind of chain identifier, where applicable? Just concerned about how consumers will be able to reliably identify chains and assets

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -255,104 +255,221 @@ impl Payment {
...
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
+pub enum ConversionProvider {
+    /// AMM (Flashnet pool) conversion between token and BTC on Spark
+    Amm,
+    /// Orchestra cross-chain conversion
+    Orchestra,
+    /// Boltz reverse-swap cross-chain conversion
+    Boltz,
+}
```
danielgranhao, line 283: I think here we should expand each variant with provider-specific info about the conversion and maybe (not sure) remove the payment-details-internal conversion info as you mentioned.

So, e.g. Boltz would have swap id, lz_guid, etc

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -255,104 +255,221 @@ impl Payment {
...
+    /// Fee in the same base units
     pub fee: u128,
-    /// Method of payment
-    pub method: PaymentMethod,
-    /// Token metadata if a token is used for payment
-    pub token_metadata: Option<TokenMetadata>,
-    /// The reason the conversion amount was adjusted, if applicable.
+    /// Number of decimals for the asset (e.g. 6 for USDC/USDT/USDB).
+    /// None for BTC sides where sats formatting is used.
+    pub decimals: Option<u32>,
```
danielgranhao: Do we need the optional? Can decimals be 0 for sats?

crates/breez-sdk/core/src/models/mod.rs
danielgranhao: Just an idea but wdyt about making this an enum like e.g.
```
enum Network {
    Spark,
    Lightning,
    ExternalChain { chain_name: String, chain_id: Option<String>}
}
```

The idea is to minimize the use of untyped variants where we can.

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -255,104 +255,221 @@ impl Payment {
...
+    pub provider: ConversionProvider,
+    /// Status of this specific conversion step
+    pub status: ConversionStatus,
+    /// Source side of the conversion
+    pub from: ConversionSide,
+    /// Destination side of the conversion
+    pub to: ConversionSide,
+    /// Reason the conversion amount was adjusted, if applicable (AMM only)
     #[serde(default)]
     pub amount_adjustment: Option<AmountAdjustmentReason>,
```
danielgranhao, line 351: If we go for provider specific info in each ConversionProvider variant, this can move to ConversionProvider::Amm

crates/breez-sdk/core/src/utils/payments.rs
```diff
@@ -50,37 +54,122 @@ pub(crate) async fn get_payment_and_emit_event(
...
     id: String,
     storage: Arc<dyn Storage>,
 ) -> Result<Payment, SdkError> {
     let mut payment = storage.get_payment_by_id(id).await?;
+    enrich_payment_conversions(&mut payment, &storage).await?;
+    Ok(payment)
+}
+
+/// Enriches a single payment with its conversion details if applicable.
+async fn enrich_payment_conversions(
```
JssDWt: Just wondering whether this and other conversion related functions should live under token_conversion?

crates/breez-sdk/core/src/models/mod.rs
dangeross: I've addressed everything except this and amount_adjustment relocation. I think removing the ConversionInfo is a too big a change for this PR, mainly because it has knock-on effects of how the payment's conversion info is currently used internally

crates/breez-sdk/core/src/utils/payments.rs
dangeross: I've moved several fns to a utils::conversions module, as token_conversion is more for AMM only conversions (mainly sats <-> USDB)

PR #818, opened 2026-04-22
Cross-chain send: token → sats AMM conversion leg

danielgranhao:
Extends the send-with-conversion pipeline (SparkAddress / SparkInvoice / Bolt11 / BitcoinAddress) to CrossChain. Primary motivation: USDB-only stable-balance wallets can now pay Boltz destinations (Boltz accepts sats only) by auto-converting through the Flashnet AMM.

## What changed

- **New types**: `SourceAsset { Bitcoin, Token(String) }`, `CrossChainFeeMode { FeesExcluded, FeesIncluded }`.
- **`CrossChainRoutePair`** gains `supported_sources: Vec<SourceAsset>` so the prepare layer can decide whether auto-conversion is needed without provider-specific knowledge.
- **`CrossChainPrepared` / `SendPaymentMethod::CrossChainAddress`** gain `source_transfer_fee_sats: u64` (LN routing budget for Boltz, Spark transfer fee for Orchestra) and `fee_mode`. Boltz's `ln_fee_sats` moved out of the provider context into this public field — the send path enforces it as a hard cap.
- **Orchestra**: `get_routes` dedups by destination endpoint but accumulates source variants into `supported_sources`. `prepare` resolves the Orchestra-side `source_asset` string via the cached `spark_routes` (replaces the `Some(_) → "USDB"` hardcode, supports any future Spark source).
- **Boltz**: `prepare` grows a two-phase `FeesIncluded` path (throwaway invoice probes LN fee → real invoice sized to `amount − probe`, with a guard if the probe drifts between calls). `send` mirrors LNURL's overpayment logic so the full `FeesIncluded` budget is consumed.
- **`prepare_cross_chain_send`**: runs a source-selection decision tree (explicit options → direct send if route accepts the token → auto-inject `ToBitcoin` when the token is the stable-balance active token → defer to stable-balance auto-inject for sats), validates `amount > 0` + token balance sufficiency, forces `FeesIncluded` on conversion paths, and rejects `FromBitcoin + CrossChain`.
- **`execute_pre_send_conversion`**: `CrossChainAddress` branch added via a new `convert_token_for_crosschain` helper; the previous `Err` arm is removed.
- **Validation**: pre/post helpers reject `amount = 0`, `FromBitcoin + CrossChain`, and effective-source-not-in-`supported_sources`.
- **CLI**: route picker prints `supported_sources=[…]`; cross-chain confirmation prompt prints `Source transfer fee` alongside `Fee`.
- **Bindings**: WASM and Flutter mirrors updated for the new types and fields.

## What's out of scope

- FromBitcoin + CrossChain (sats → non-stable token send).
- Non-USDB token source routed through USDB via AMM (USDC → USDB → Orchestra).
- Idempotency key support for CrossChain conversion paths (tracked separately; retry may re-run the AMM leg).

### DIFF
--- a/crates/breez-sdk/core/src/cross_chain/boltz.rs
+++ b/crates/breez-sdk/core/src/cross_chain/boltz.rs
@@ -107,10 +107,12 @@ impl BoltzService {
     /// Two-phase prepare for `FeesIncluded`: size the real invoice to
     /// `amount - ln_fee_probe_sats` so `invoice_sats + ln_fee_probe <= amount`.
     ///
-    /// Mirrors LNURL pay's `FeesIncluded` pattern at `lnurl.rs`. A throwaway swap
-    /// is created to probe the LN fee; it times out server-side (~24h). The
-    /// probe value is stored on the prepare as `source_transfer_fee_sats`
-    /// and enforced as a hard cap at send time.
+    /// Mirrors LNURL pay's `FeesIncluded` pattern at `lnurl.rs`. Phase 1 uses
+    /// boltz-client's lightweight probe-invoice API — random preimage, short
+    /// server-side expiry, no HD index burn / DB row / WS subscription — so
+    /// the probe doesn't pile up persistent state. The probe value is stored
+    /// on the prepare as `source_transfer_fee_sats` and enforced as a hard
+    /// cap at send time.
     async fn prepare_fees_included(
         &self,
         recipient_address: &str,
@@ -124,16 +126,16 @@ impl BoltzService {
             route.chain
         );
 
-        // Phase 1: throwaway invoice at `total_sats` to probe LN fee.
-        let (_throwaway_prepared, throwaway_created) = self
-            .create_swap(
+        // Phase 1: throwaway probe invoice at `total_sats` to probe LN fee.
+        let probe_invoice = self
+            .fetch_probe_invoice(
                 recipient_address,
                 chain.clone(),
                 total_sats,
                 max_slippage_bps,
             )
             .await?;
-        let ln_fee_probe_sats = self.fetch_ln_fee(&throwaway_created.invoice).await?;
+        let ln_fee_probe_sats = self.fetch_ln_fee(&probe_invoice).await?;
 
         if total_sats <= ln_fee_probe_sats {
             return Err(SdkError::InvalidInput(format!(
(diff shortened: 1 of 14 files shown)


crates/breez-sdk/core/src/cross_chain/boltz.rs
```diff
@@ -65,6 +66,196 @@ impl BoltzService {
...
+    }
+
+    /// Two-phase prepare for `FeesIncluded`: size the real invoice to
+    /// `amount - ln_fee_probe_sats` so `invoice_sats + ln_fee_probe <= amount`.
+    ///
+    /// Mirrors LNURL pay's `FeesIncluded` pattern at `lnurl.rs`. A throwaway swap
+    /// is created to probe the LN fee; it times out server-side (~24h). The
+    /// probe value is stored on the prepare as `source_transfer_fee_sats`
+    /// and enforced as a hard cap at send time.
+    async fn prepare_fees_included(
```
dangeross, line 116: I think my only concern is we're 2x burning HD key indexes and storing swap data just for quotes. I know this is just for the fee included pattern, so maybe its a small concern.

(not a change request, just observation)

crates/breez-sdk/core/src/cross_chain/orchestra.rs
```diff
@@ -353,12 +383,38 @@ impl CrossChainService for OrchestraService {
...
+        let source_asset = match matched {
+            Some(r) => r.source.asset.clone(),
+            None => {
+                return Err(SdkError::InvalidInput(format!(
+                    "Orchestra does not offer a route for source {} → {}/{}",
+                    token_identifier.as_deref().unwrap_or("BTC"),
+                    dest_chain,
+                    dest_asset
+                )));
+            }
```
dangeross: I'm wondering why we need to do this, to check the token identifier is correct? Can we use the `route.supported_sources`'s `SourceAsset::Token(id)` to verify it?

crates/breez-sdk/core/src/sdk/payments.rs
```diff
@@ -594,10 +603,102 @@ impl BreezSdk {
...
+
+        // Boltz conversion + AmountIn + FeesExcluded is unfundable in the
+        // typical stable-balance case: the AMM consumes the token balance, so
+        // there's no separate sats reserve to cover LN routing fees. Reject
+        // the combination rather than supporting a partial set of cases.
+        if is_conversion && fee_policy == FeePolicy::FeesExcluded {
+            return Err(SdkError::InvalidInput(
+                "Token-conversion CrossChain sends require FeesIncluded fee policy.".to_string(),
+            ));
+        }
```
dangeross: For stable balance the API should act like you're sending sats as stable balance is auto filling conversion options. But it seems here the FeesIncluded has to be explicitly set, which goes against this notion. Perhaps passing the `Option<FeePolicy>` and resolving based on if the conversion options were auto filled here?

dangeross (commented):
Looking good

crates/breez-sdk/core/src/cross_chain/orchestra.rs
danielgranhao: It's to resolve the source asset symbol. It's what's used in the orchestra API as an identifier. `spark_routes()` is cached so this is cheap to do and IMO preferable to including the symbol in the SourceAsset::Token because it leaks orchestra details in the API

crates/breez-sdk/core/src/cross_chain/boltz.rs
danielgranhao: I've added a "probe" method to the client crate that creates a swap with a random preimage, minimum invoice expiry and doesn't subscribe to updates or persist to storage.

PR #820, opened 2026-04-23
Core: Add `Unrecoverable` error at startup for storage init

### LINKED ISSUE #815: Handle unrecoverable SDK initialization errors
When SDK initialization fails with an unrecoverable error (e.g., corrupted storage), app developers must take action by clearing the app's recoverable data and retrying initialization. We should detect these cases, return an appropriate error, and document the recommended recovery steps for developers.

hydra-yse:
Closes #815
This PR adds a `CorruptStorage` error variant to `SdkError`, making **migration** errors from `connect` more actionable for the user so they can know when to reset their storage. An example of how to catch the error and reset the storage has also been provided in the docs

### DIFF
--- a/crates/breez-sdk/core/src/error.rs
+++ b/crates/breez-sdk/core/src/error.rs
@@ -34,8 +34,8 @@ pub enum SdkError {
     NetworkError(String),
 
     /// Storage error
-    #[error("Storage error: {0}")]
-    StorageError(String),
+    #[error("Storage error: {message}")]
+    StorageError { recoverable: bool, message: String },
 
     #[error("Chain service error: {0}")]
     ChainServiceError(String),
@@ -117,7 +117,10 @@ impl From<crate::token_conversion::ConversionError> for SdkError {
                 SdkError::Generic("Duplicate transfer: conversion already handled".to_string())
             }
             ConversionError::Sdk(e) => e,
-            ConversionError::Storage(e) => SdkError::StorageError(e.to_string()),
+            ConversionError::Storage(e) => SdkError::StorageError {
+                recoverable: true,
+                message: e.to_string(),
+            },
             ConversionError::Wallet(e) => SdkError::SparkError(e.to_string()),
         }
     }
@@ -127,7 +130,18 @@ impl From<persist::StorageError> for SdkError {
     fn from(e: persist::StorageError) -> Self {
         match e {
             persist::StorageError::NotFound => SdkError::InvalidInput("Not found".to_string()),
-            _ => SdkError::StorageError(e.to_string()),
+            // Migration failures indicate corrupted storage that cannot be recovered by retrying.
+            // The app must clear its local data (storage directory) and call `connect` again.
+            persist::StorageError::MigrationError(err) => SdkError::StorageError {
+                recoverable: false,
+                message: format!(
+                    "Unrecoverable storage error, delete the storage directory and try again: {err}"
+                ),
+            },
+            _ => SdkError::StorageError {
+                recoverable: true,
+                message: e.to_string(),
+            },
         }
     }
 }
--- a/crates/breez-sdk/core/src/sdk_builder.rs
+++ b/crates/breez-sdk/core/src/sdk_builder.rs
@@ -635,8 +635,7 @@ impl SdkBuilder {
                         pool.clone(),
                         identity,
                     )
-                    .await
-                    .map_err(|e| SdkError::Generic(e.to_string()))?,
+                    .await?,
                 ));
             }
 
@@ -649,8 +648,7 @@ impl SdkBuilder {
             {
                 s = Some(Arc::new(
                     crate::persist::mysql::MysqlStorage::new_with_pool(pool.clone(), identity)
-                        .await
-                        .map_err(|e| SdkError::Generic(e.to_string()))?,
+                        .await?,
                 ));
             }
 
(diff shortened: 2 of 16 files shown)


crates/breez-sdk/core/src/error.rs
```diff
@@ -62,6 +62,11 @@ pub enum SdkError {
 
     #[error("Error: {0}")]
     Generic(String),
+
+    /// Initialization failed with an error that cannot be recovered from by retrying.
+    /// The app must clear its local data (storage directory) and call `connect` again.
+    #[error("Unrecoverable error: {0}")]
+    Unrecoverable(String),
```
JssDWt: What if we make this a storage error, and consider a storage error an unrecoverable error in our documentation? 

```suggestion
    Storage(String),
```

The comment about clearing the data directory is not relevant for postgres, let's consider that. I think we should document in the website that it's advised to consider the database unreachable (permissions?) or corrupted. Advising to clear the data directory is a good idea.

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -451,7 +451,8 @@ impl SdkBuilder {
                     .get_identity_public_key()
                     .await
                     .map_err(|e| SdkError::Generic(e.to_string()))?;
-                default_storage(&storage_dir, self.config.network, &identity_pub_key)?
+                default_storage(&storage_dir, self.config.network, &identity_pub_key)
+                    .map_err(|e| SdkError::Unrecoverable(e.to_string()))?
```
roeierez: This is too generic and seems too risky to treat every storage error as unrecoverable especially if we instruct developers to delete the working directory based on this error.
Let's start by pin point the specific error from the support case we had and we should move incrementally here upon discovering more cases.

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -469,7 +470,7 @@ impl SdkBuilder {
                 Arc::new(
                     crate::persist::postgres::PostgresStorage::new_with_pool(pool.clone())
                         .await
-                        .map_err(|e| SdkError::Generic(e.to_string()))?,
+                        .map_err(|e| SdkError::Unrecoverable(e.to_string()))?,
```
roeierez: I don't think we should return this in the postgres case. This is used on backend service where the caller has full control of the process.

roeierez (changes requested):

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -684,7 +690,14 @@ fn default_storage(
...
 ) -> Result<Arc<dyn Storage>, SdkError> {
+    use crate::persist::StorageError;
+
     let db_path = crate::default_storage_path(data_dir, &network, identity_pub_key)?;
-    let storage = Arc::new(crate::SqliteStorage::new(&db_path)?);
-    Ok(storage)
+    let storage = crate::SqliteStorage::new(&db_path).map_err(|e| match e {
+        // Migration failures indicate corrupted storage that cannot be recovered by retrying.
+        // The app must clear its local data (storage directory) and call `connect` again.
+        StorageError::MigrationError(_) => SdkError::CorruptStorage(e.to_string()),
```
JssDWt: I wonder whether we still need to be more specific here. I think there's quite some cases where there are storage errors that are not fixed with deleting the database / data directory that will also fall in this category. Should we be more specific in this error mapping?

JssDWt (commented):
Looking good!

crates/breez-sdk/core/src/sdk_builder.rs
hydra-yse: > I think there's quite some cases where there are storage errors that are not fixed with deleting the database / data directory that will also fall in this category.

Just thinking, but my idea was that `CorruptStorage` should be called only when it's recoverable by db deletion. Regarding the other (critical) errors, we can definitely be more explicit, but that would require a different error variant. What do you think?

JssDWt: I think other critical errors that do not fall under the category of 'recoverable by db deletion' can go under 'Generic' then.

docs/breez-sdk/snippets/swift/BreezSdkSnippets/Sources/GettingStarted.swift
```diff
@@ -124,4 +125,32 @@ func gettingStartedSparkStatus() async throws {
...
+    } catch {
+        if case SdkError.storageError(let recoverable, _) = error, !recoverable {
+            // The SDK storage is corrupted and cannot be recovered by retrying.
+            // Clear the storage directory and reconnect with fresh storage.
+            try? FileManager.default.removeItem(atPath: storageDir)
+            return try await connect(request: makeRequest())
+        }
+        throw error
+    }
+}
```
roeierez, line 155: Not sure we want to set this as the default code to use the sdk. If this is the right approach than why not doing it automatically inside the sdk...
Sounds like better to propagate the error upstream to the app level and let the app handle that? Perhaps need some user interaction here?

docs/breez-sdk/src/guide/initializing.md
```diff
@@ -34,6 +34,19 @@ Instead of managing mnemonics directly, you can use passkeys to derive wallet se
 
 See [Connecting with a Passkey](passkey.md) for the full setup guide including PRF provider implementation, platform configuration, and label management.
 
+## Handling Initialization Failures
+
+Some initialization failures indicate that the local SDK storage is corrupted and retrying without action will not help. When {{#name connect}} returns a `StorageError` with `recoverable` set to false, the app must clear the SDK storage directory and retry.
```
roeierez, line 39: Not sure we should be so decisive here using the terms "must clear the sdk storage".
It basically means the storage is corrupted and given it is a rare case perhaps the app needs to upstream this issue to the user and explicitly ask the user to reset (with a clear warning on the local settings).

roeierez (changes requested):

PR #822, opened 2026-04-23
Fix static deposit claim: retry transfer lookup and clean up stale records

danielgranhao:
Fixes two issues in the static deposit claim flow:

- **Transfer lookup race**: after the SSP commits a claim, the receiver side can briefly fail to see the transfer via `query_all_transfers`, causing a spurious `"transfer not found"` error. Now retried with exponential backoff (up to 5 attempts, 500ms–5s).
- **Stale unclaimed-deposit record**: when the sync claim path fails but the event-driven path later succeeds, the deposit record with a `claim_error` lingered in storage until the next background claim iteration. On `TransferClaimed`, if the transfer originated from a `ClaimStaticDeposit` request, the corresponding `(txid, vout)` record is now deleted.

### DIFF
--- a/crates/spark-wallet/src/wallet.rs
+++ b/crates/spark-wallet/src/wallet.rs
@@ -660,8 +660,7 @@ impl SparkWallet {
     }
 
     /// Submits a static deposit claim and returns the resulting transfer id.
-    /// Use [`Self::query_static_deposit_claim_transfer`] to fetch the full
-    /// transfer when needed.
+    /// The transfer can then be fetched with [`Self::list_transfers`].
     pub async fn claim_static_deposit(
         &self,
         quote: StaticDepositQuote,
@@ -671,26 +670,6 @@ impl SparkWallet {
         self.maybe_start_optimization().await;
 
         Ok(transfer_id)
-    }
-
-    /// Looks up the transfer produced by a static deposit claim, retrying
-    /// while the operator pool has not yet indexed it.
-    pub async fn query_static_deposit_claim_transfer(
-        &self,
-        transfer_id: String,
-    ) -> Result<WalletTransfer, SparkWalletError> {
-        let transfer = self
-            .deposit_service
-            .query_static_deposit_claim_transfer(transfer_id)
-            .await?;
-
-        Ok(WalletTransfer::from_transfer(
-            transfer,
-            None,
-            None,
-            self.identity_public_key,
-            self.config.service_provider_config.identity_public_key,
-        ))
     }
 
     pub async fn refund_static_deposit(
--- a/crates/spark-wallet/src/wallet.rs
+++ b/crates/spark-wallet/src/wallet.rs
@@ -659,13 +659,30 @@ impl SparkWallet {
         Ok(deposit_nodes.into_iter().map(WalletLeaf::from).collect())
     }
 
+    /// Submits a static deposit claim and returns the resulting transfer id.
+    /// Use [`Self::query_static_deposit_claim_transfer`] to fetch the full
+    /// transfer when needed.
     pub async fn claim_static_deposit(
         &self,
         quote: StaticDepositQuote,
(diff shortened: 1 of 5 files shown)


dev-vektr-ai:
Looks good to me overall. The retry/backoff around `query_all_transfers` addresses the receiver-side visibility race cleanly, and cleaning up the `(txid, vout)` record on `TransferClaimed` closes the stale `claim_error` storage gap from the sync/event-path split.

I did not spot any blocking issues in the changes shown here. A follow-up test covering the transient empty-result path and the event-driven cleanup would be nice if it is easy to add, but nothing in this patch looks incorrect to me.

crates/spark/src/services/deposit.rs
```diff
@@ -40,6 +41,13 @@ use super::ServiceError;
 
 const CLAIM_STATIC_DEPOSIT_ACTION: &str = "claim_static_deposit";
 
+// Retry parameters for looking up the transfer created by a static deposit
+// claim. The SSP creates the transfer synchronously, but it can take a brief
+// window before it becomes visible to the receiver via the operator pool.
+const CLAIM_TRANSFER_LOOKUP_MAX_ATTEMPTS: u32 = 5;
+const CLAIM_TRANSFER_LOOKUP_BASE_DELAY_MS: u64 = 500;
+const CLAIM_TRANSFER_LOOKUP_MAX_DELAY_MS: u64 = 5_000;
```
roeierez: My concern it will may take a lot of time and hold the sync. Perhaps 3 attempt max with 2 seconds max delay?

roeierez (commented):
Still going over this. Just a small comment until I finish.

crates/spark/src/services/deposit.rs
danielgranhao: Done

JssDWt:
I think it would be possible also to not return the transfer if it doesn't exist yet on the SO side. Return `None`, or a partially initialized struct, or never return it at all, only the ssp details. We'll get the transfer eventually via the event. That would take the time pressure off the sync. It would be eventually consistent. Only from the publicly exposed claim_deposit we need to return a fully initialized payment. Perhaps we can query the transfer there?

danielgranhao:
> I think it would be possible also to not return the transfer if it doesn't exist yet on the SO side. Return `None`, or a partially initialized struct, or never return it at all, only the ssp details. We'll get the transfer eventually via the event. That would take the time pressure off the sync. It would be eventually consistent. Only from the publicly exposed claim_deposit we need to return a fully initialized payment. Perhaps we can query the transfer there?

Good idea. Now we don't query the transfer on the sync path and query the transfer only on public `claim_deposit`.

crates/spark-wallet/src/wallet.rs
```diff
@@ -659,14 +659,31 @@ impl SparkWallet {
...
+        let transfer_id = self.deposit_service.claim_static_deposit(quote).await?;
 
         self.maybe_start_optimization().await;
 
+        Ok(transfer_id)
+    }
+
+    /// Looks up the transfer produced by a static deposit claim, retrying
+    /// while the operator pool has not yet indexed it.
+    pub async fn query_static_deposit_claim_transfer(
```
roeierez: This is actually query_transfer generic code right? Not related specifically to deposit.
Given we already have list_transfers (with the ability to pass a specific transfer id) wouldn't this be redundant as public API?
Also since it is used only at the sdk level perhaps we don't need the implementation of it in the deposit service and implement a small retry loop at the sdk level?

roeierez (changes requested):

crates/spark-wallet/src/wallet.rs
danielgranhao: You're right. Done

roeierez (approved):
LGTM

PR #824, opened 2026-04-24
point frost packages to lightspark repository

JssDWt:
The unofficial frost crates were yanked.

Use the same frost crates as [spark library](https://github.com/buildonspark/spark/blob/4101f4ce9be44ad67ebfeea7474da28a8adfca2a/signer/Cargo.toml#L20-L21)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -746,7 +746,7 @@ dependencies = [
  "built",
  "chrono",
  "flashnet",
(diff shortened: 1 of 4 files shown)


roeierez (approved):

danielgranhao (approved):

dangeross (approved):
Looks like the Flutter `Cargo.lock` needs updating, otherwise LGTM

PR #825, opened 2026-04-24
bump SO protos

JssDWt:
Spark added a heartbeat to the SO event stream, causing us to log a warning every 5 seconds:

```
2026-04-24T07:29:06.372600Z  WARN spark::events::server_stream: 81: Received empty event, skipping
2026-04-24T07:29:11.373407Z  WARN spark::events::server_stream: 81: Received empty event, skipping
```

Bump the protos so we understand this event, and the log goes to trace level.

### DIFF
--- a/crates/spark/protos/spark/spark.proto
+++ b/crates/spark/protos/spark/spark.proto
@@ -93,6 +93,8 @@ service SparkService {
 
     rpc update_wallet_setting(UpdateWalletSettingRequest) returns (UpdateWalletSettingResponse) {}
     rpc query_wallet_setting(QueryWalletSettingRequest) returns (QueryWalletSettingResponse) {}
+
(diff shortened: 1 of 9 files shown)


crates/spark/protos/spark/spark.proto
```diff
@@ -106,17 +108,29 @@ message SubscribeToEventsResponse {
         DepositEvent deposit = 2;
         ConnectedEvent connected = 3;
         TransferEvent sender_transfer = 4;
+        HeartbeatEvent heartbeat = 5;
+        TokenTransactionEvent token_transaction = 6;
```
dangeross, line 112: This might be something we can use to trigger a sync @danielgranhao

dangeross (approved):
LGTM

crates/spark/protos/spark/spark.proto
danielgranhao: Nice -> #826

danielgranhao (approved):

PR #827, opened 2026-04-24
Set rust-toolchain to 1.95 and fix clippy lints

danielgranhao:


JssDWt (approved):
What should our bumping strategy look like?

danielgranhao:
> What should our bumping strategy look like?

Up until now we have been on whatever the CI image includes. This results in the version changing unexpectedly and causing sudden clippy errors. This proposes sticking to a specific version on the project so that the same version is used everywhere. I went for the latest stable version, but I'm also fine with going with another slightly older version to adopt a lagging approach.

JssDWt:
> Up until now we have been on whatever the CI image includes. This results in the version changing unexpectedly and causing sudden clippy errors. This proposes sticking to a specific version on the project so that the same version is used everywhere. I went for the latest stable version, but I'm also fine with going with another slightly older version to adopt a lagging approach.

I think latest is good. We can just bump the version when we think the time is right.

roeierez (approved):

PR #828, opened 2026-04-24
improve sync time

JssDWt:
When there are no htlcs to be queried, `transfer_ids` was empty. That caused a query for all htlcs. In an integration test I added that accidentally used a fixed, already used seed, that made calls to query_htlc taking 7-9 seconds, because it was querying all htlcs. By not querying when we don't have to, we save that time.

Additionally, some integration tests that do payments use a non-random seed. So sync takes an awfully long time at some point. Using a random seed should speed things up because we start with a blank slate.


roeierez (approved):
Good optimization!

PR #829, opened 2026-04-24
LNURL: Atomic lightning address transfer

dangeross:
Adds atomic Lightning address username transfer between pubkeys. A user can now hand a registered username from one pubkey to another in a single server-side transaction — closing the race window where today's delete-then-reregister flow could let a third party snipe the name.

**How it works**
1. **Current owner (A)** calls `accept_lightning_address_transfer` with the new owner's `identity_pubkey`. The method returns a `LightningAddressTransfer` containing `pubkey` and `signature`, which is the authorization that grants B the right to take over the username.
2. **New owner (B)** calls `register_lightning_address` with the new optional `transfer: Some({ pubkey, signature })` field populated. The SDK routes to the server's new `POST /lnurlpay/{to_pubkey}/transfer` endpoint, which verifies both signatures and swaps ownership in one DB transaction (`SELECT … FOR UPDATE` → `DELETE` source → `INSERT … ON CONFLICT DO UPDATE` target).


crates/breez-sdk/lnurl/src/postgresql/repository.rs
```diff
@@ -105,6 +105,59 @@ impl crate::repository::LnurlRepository for LnurlRepository {
...
+                .bind(domain)
+                .bind(from_pubkey)
+                .fetch_optional(&mut *tx)
+                .await?;
+        match source_name {
+            Some((name,)) if name == username => {}
+            _ => return Err(LnurlRepositoryError::SourceNotOwner),
+        }
+
+        sqlx::query("DELETE FROM users WHERE domain = $1 AND pubkey = $2")
```
JssDWt: DELETE has a RETURNING option. We could do RETURNING name, use that as the source_name, and error if source not owner. It will work because the tx is not committed then.

docs/breez-sdk/src/guide/receive_lnurl_pay.md
```diff
@@ -65,6 +65,26 @@ You can retrieve information about the currently registered Lightning address.
 
 {{#tabs lightning_address:get-lightning-address}}
 
+### Transferring a Lightning address
+
+A user who already owns a registered Lightning address can hand it over to a different owner (pubkey) in a single atomic server operation — ownership is removed from the old pubkey and the new pubkey takes it in one step, without exposing a window during which the username could be snatched by a third party.
+
+The flow has two steps, run on two different SDKs:
+
+**Step 1 — current owner (pubkey A):** produce a transfer authorization by signing a fixed message of the form `transfer:{pubkey_a}-{username}-{pubkey_b}`. Use {{#name sign_message}} on the SDK that currently owns the username. `pubkey_b` is the {{#name identity_pubkey}} of the receiving pubkey (available via {{#name get_info}}). The `username` must be the sanitized (lowercased and trimmed) form.
```
JssDWt: Should we add an sdk function `accept_lightning_address_transfer`, returning the needed info for the `transfer` field? This is not the worst DX, but not the best either.

docs/breez-sdk/src/guide/receive_lnurl_pay.md
```diff
@@ -65,6 +65,26 @@ You can retrieve information about the currently registered Lightning address.
 
 {{#tabs lightning_address:get-lightning-address}}
 
+### Transferring a Lightning address
```
JssDWt, line 68: I think it's worth a note that your existing payments are not transferred to the new owner. Only the address.

JssDWt (commented):
Yep that will work! I like how both parties agree on the transfer this way.

JssDWt (approved):
LGTM

Maybe the rtsync integration test error is related though:
`Error: SparkSdkError: Service error: service provider error: serialization error: Json error: expected value at line 1 column 1`

dangeross:
> Maybe the rtsync integration test error is related though

Isn't it an SSP error?

JssDWt:
> Isn't it an SSP error?

Ah yes it is

roeierez:
@dangeross I think we mentioned that offline once. AFAIU the use case is a migration of one user to a new account.
In that case I believe the transfer can be executed using one request containing the address and the new pubkey signed by both keys.
Any drawbacks with this approach?

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -1520,6 +1520,43 @@ pub struct RegisterLightningAddressRequest {
...
+/// owner. The transferee passes it as the `transfer` field when registering.
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct LightningAddressTransfer {
+    /// Hex-encoded secp256k1 compressed public key of the current owner.
+    pub pubkey: String,
+    /// Hex-encoded DER ECDSA signature by the current owner over
+    /// `"transfer:{username}-{transferee_pubkey}"`. The transferee signs the
+    /// same canonical message with their own key when submitting.
+    pub signature: String,
```
roeierez: This doesn't contain any timestamp for validation which makes me think it can be re-used later by the same tranferee?

crates/breez-sdk/core/src/sdk/lightning_address.rs
```diff
@@ -68,6 +76,43 @@ impl BreezSdk {
...
+        let message = format!(
+            "transfer:{}-{}",
+            address_info.username, request.transferee_pubkey,
+        );
+        let signature = self.spark_wallet.sign_message(&message).await?;
+        Ok(LightningAddressTransfer {
+            pubkey: self_pubkey,
+            signature: signature.serialize_der().to_lower_hex_string(),
+        })
+    }
```
roeierez: I wonder if we need this API. We do have sign_message at the sdk level so a use that migrate from one wallet to another can prepare the whole transfer package and use the public api to sign.

roeierez (changes requested):

crates/breez-sdk/core/src/models/mod.rs
dangeross: Yes, but can only be executed once. Any replay is rejected by the server

crates/breez-sdk/core/src/sdk/lightning_address.rs
JssDWt: It is quite specific though isn't it? Signing transfer:{}-{} is a protocol-matter. I think it's nice to keep this encapsulated, so it stays under our control. We can argue easier about changes, etc. For example adding a timestamp would be one of these changes that we couldn't do if people construct that message themselves.

roeierez: Here is how I see the flow in my suggestion with two users A(the address owner) B(the address receiver):
1. User B prepares a "transfer" package with all details (address, both pubkeys, timestamp, etc...) our own API
2. User B passes this package to sign with user A public `sign_message` API
3. User B use the signed package in the register API to transfer the address.

I think it is still encapsulated as the package to sign is created by the sdk with the advantage of not adding business logic also for the sender in this (a bit) complicated use case.
We can discuss this more offline.

JssDWt (approved):
LGTM

roeierez (approved):

PR #830, opened 2026-04-25
ssp client: retry on 5xx responses

JssDWt:
In #829 CI [failed](https://github.com/breez/spark-sdk/actions/runs/24909842211/job/72948579576?pr=829) with an integration test error. I'm assuming the response is something like a 502 error. 

```
Error: SparkSdkError: Service error: service provider error: serialization error: Json error: expected value at line 1 column 1
```

This PR adds retry to the graphql client for the SSP. So if 502 is returned, we retry once with a delay of 50ms. The retry is configurable.


roeierez (approved):
LGTM

danielgranhao (approved):

PR #833, opened 2026-04-27
Robust cancel reservation: verify operator state before releasing leaves

roeierez:
Two layers of defense against orphaned leaves where the operator-side state diverges from the wallet's local view after an ambiguous RPC failure.

**Phase 1 - SSP request_swap retry**
After `initiate_swap_primary_transfer` succeeds, an SSP `request_swap` failure used to leave the operator holding the leaves while the SDK cancelled the local reservation. Wrap the SSP call in a bounded retry (5 attempts, 100→3000 ms exp backoff) on Network errors. Idempotency key is the existing `user_outbound_transfer_external_id`

**Phase 2 — cancel-time leaf verification**
`TreeService::cancel_reservation` now queries the coordinator before releasing any leaves. Leaves the coordinator reports as not-Available, not owned by the wallet's identity key, or missing from the response are deleted from the local store rather than returned to the available pool. Pessimistic on coordinator-query failure (drop everything) as it is better missing leaves than risking in stale leaves as available. Next refresh will sync if needed.

Added NodeJS bench framework to verify using sqlite and postgres.

### DIFF
--- a/crates/breez-sdk/breez-bench/js/README.md
+++ b/crates/breez-sdk/breez-bench/js/README.md
@@ -1,0 +1,73 @@
+# Concurrent Spark transfer benchmark — Node.js wasm
+
+Counterpart to the Rust `concurrent-perf` bench (`crates/breez-sdk/breez-bench/src/bin/concurrent_perf.rs`). Measures send-payment throughput for `@breeztech/breez-sdk-spark/nodejs` against a real regtest network using the public Lightspark faucet.
+
+## Setup
+
+Requires Node.js with native module support for `better-sqlite3`. Tested on Node 22.
+
+```sh
+cd crates/breez-sdk/breez-bench/js
+nvm use 22   # or any node >=18
+npm install
+```
+
+The package depends on `../../../../packages/wasm` (the local wasm package). Rebuild the wasm package via `cargo xtask package wasm::node` if you change SDK source.
+
+## Required env
+
+- `FAUCET_USERNAME`, `FAUCET_PASSWORD` — Lightspark regtest faucet credentials.
+- `BREEZ_API_KEY` — optional (regtest doesn't require one).
+
+## Backends
+
+The bench supports SQLite (default, single instance) and Postgres (single or multi-instance).
+
+For Postgres, point `--sender-postgres` and `--receiver-postgres` at running databases. A local docker postgres works:
+
+```sh
+docker run -d --name spark-perf-pg -e POSTGRES_PASSWORD=postgres -p 5544:5432 postgres:16
+docker exec spark-perf-pg psql -U postgres -c 'CREATE DATABASE bench_sender_node;'
+docker exec spark-perf-pg psql -U postgres -c 'CREATE DATABASE bench_receiver_node;'
+```
+
+Multi-instance senders (`--sender-instances N` where N > 1) require `--sender-postgres` so all instances share one tree store.
+
+## Run
+
+SQLite, single instance:
+
+```sh
+FAUCET_USERNAME=… FAUCET_PASSWORD=… node concurrent_perf.js \
+    --total-payments 50 --concurrency 6
+```
+
+Postgres, multi-instance shared wallet:
+
+```sh
+FAUCET_USERNAME=… FAUCET_PASSWORD=… node concurrent_perf.js \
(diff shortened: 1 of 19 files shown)


crates/breez-sdk/breez-bench/js/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+Counterpart to the Rust `concurrent-perf` bench (`crates/breez-sdk/breez-bench/src/bin/concurrent_perf.rs`). Measures send-payment throughput for `@breeztech/breez-sdk-spark/nodejs` against a real regtest network using the public Lightspark faucet.
+
+## Setup
+
+Requires Node.js with native module support for `better-sqlite3`. Tested on Node 22.
+
+```sh
+cd crates/breez-sdk/breez-bench/js
+nvm use 22   # or any node >=18
```
danielgranhao: We require [node >= 22](https://sdk-doc-spark.breez.technology/guide/install_javascript.html)
```suggestion
nvm use 22   # or any node >=22
```

crates/spark/src/tree/service.rs
```diff
@@ -381,8 +387,18 @@ impl SynchronousTreeService {
...
+                    .leaves
+                    .iter()
+                    .map(|l| l.id.to_string())
+                    .collect();
+                warn!(
+                    "leaf_lifecycle swap_failed_in_select: reservation={} leaf_ids={:?} error={:?}",
+                    reservation.id, reserved_leaf_ids, e
+                );
+                if let Err(cancel_err) =
+                    <Self as TreeService>::cancel_reservation(self, reservation).await
```
danielgranhao:
```suggestion
                if let Err(cancel_err) = self.cancel_reservation(reservation).await {
```

crates/spark/src/tree/service.rs
```diff
@@ -520,15 +628,16 @@ impl SynchronousTreeService {
...
     ) -> Result<LeavesReservation, TreeServiceError> {
+        let id = reservation.id.clone();
+        let cancel_input = reservation.clone();
         let new_leaves = self
             .check_renew_nodes(reservation.leaves, async |e| {
-                // Cancel the reservation if the timelock check fails
-                if let Err(e) = self.state.cancel_reservation(&reservation.id).await {
-                    error!("Failed to cancel reservation: {e:?}");
+                if let Err(err) =
+                    <Self as TreeService>::cancel_reservation(self, cancel_input).await
```
danielgranhao:
```suggestion
                if let Err(err) = self.cancel_reservation(cancel_input).await {
```

crates/spark/src/tree/service.rs
```diff
@@ -44,8 +44,14 @@ impl TreeService for SynchronousTreeService {
...
-        self.state.cancel_reservation(&id).await
+    async fn cancel_reservation(
+        &self,
+        reservation: LeavesReservation,
+    ) -> Result<(), TreeServiceError> {
+        let leaves_to_keep = self.verify_leaves_against_coordinator(&reservation).await;
+        self.state
+            .cancel_reservation(&reservation.id, &leaves_to_keep)
+            .await
     }
```
danielgranhao, line 55: Worth noting that fetch_coop_exit_fee_quote cancels reservation on the happy path and never actually does anything with them, so there would be no need to verify and we are adding latency.

But not something for this PR. I think the solution is to stop reserving leaves for quoting in the first place. There's no need to hang on to the leaves

danielgranhao (commented):
Looks good, just nits and a comment where no action is needed

crates/spark/src/tree/service.rs
roeierez: I agree with everything you said. I wanted to introduce first a generic mechanism (even with some slightly more overhead on edge cases).

crates/breez-sdk/breez-bench/js/README.md
roeierez: fixed

crates/spark/src/tree/service.rs
roeierez: fixed

crates/spark/src/tree/service.rs
roeierez: fixed

danielgranhao (approved):

PR #834, opened 2026-04-27
Require amount when buying BTC via Cash App

erdemyerebasmaz:
This PR makes the amount mandatory when buying BTC via Cash App.

We removed amountless invoice support on Cash App because the scope of the `buy_bitcoin` API is on-ramping. With an amountless invoice, Cash App only lets the payer fund from their existing Cash App BTC balance, which isn't on-ramping. When the invoice carries an amount, Cash App opens up funding via fiat balance and debit card.

### DIFF
--- a/docs/breez-sdk/src/guide/buy_bitcoin.md
+++ b/docs/breez-sdk/src/guide/buy_bitcoin.md
@@ -23,11 +23,14 @@ MoonPay supports Apple Pay and Google Pay, but these payment methods will not wo
 
 ## CashApp
 
-CashApp uses **Lightning (bolt11 invoices)** to receive purchased funds. The SDK generates a Lightning invoice and returns a CashApp deep link (`cash.app/launch/lightning/...`) that opens the CashApp for the user to complete payment.
+CashApp uses **Lightning (bolt11 invoices)** to receive purchased funds. The caller specifies the amount in satoshis; the SDK generates a bolt11 invoice for that amount and returns a CashApp deep link (`cash.app/launch/lightning/...`) that opens CashApp so the user can complete payment.
 
 <div class="warning">
-<h4>Developer note</h4>
-CashApp is only available on <strong>mainnet</strong>. Attempting to use CashApp on testnet or regtest will return an error.
+<h4>Developer notes</h4>
+<ul>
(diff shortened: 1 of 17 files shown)


roeierez (approved):

docs/breez-sdk/src/guide/buy_bitcoin.md
```diff
@@ -23,11 +23,14 @@ MoonPay supports Apple Pay and Google Pay, but these payment methods will not wo
...
-CashApp uses **Lightning (bolt11 invoices)** to receive purchased funds. The SDK generates a Lightning invoice and returns a CashApp deep link (`cash.app/launch/lightning/...`) that opens the CashApp for the user to complete payment.
+CashApp uses **Lightning (bolt11 invoices)** to receive purchased funds. The caller specifies the amount in satoshis; the SDK generates a bolt11 invoice for that amount and returns a CashApp deep link (`cash.app/launch/lightning/...`) that opens CashApp so the user can complete payment.
 
 <div class="warning">
-<h4>Developer note</h4>
-CashApp is only available on <strong>mainnet</strong>. Attempting to use CashApp on testnet or regtest will return an error.
+<h4>Developer notes</h4>
+<ul>
+<li>CashApp is only available on <strong>mainnet</strong>. Using CashApp on testnet or regtest returns an error.</li>
+<li>The amount is <strong>required</strong>. With an amountless invoice, Cash App only lets the payer fund from their existing Cash App BTC balance. When the invoice carries an amount, Cash App opens up funding via fiat balance and debit card.</li>
```
danielgranhao, line 32: Given the SDK forces an amount to be provided, do we need this note?

danielgranhao (approved):
Looks good, just a question on docs but doesn't block merge

docs/breez-sdk/src/guide/buy_bitcoin.md
erdemyerebasmaz: It's not needed, but I thought it would help partners to understand the constraint and it gives us something to point to if someone asks about why amountless invoices are not supported on Cash App.

danielgranhao (approved):

PR #835, opened 2026-04-27
Split kotlin-multiplatform CI into three single-purpose pathways

erdemyerebasmaz:
The Docs / kotlin-mpp job takes ~30 minutes because it builds every Kotlin Multiplatform target (JVM, Android, three iOS targets) and uses uncached Gradle distributions, Konan toolchains, and `cargo install` builds for the binary tools, even though the doc-snippets check only needs the JVM publication.

This PR splits the work across three single-purpose jobs that run in parallel:

| Job | Runs on | What it covers |
|---|---|---|
| **Docs / kotlin-multiplatform** | `ubuntu-latest` | Snippet shape against the SDK's JVM publication |
| **CLI / kotlin-multiplatform-android** | `ubuntu-latest` | NDK cross-compile of Rust bindings + Android Kotlin compile + AAR assembly |
| **CLI / kotlin-multiplatform-ios** | `macos-latest` | All three iOS Kotlin/Native target compiles (`compileKotlinIos{Arm64,X64,SimulatorArm64}`) |

The `Docs / kotlin-multiplatform` job also gets an opt-in `-PskipIosTargets` Gradle flag and a new `build-kotlin-multiplatform-bindgen-host` Make target, replacing `package-kotlin-multiplatform-dummy-binaries`. 

The Android job uses a new `build-kotlin-multiplatform-bindgen-android-arm64` Make target that runs `cargo-ndk` for one arch + gobley + copies the `.so` into `androidMain/jniLibs/` for AAR assembly. 

**Caches added:** 
- Gradle (`gradle/actions/setup-gradle@v4`), 
- `~/.konan` (iOS job),
- `~/.cargo/bin/gobley-uniffi-bindgen` (all three jobs), 
- `~/.cargo/bin/cargo-ndk` (Android job).

**Also (incidental cleanup):** 
- `Docs / *` matrix is rust-first then alphabetical;
- `CLI / *` matrix groups `kotlin-multiplatform`/`-android`/`-ios` together;
-  the `go` doc-snippet platform is renamed to `golang` to match `cli-ci`.

## Coverage

Total per-PR coverage is preserved
- JVM, Android, and iOS Kotlin/Native are all compiled per PR; 
- multi-variant publish (POM/module-file generation for every variant) is exercised at release time via `publish-kotlin-multiplatform.yml`.

## Timings

| Job | Before | After (cold) | After (warm) |
|---|---|---|---|
| `Docs / kotlin-multiplatform` | bundled in ~28 min | 8m 45s | 6m 58s |
| `CLI / kotlin-multiplatform` | ~8 min | 7m 54s | 4m 46s |
| `CLI / kotlin-multiplatform-android` | - | 8m 52s | 5m 23s |
| `CLI / kotlin-multiplatform-ios` | - | 11m 20s | 6m 26s |
| **Critical-path wall time** | **~28 min** | **~11 min** | **~7 min** |

###### **cold** = first run after force-push with caches evicted; **warm** = subsequent run reading caches saved by the previous one. 

Cache hit confirmation on the warm run (other than Swatinem rust-cache, which hits everywhere):
- iOS: ✅ Konan, ✅ gobley
- Android: ✅ cargo-ndk, ✅ gobley
- Docs: ✅ gobley

The Gradle User Home cache misses on every job because `gradle/actions/setup-gradle@v4` is read-only on PR runs by default. Enabling write-through-on-internal-PRs would shave another ~1 min off Docs in particular; left as a follow-up.

### DIFF
--- a/.github/workflows/cli-ci.yml
+++ b/.github/workflows/cli-ci.yml
@@ -3,6 +3,20 @@ name: CLI static analysis
 on:
   workflow_call:
 
+# Version pins for the kmp-* matrix jobs. Bump these in lockstep with the
+# canonical source so the cache keys invalidate alongside the binary itself.
+env:
+  # cargo-ndk 4.x is required by the `--link-libcxx-shared` flag the
+  # bindings Makefile passes (see `build-ndk-release-target-%`); the flag
+  # was added in 4.0.
+  CARGO_NDK_VERSION: 4.1.2
+  # Mirror of the gobley branch pinned in
+  # `crates/breez-sdk/bindings/Makefile` (`install-uniffi-bindgen-gobley`).
+  GOBLEY_REF: v0.3.7-with-fixes
+  # Mirror of `kotlin_target_version` in
+  # `crates/breez-sdk/bindings/uniffi.kotlin-multiplatform.toml`.
+  KOTLIN_NATIVE_VERSION: 1.9.21
+
 jobs:
   check:
     name: ${{ matrix.lang }}
@@ -89,9 +103,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: ~/.konan
-          # Bump the suffix when the Kotlin version pinned in
-          # crates/breez-sdk/bindings/uniffi.kotlin-multiplatform.toml changes.
-          key: konan-1.9.21-${{ runner.os }}
+          key: konan-${{ env.KOTLIN_NATIVE_VERSION }}-${{ runner.os }}
 
       - name: Add Android Rust target
         if: matrix.android-ndk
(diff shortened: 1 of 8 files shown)


.github/workflows/cli-ci.yml
```diff
@@ -63,6 +80,64 @@ jobs:
...
+      - name: Add Android Rust target
+        if: matrix.android-ndk
+        run: rustup target add aarch64-linux-android
+
+      - name: Cache cargo-ndk
+        if: matrix.android-ndk
+        uses: actions/cache@v4
+        with:
+          path: ~/.cargo/bin/cargo-ndk
+          key: cargo-ndk-${{ runner.os }}
```
danielgranhao: Should we also add a version pin to cache key?

danielgranhao (commented):
Nice!

.github/workflows/cli-ci.yml
erdemyerebasmaz: Yeah, it was an oversight. I'm adding the version pin and moving them all to env variables for more visibility & so that they are easier to update.

erdemyerebasmaz: Hoisted all three version pins (`CARGO_NDK_VERSION`, `GOBLEY_REF`, `KOTLIN_NATIVE_VERSION`) to a workflow-level `env` block in a6a8e5ee. Cache keys evaluate to the same strings, so existing caches still hit.

erdemyerebasmaz: Addressed:
- [ci: pin cargo-ndk version in cache key and install](https://github.com/breez/spark-sdk/pull/835/commits/08145f02047ad1b7b22b4c699986beed460de679)
- [ci: hoist kmp version pins to workflow env](https://github.com/breez/spark-sdk/pull/835/commits/a6a8e5eea6bf7dd884d1373369f59b436ae6942a)

danielgranhao (approved):

.github/workflows/cli-ci.yml
erdemyerebasmaz: Squashed and force-pushed in c970876e. Final state: `CARGO_NDK_VERSION: 4.1.2` (cargo-ndk 4.x is required for the `--link-libcxx-shared` flag the Makefile passes; that flag was added in cargo-ndk 4.0). Verified locally with cargo-ndk 4.1.2.

PR #836, opened 2026-04-28
Run stale-reservation cleanup before the active-swap gate

roeierez:
Postgres tree store skips set_leaves when an active Swap reservation exists, but the per-call cleanup of expired reservations runs only after that early-return. A swap reservation that crashed before finalize/cancel keeps has_active_swap true forever, blocks its own cleanup, and freezes the local leaf/output set. Operator-side balance updates are never applied.

Tree store: move cleanup_stale_reservations to before the swap-gate check (Rust + JS Postgres); drop the now-redundant second call.

Token store: introduce cleanup_stale_reservations (matching the tree store's RESERVATION_TIMEOUT_SECS = 300s) and call it in the same spot (Rust + JS Postgres). The schema already carries created_at on token_reservations, so no migration is needed.

Add Rust regression tests for both stores: a backdated Swap reservation must be cleaned up and the refresh must apply.


danielgranhao (approved):
LGTM

PR #838, opened 2026-04-29
Sync: Reconcile pending payments older than 1 minute

hydra-yse:
Closes #777

This PR fixes a scenario in which the optimized sync process skips over payments which remain pending between syncs, causing them to fall behind the current tip and never be updated again.

### DIFF
--- a/crates/breez-sdk/core/src/sync.rs
+++ b/crates/breez-sdk/core/src/sync.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashMap, str::FromStr, sync::Arc};
+use std::{str::FromStr, sync::Arc};
 
 use spark_wallet::{
     ListTokenTransactionsRequest, ListTransfersRequest, Order, PagingFilter, SparkWallet,
@@ -171,19 +171,10 @@ impl SparkSyncService {
             }
         };
 
-        // Build a map of payment ID → status. Payment IDs for bitcoin/lightning payments are transfer UUIDs;
-        // token payment IDs use a different format. Parse each to filter for transfers only.
-        let (transfer_ids, pending_by_id): (Vec<TransferId>, HashMap<String, PaymentStatus>) =
-            pending_payments
-                .iter()
-                .fold((Vec::new(), HashMap::new()), |(mut ids, mut map), p| {
-                    if let Ok(transfer_id) = TransferId::from_str(&p.id) {
-                        ids.push(transfer_id);
-                        map.insert(p.id.clone(), p.status);
-                    }
-                    (ids, map)
-                });
-
+        let transfer_ids: Vec<TransferId> = pending_payments
+            .iter()
+            .filter_map(|p| TransferId::from_str(&p.id).ok())
+            .collect();
         if transfer_ids.is_empty() {
             return;
         }
@@ -216,16 +207,13 @@ impl SparkSyncService {
                 }
             };
 
-            // Only act if the status has changed from what we have stored
-            let maybe_existing_status = pending_by_id.get(&payment.id).copied();
-
-            if !maybe_existing_status.is_some_and(|s| s != payment.status) {
+            if payment.status == PaymentStatus::Pending {
                 continue;
             }
 
             info!(
-                "Reconciliation: payment {} status changed from {:?} to {:?}",
-                payment.id, maybe_existing_status, payment.status
+                "Reconciliation: payment {} status changed from Pending to {:?}",
+                payment.id, payment.status
             );
 
             if let Err(e) = self.storage.insert_payment(payment.clone()).await {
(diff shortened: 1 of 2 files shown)


crates/breez-sdk/core/src/sync.rs
```diff
@@ -141,9 +142,104 @@ impl SparkSyncService {
...
+            return;
+        }
+
+        info!(
+            "Reconciling {} locally-pending bitcoin payments against server",
+            transfer_ids.len()
+        );
+        let transfers_response = match self
+            .spark_wallet
+            .list_transfers(ListTransfersRequest {
```
roeierez, line 188: I think this is fine. I recall in the past we also needed to query the ssp but I think now we don't need to.  @danielgranhao could you please validate my assumption here?

crates/breez-sdk/core/src/sync.rs
```diff
@@ -141,9 +142,104 @@ impl SparkSyncService {
             next_filter = transfers_response.next;
         }
 
+        // Re-check all locally-stored pending payments to catch status transitions
+        // that occurred before our current offset window (e.g. pending → failed).
+        self.reconcile_pending_payments(initial_sync_complete).await;
```
roeierez, line 147: Perhaps exclude these that were just inserted by this sync? Otherwise we will reconsile for every coming pending payment.

crates/breez-sdk/core/src/sync.rs
```diff
@@ -141,9 +142,104 @@ impl SparkSyncService {
...
+
         Ok(())
     }
 
+    /// Re-fetches all locally-stored pending payments from the server and updates
+    /// any whose status has changed. This catches cases where a payment transitioned to
+    /// failed/completed before the current sync offset window began.
+    /// Skip payments younger than 1 minute to avoid unnecessary server load.
+    async fn reconcile_pending_payments(&self, initial_sync_complete: bool) {
```
roeierez, line 156: let's add a test for the new functionality. An integration test would be nice.

roeierez (changes requested):
Looks good

crates/breez-sdk/core/src/sync.rs
roeierez: I just saw the timestamp usage bellow which should cover my feedback above, resolving this.

crates/breez-sdk/core/src/sync.rs
danielgranhao: Yes, internally `list_transfers` queries the SSP.

crates/breez-sdk/core/src/sync.rs
```diff
@@ -141,9 +142,104 @@ impl SparkSyncService {
...
+        let (transfer_ids, pending_by_id): (Vec<TransferId>, HashMap<String, PaymentStatus>) =
+            pending_payments
+                .iter()
+                .fold((Vec::new(), HashMap::new()), |(mut ids, mut map), p| {
+                    if let Ok(transfer_id) = TransferId::from_str(&p.id) {
+                        ids.push(transfer_id);
+                        map.insert(p.id.clone(), p.status);
+                    }
+                    (ids, map)
+                });
```
danielgranhao: Not sure I follow. Aren't all these payments pending? Why do we need the map?

crates/breez-sdk/core/src/sync.rs
```diff
@@ -141,9 +142,104 @@ impl SparkSyncService {
...
+                    continue;
+                }
+            };
+
+            // Only act if the status has changed from what we have stored
+            let maybe_existing_status = pending_by_id.get(&payment.id).copied();
+
+            if !maybe_existing_status.is_some_and(|s| s != payment.status) {
+                continue;
+            }
```
danielgranhao: I think we can get rid of map and replace this with:

```
if payment.status == PaymentStatus::Pending {              
    continue;                                                                                                                                                                                                                                                          
}
```

danielgranhao (commented):
Looks good. Just the reconciliation logic can be simplified IMO

One concern about this strategy is that we have an unbounded refetch of all pending payments on every sync. It's fine for most cases, but it could become an issue in some high-volume server-side applications. I wouldn't block merging, but might be worth checking with Lightspark if there's an alternative: something like support ordering transfers query by an `updated_at` timestamp?

crates/breez-sdk/core/src/sync.rs
hydra-yse: My bad, that was an artifact from the previous iteration. We were doing a db call per transfer (on each reconciliation attempt), which I thought was excessive. Doesn't make sense to check the status there now. Fixed in e25333755ab6

crates/breez-sdk/core/src/sync.rs
hydra-yse: Done in e25333755ab6 🙏

crates/breez-sdk/core/src/sync.rs
hydra-yse: Done f74369554693

danielgranhao (approved):
LGTM

roeierez (approved):
LGTM

PR #841, opened 2026-04-30
Reduce advisory locks and optimize balance query

roeierez:
Advisory locks are only needed to reserve leaves not to finalize/cancel reservations as these will not ever race with the reservation itself so this PR fixes the over conservative approach and leave it only there.
Also balance calculation now runs directly on the store level and doesn't require fetching all leaves which maybe a lot in a production environment.


roeierez:
replace by https://github.com/breez/spark-sdk/pull/845

PR #843, opened 2026-05-04
SDK: Add fallback address support for LN invoices

hydra-yse:
This PR adds fallback (previously named MRH) support to BOLT11 invoices.
On the send side, it allows the SDK to detect either MRH (legacy) or fallback addresses/spark invoices embedded in a lightning invoice, allowing the SDK to pay directly via Spark.
On the receive side, it adds a `FallbackMethod` to the receive payment request, such that users may be able to manually request whether to fallback to an address or embed a Spark invoice. It also changes the sync logic so that we're able to pick up on transfers related to the invoice and settle both payments.

## TO-DOs
- [ ] Fix `spark_invoice` not being populated on `query_all_transfers` - currently we're forwarding all the necessary information to the SSP to know this is a fallback payment, yet the field is never populated. We should resolve this internally
- [ ] Add docs


danielgranhao:
Closing due to staleness. Opened issue #1011

PR #844, opened 2026-05-04
Sync: Add timestamp-based syncing

hydra-yse:
Adds timestamp-based syncing based on payment creation time. By always grabbing the minimum creation time of a pending payment (or the last timestamp, if there are no pending payments), we effectively know the true offset to sync from each time, with no need for reconciliation

**Note:** Needs to be changed to update time. Waiting on Spark.


roeierez:
@hydra-yse ping here. Are spark necessary changes in place?

PR #845, opened 2026-05-04
Performance improvements

roeierez:
**Performance fixes for high-leaf-count wallets:**

1. Slim leaf reservation (Rust + JS Postgres tree store): SELECT only (id, value) for candidates, resolve full TreeNode data only for picked leaves; value <= max_target pre-filter cuts the candidate set from ~2,047 to ~2 for small payments. Selection logic stays single-source-of-truth via a new LeafLike trait. Bench: +43% Rust, +57% JS payments/min.

2. Refresh gating: `select_leaves_with_retry` now triggers `refresh_leaves` at most once per SparkWallet instance (gated by OnceCell); steady-state staleness is handled by the periodic + post-payment sync. Stops the pressure where every concurrent reservation failure re-triggered full operator pagination.

3.  Operator-failure tolerance: refresh_leaves no longer aborts when one operator query fails — flags those leaves is_missing_from_operators for that operator's view (preserving the multi-operator safety guarantee) and continues.

4. Balance query: fewer advisory locks + slimmer read path; mirrored in the JS Postgres store.

### DIFF
--- a/crates/breez-sdk/wasm/js/postgres-token-store/index.cjs
+++ b/crates/breez-sdk/wasm/js/postgres-token-store/index.cjs
@@ -75,7 +75,9 @@ class PostgresTokenStore {
   }
 
   /**
-   * Run a function inside a transaction with the advisory lock.
+   * Run a function inside a transaction with the advisory lock. Reserved for
+   * operations whose correctness depends on serializing the available-output
+   * set (`reserveTokenOutputs`, `setTokensOutputs`).
    * @param {function(import('pg').PoolClient): Promise<T>} fn
    * @returns {Promise<T>}
    * @template T
@@ -85,6 +87,30 @@ class PostgresTokenStore {
     try {
       await client.query("BEGIN");
       await client.query(`SELECT pg_advisory_xact_lock(${TOKEN_STORE_WRITE_LOCK_KEY})`);
+      const result = await fn(client);
+      await client.query("COMMIT");
+      return result;
+    } catch (error) {
+      await client.query("ROLLBACK").catch(() => {});
+      throw error;
+    } finally {
+      client.release();
+    }
+  }
+
+  /**
+   * Run a function inside a transaction without the advisory lock. Used by
+   * operations scoped to a single reservation_id (`cancelReservation`,
+   * `finalizeReservation`) where MVCC + row-level locks suffice and the global
+   * lock would only add contention.
+   * @param {function(import('pg').PoolClient): Promise<T>} fn
+   * @returns {Promise<T>}
+   * @template T
+   */
+  async _withTransaction(fn) {
+    const client = await this.pool.connect();
+    try {
+      await client.query("BEGIN");
       const result = await fn(client);
       await client.query("COMMIT");
       return result;
@@ -266,6 +292,58 @@ class PostgresTokenStore {
    * List all token outputs grouped by status.
    * @returns {Promise<Array<{metadata: Object, available: Array, reservedForPayment: Array, reservedForSwap: Array}>>}
    */
+  /**
+   * Returns the spendable per-token balances aggregated server-side.
+   * Each entry includes full token metadata + the available + swap-reserved sum.
+   * Tokens with zero spendable balance are filtered out by the HAVING clause.
+   * @returns {Promise<Array<{metadata: object, balance: string}>>}
+   */
+  async getTokenBalances() {
+    try {
+      const result = await this.pool.query(`
+        SELECT m.identifier, m.issuer_public_key, m.name, m.ticker, m.decimals,
+               m.max_supply, m.is_freezable, m.creation_entity_public_key,
+               COALESCE(SUM(
+                 CASE
+                   WHEN o.reservation_id IS NULL THEN o.token_amount::numeric
+                   WHEN r.purpose = 'Swap' THEN o.token_amount::numeric
+                   ELSE 0
+                 END
+               ), 0)::text AS balance
+        FROM token_metadata m
+        JOIN token_outputs o ON o.token_identifier = m.identifier
+        LEFT JOIN token_reservations r ON o.reservation_id = r.id
+        GROUP BY m.identifier, m.issuer_public_key, m.name, m.ticker,
+                 m.decimals, m.max_supply, m.is_freezable, m.creation_entity_public_key
+        HAVING COALESCE(SUM(
+                 CASE
+                   WHEN o.reservation_id IS NULL THEN o.token_amount::numeric
+                   WHEN r.purpose = 'Swap' THEN o.token_amount::numeric
+                   ELSE 0
+                 END
+               ), 0) > 0
+      `);
+      return result.rows.map((row) => ({
+        metadata: {
+          identifier: row.identifier,
+          issuerPublicKey: row.issuer_public_key,
+          name: row.name,
+          ticker: row.ticker,
+          decimals: row.decimals,
+          maxSupply: row.max_supply,
+          isFreezable: row.is_freezable,
+          creationEntityPublicKey: row.creation_entity_public_key,
+        },
+        balance: row.balance,
+      }));
+    } catch (error) {
+      throw new TokenStoreError(
+        `Failed to get token balances: ${error.message}`,
+        error
+      );
+    }
+  }
+
   async listTokensOutputs() {
     try {
       const result = await this.pool.query(
@@ -614,7 +692,7 @@ class PostgresTokenStore {
    */
   async cancelReservation(id) {
     try {
-      await this._withWriteTransaction(async (client) => {
+      await this._withTransaction(async (client) => {
         // Clear reservation_id from outputs
         await client.query(
           "UPDATE token_outputs SET reservation_id = NULL WHERE reservation_id = $1",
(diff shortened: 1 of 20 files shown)


crates/breez-sdk/wasm/js/postgres-tree-store/index.cjs
```diff
@@ -432,6 +522,34 @@ class PostgresTreeStore {
...
+  _maxTargetForPrefilter(targetAmounts) {
+    if (!targetAmounts) return Number.MAX_SAFE_INTEGER;
+    if (targetAmounts.type === "amountAndFee") {
+      return Math.max(targetAmounts.amountSats, targetAmounts.feeSats || 0);
+    }
+    if (targetAmounts.type === "exactDenominations") {
+      return targetAmounts.denominations.reduce((m, v) => Math.max(m, v), 0);
+    }
+    return Number.MAX_SAFE_INTEGER;
+  }
```
danielgranhao: Shouldn't the max target for `amountAndFee` and `exactDenominations` be the sum instead of max of values?

Thinking of following case:
- There are 4 1024 sat leaves
- Request: AmountAndFee { amount=1000, fee=500 } → total target 1500
- max_target = max(1000, 500) = 1000
- Slim set: leaves ≤1000 = none, plus smallest >1000 = [1024] (just one)
- Exact match: 1024 > 1000 → fail
- Minimum amount: 1024 < 1500 → returns None
- But the true total available query sees 4096 ≥ 1500 → returns WaitForPending (or eventually surfaces as InsufficientFunds) even though the wallet trivially has the funds and would have succeeded pre‑PR

Maybe worth adding tests to store test suite that cover this kind of scenario?

crates/breez-sdk/wasm/src/tree_store/mod.rs
```diff
@@ -378,6 +440,7 @@ type ReserveResult =
 export interface TreeStore {
     addLeaves: (leaves: TreeNode[]) => Promise<void>;
     getLeaves: () => Promise<Leaves>;
+    getAvailableBalance: () => Promise<bigint | number>;
```
danielgranhao: Do we need to support both bigint and number? Seems like if we pinned to bigint we could simplify boudary logic above, including getting rid of `as_f64()` path

crates/breez-sdk/wasm/src/token_store/mod.rs
```diff
@@ -552,6 +610,7 @@ type WasmReservationTarget =
 export interface TokenStore {
     setTokensOutputs: (tokenOutputs: WasmTokenOutputs[], refreshStartedAtMs: number) => Promise<void>;
     listTokensOutputs: () => Promise<WasmTokenOutputsPerStatus[]>;
+    getTokenBalances: () => Promise<WasmTokenBalance[]>;
```
danielgranhao, line 618: We also need to add the `WasmTokenBalance` type above

crates/spark-wallet/src/wallet.rs
```diff
@@ -1663,8 +1673,28 @@ impl SparkWallet {
...
+            // payment spike sees the first caller drive the refresh while
+            // the rest suspend on the cell. Once it completes, all later
+            // callers short-circuit with a single atomic load — no spurious
+            // InsufficientFunds bouncing during the very first refresh.
+            self.select_leaves_refresh
+                .get_or_init(|| async {
+                    info!("First select-leaves failure: refreshing leaves once");
+                    if let Err(e) = self.tree_service.refresh_leaves().await {
+                        warn!("Initial refresh_leaves failed (will rely on periodic sync): {e:?}");
+                    }
```
danielgranhao, line 1695: Not critical, but why not init the cell only if the refresh succeeded?

crates/breez-sdk/wasm/js/postgres-token-store/index.cjs
```diff
@@ -266,6 +292,58 @@ class PostgresTokenStore {
    * List all token outputs grouped by status.
    * @returns {Promise<Array<{metadata: Object, available: Array, reservedForPayment: Array, reservedForSwap: Array}>>}
    */
+  /**
+   * Returns the spendable per-token balances aggregated server-side.
+   * Each entry includes full token metadata + the available + swap-reserved sum.
+   * Tokens with zero spendable balance are filtered out by the HAVING clause.
```
danielgranhao: IIUC this diverges from the in-memory default implementation, which still returns tokens with 0 spendable balance if there is e.g. reserved balance. I don't think it currently results in any practical issues but just bringing it up

crates/spark/src/tree/service.rs
```diff
@@ -179,26 +179,44 @@ impl TreeService for SynchronousTreeService {
...
-                Ok(leaves) => operator_leaves_vec.push(leaves),
+                Ok(leaves) => successful_operators.push((id, leaves)),
                 Err(e) => {
-                    error!("Failed to query operator {id}: {e:?}");
-                    return Err(e);
+                    error!("Failed to query operator {id}, treating its leaves as missing: {e:?}");
+                    failed_operator_ids.push(id);
                 }
             }
         }
```
danielgranhao: Maybe I'm missing something, but won't a single operator failure now result in all leaves being marked `is_missing_from_operators` and thus unavailable?

If that is the case, does it help? Wouldn't it be preferable to keep the old behavior, where we fail the refresh and keep the old view of leaves, where at least some spending ability would be preserved while operators are flaky? Or, alternatively, just log the failure but still treat leaves as available (downside is that it could mask a genuine operator disagreement, but that would also be resolved on next refresh)

Also, seems like the oncecell refresh gating solution already covers at least part of the motivation to introduce this change

crates/spark-postgres/src/tree_store.rs
```diff
@@ -296,9 +329,6 @@ impl TreeStore for PostgresTreeStore {
         let mut client = self.pool.get().await.map_err(map_err)?;
         let tx = client.transaction().await.map_err(map_err)?;
 
-        // Acquire advisory lock to prevent deadlocks with concurrent operations
-        Self::acquire_write_lock(&tx).await?;
-
```
danielgranhao, line 301: I think removing the lock on finalize can result in a race with `set_leaves` that ends up surfacing spent leaves as available (next refresh would then fix issue).

`set_leaves` filters out spent leaves in 2 steps inside one transaction:
1. read tree_spent_leaves into a HashSet (spent_ids)
2. upsert the operator-reported leaves into tree_leaves, skipping IDs in that set

`finalize_reservation` writes a new spent marker for the leaves it just consumed.

With the lock removed from finalize_reservation, finalize can commit between steps 1 and 2 of an in-flight set_leaves. Its new spent marker isn't in the already-captured spent_ids snapshot, so step 2 happily writes the just-spent leaf back into tree_leaves as available.

Same thing on token store.

danielgranhao (changes requested):
Some great improvements! Besides the few issues I found, I think it's worth reviewing the added logging as it may be too noisy (maybe worth reducing log level and checking if it's necessary to log entire payment payload as done in `events.rs` )

crates/spark-wallet/src/wallet.rs
roeierez: Even before the refresh succeed there is time where multiple payments attempt to use it which was the original issue and created unreasonable load on the server.

crates/breez-sdk/wasm/js/postgres-tree-store/index.cjs
roeierez: Good catch! fixed.

crates/breez-sdk/wasm/src/tree_store/mod.rs
roeierez: Correct, fixed.

crates/breez-sdk/wasm/src/token_store/mod.rs
roeierez: Added

crates/breez-sdk/wasm/js/postgres-token-store/index.cjs
roeierez: Aligned the behaviors.

crates/spark/src/tree/service.rs
roeierez: That's a good catch. This came up (wrongly) by mixing two issues together in production and lead to the wrong decision (user removed manually all leaves). I reverted that.

crates/spark-postgres/src/tree_store.rs
roeierez: Indeed a race. Although I think we can optimize the queries a bit more and don't use the lock I decided to add it. I also ran the bench suite on both before/after the fix and no additional overhead so we simplest  and safest is to bring it back.

crates/spark-wallet/src/wallet.rs
```diff
@@ -183,6 +183,15 @@ pub struct SparkWallet {
...
     leaf_optimizer: Arc<LeafOptimizer>,
+    /// One-shot, single-flight guard for `select_leaves_with_retry`'s call to
+    /// `refresh_leaves`. The cell's `get_or_init` blocks concurrent callers
+    /// during the in-flight refresh and short-circuits with a single atomic
+    /// load afterwards, so a startup payment burst triggers at most one
+    /// refresh and steady-state callers pay no synchronization cost. The
+    /// closure swallows refresh errors after logging, since "once per
+    /// lifetime" is what we want here regardless of outcome — subsequent
+    /// staleness is handled by the periodic + post-payment sync.
+    select_leaves_refresh: Arc<tokio::sync::OnceCell<()>>,
```
danielgranhao: nit: seems like Arc is redundant here

danielgranhao (approved):

PR #846, opened 2026-05-04
Handle token tx SO events

### LINKED ISSUE #826: Detect incoming token payments in real time
Currently, incoming token payments are only detected on sync. The SO's have now introduced a `TokenTransactionEvent` that we can use to detect and surface incoming payments sooner.

danielgranhao:
Closes #826

### DIFF
--- a/crates/breez-sdk/wasm/js/mysql-token-store/index.cjs
+++ b/crates/breez-sdk/wasm/js/mysql-token-store/index.cjs
@@ -566,6 +566,37 @@ class MysqlTokenStore {
     }
   }
 
+  async removeTokenOutputs(prevTxRefs) {
+    if (!prevTxRefs || prevTxRefs.length === 0) return;
+    try {
+      await this._withTransaction(async (conn) => {
+        for (const [txHash, vout] of prevTxRefs) {
+          const [rows] = await conn.query(
+            "SELECT output_id FROM token_outputs WHERE user_id = ? AND prev_tx_hash = ? AND prev_tx_vout = ?",
+            [this.identity, txHash, vout]
+          );
+          if (rows.length > 0) {
+            const outputId = rows[0].output_id;
+            await conn.query(
+              "DELETE FROM token_outputs WHERE user_id = ? AND output_id = ?",
+              [this.identity, outputId]
+            );
+            await conn.query(
+              "INSERT IGNORE INTO token_spent_outputs (user_id, output_id, spent_at) VALUES (?, ?, NOW())",
+              [this.identity, outputId]
+            );
+          }
+        }
+      });
+    } catch (error) {
+      if (error instanceof TokenStoreError) throw error;
+      throw new TokenStoreError(
+        `Failed to remove token outputs: ${error.message}`,
+        error
+      );
+    }
+  }
+
   async reserveTokenOutputs(
     tokenIdentifier,
     target,
--- a/crates/breez-sdk/wasm/js/mysql-token-store/index.cjs
+++ b/crates/breez-sdk/wasm/js/mysql-token-store/index.cjs
@@ -548,13 +548,13 @@ class MysqlTokenStore {
         if (outputsToRemove && outputsToRemove.length > 0) {
           for (const [txHash, vout] of outputsToRemove) {
             const [rows] = await conn.query(
-              "SELECT output_id FROM brz_token_outputs WHERE user_id = ? AND prev_tx_hash = ? AND prev_tx_vout = ?",
+              "SELECT id FROM brz_token_outputs WHERE user_id = ? AND prev_tx_hash = ? AND prev_tx_vout = ?",
               [this.identity, txHash, vout]
             );
             if (rows.length > 0) {
-              const outputId = rows[0].output_id;
+              const outputId = rows[0].id;
               await conn.query(
-                "DELETE FROM brz_token_outputs WHERE user_id = ? AND output_id = ?",
+                "DELETE FROM brz_token_outputs WHERE user_id = ? AND id = ?",
                 [this.identity, outputId]
               );
               await conn.query(
--- a/crates/breez-sdk/wasm/js/mysql-token-store/index.cjs
+++ b/crates/breez-sdk/wasm/js/mysql-token-store/index.cjs
@@ -535,55 +535,54 @@ class MysqlTokenStore {
     }
   }
 
-  async insertTokenOutputs(tokenOutputs) {
+  /**
+   * Atomically remove spent outputs and insert new outputs.
+   * @param {Array<[string, number]>} outputsToRemove - Array of [prevTxHash, prevTxVout] tuples
+   * @param {Object|null} outputsToAdd - Token outputs to insert (with metadata)
+   * @returns {Promise<void>}
+   */
+  async updateTokenOutputs(outputsToRemove, outputsToAdd) {
     try {
       await this._withTransaction(async (conn) => {
-        await this._upsertMetadata(conn, tokenOutputs.metadata);
-
-        const outputIds = tokenOutputs.outputs.map((o) => o.output.id);
-        if (outputIds.length > 0) {
-          const placeholders = buildPlaceholders(outputIds.length);
-          await conn.query(
-            `DELETE FROM brz_token_spent_outputs WHERE user_id = ? AND output_id IN (${placeholders})`,
-            [this.identity, ...outputIds]
-          );
-        }
-
-        for (const output of tokenOutputs.outputs) {
-          await this._insertSingleOutput(
-            conn,
-            tokenOutputs.metadata.identifier,
-            output
-          );
-        }
-      });
-    } catch (error) {
-      if (error instanceof TokenStoreError) throw error;
-      throw new TokenStoreError(
-        `Failed to insert token outputs: ${error.message}`,
-        error
-      );
-    }
-  }
-
-  async removeTokenOutputs(prevTxRefs) {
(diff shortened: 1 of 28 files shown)


roeierez:
@hydra-yse  should we review this? Asking because it is still in draft but Spark team have done their part.

hydra-yse (approved):
Looks good, added an itest and a call to refresh the token outputs in the spark wallet handler, so that the balance can update without syncing.

hydra-yse:
**Edit:** To improve performance, I've removed the `refresh_token_outputs` call in favor of manual insertion of the token outputs when we receive the event in the spark crate.

crates/breez-sdk/core/src/sdk/sync.rs
```diff
@@ -185,12 +189,117 @@ impl BreezSdk {
...
                     .await;
             }
+            WalletEvent::TokenTransaction(transaction) => {
+                info!("Token transaction event: {}", transaction.hash);
+                if let Err(e) = self.process_token_transaction_event(transaction).await {
+                    error!("Failed to process token transaction event: {e:?}");
+                }
+                self.sync_coordinator
+                    .trigger_sync_no_wait(super::SyncType::WalletState, true)
+                    .await;
```
dangeross: Do we need to do this if we've already processed the transaction?

hydra-yse: Done fb26c273be08

crates/breez-sdk/core/src/sdk/sync.rs
```diff
@@ -239,6 +241,11 @@ impl BreezSdk {
                 .map(|p| p.status);
             let status_changed = existing_status.is_none_or(|s| s != payment.status);
 
+            if let Err(e) = sync_service.apply_payment_metadata(&payment).await {
+                error!("Failed to apply payment metadata for token payment: {e:?}");
+                continue;
```
dangeross: Are we skipping over inserting the payment here so we retry on periodic sync?

hydra-yse: Yes, it will get picked up by sync. No strong opinion on whether we should skip if the call fails, the metadata can be appended later by the sync, too. @danielgranhao wydt?

Perhaps since metadata is non-critical we can try inserting anyway.

dangeross: I think the sync will just do the same again anyway, logging the error and continuing to insert the payment

hydra-yse: Done 0a922beb3b4

crates/spark-wallet/src/wallet.rs
```diff
@@ -2186,6 +2189,55 @@ impl BackgroundProcessor {
...
+                .ok_or_else(|| {
+                    SparkWalletError::Generic(format!("Metadata not found for token {token_id}"))
+                })?;
+            self.token_service
+                .insert_token_outputs(&TokenOutputs {
+                    metadata,
+                    outputs: our_outputs,
+                })
+                .await?;
+        }
```
danielgranhao: We should also remove the inputs spent by this tx from our local store. Otherwise, instances where a payment with change wasn't initiated locally will temporarily display a higher-than-actual balance.

roeierez: @hydra-yse let's also add a test for this case.

hydra-yse: Done in 5e9d64f

crates/spark/src/token/store/tests.rs
```diff
@@ -440,3 +440,15 @@ async fn test_insert_outputs_clears_spent_status() {
...
+#[async_test_all]
+async fn test_remove_token_outputs_by_prev_tx_ref() {
+    let store = InMemoryTokenOutputStore::default();
+    shared_tests::test_remove_token_outputs_by_prev_tx_ref(&store).await;
+}
+
+#[async_test_all]
+async fn test_remove_token_outputs_prevents_refresh_readd() {
+    let store = InMemoryTokenOutputStore::default();
+    shared_tests::test_remove_token_outputs_prevents_refresh_readd(&store).await;
```
danielgranhao, line 453: We are missing wiring these tests for postgres and mysql implementations

danielgranhao (commented):
Other than this, LGTM

crates/spark/src/token/store/tests.rs
hydra-yse: Done 305d88c

crates/spark/src/token/token_service.rs
```diff
@@ -141,6 +141,69 @@ impl TokenService {
...
+        // 1. Remove spent inputs so the balance drops immediately.
+        if let TokenInputs::Transfer(ref transfer_input) = transaction.inputs {
+            let spent_refs: Vec<(String, u32)> = transfer_input
+                .outputs_to_spend
+                .iter()
+                .map(|o| (o.prev_token_tx_hash.clone(), o.prev_token_tx_vout))
+                .collect();
+            if !spent_refs.is_empty() {
+                self.token_output_service
+                    .remove_token_outputs(&spent_refs)
```
roeierez: Isn't it better to use the current method `insert_token_outputs`? Perhaps rename it to `update_token_outputs` and pass it both `to_remove` and `to_add` ?
This basically will allow the whole update to be atomic and minimize the API surface.

roeierez (changes requested):

crates/breez-sdk/core/src/sdk/sync.rs
```diff
@@ -185,12 +189,123 @@ impl BreezSdk {
...
+        // query the new balance.
+        if let Err(e) = update_balances(self.spark_wallet.clone(), self.storage.clone()).await {
+            error!("Failed to update balances before token transaction event: {e:?}");
+        }
+
+        let sync_service = SparkSyncService::new(
+            self.spark_wallet.clone(),
+            self.storage.clone(),
+            self.event_emitter.clone(),
+        );
```
roeierez: What is the case here? Which metadata are we adding?

roeierez (changes requested):

crates/breez-sdk/core/src/sdk/sync.rs
hydra-yse: It's to address this: https://github.com/breez/spark-sdk/pull/846#discussion_r3234891289

Part of the sync routine was to apply payment metadata (so moving from cache to storage). Since we're not syncing per notification any longer, we manually call the method.

crates/spark/src/token/token_service.rs
hydra-yse: Makes sense 👍

crates/breez-sdk/core/src/sdk/sync.rs
roeierez: Is it specifically for tokens? Because I don't see it in sats claim. cc @dangeross

crates/spark/src/token/token_service.rs
hydra-yse: Done fd051cc4a249

crates/breez-sdk/core/src/sdk/sync.rs
dangeross: It happens for both spark transfers and token transactions in the sync. But actually metadata is only cached using this method in the flashnet conversion process if there is not yet a token transaction available to determine the vout
- https://github.com/breez/spark-sdk/blob/0d45c6e1d463bcdc098fd7bc3b01705e2de87113/crates/breez-sdk/core/src/sync.rs#L87
- https://github.com/breez/spark-sdk/blob/0d45c6e1d463bcdc098fd7bc3b01705e2de87113/crates/breez-sdk/core/src/sync.rs#L387

crates/breez-sdk/wasm/js/mysql-token-store/index.cjs
```diff
@@ -535,32 +535,62 @@ class MysqlTokenStore {
...
-          const placeholders = buildPlaceholders(outputIds.length);
-          await conn.query(
-            `DELETE FROM brz_token_spent_outputs WHERE user_id = ? AND output_id IN (${placeholders})`,
-            [this.identity, ...outputIds]
-          );
+        // 1. Remove spent outputs and mark as spent.
+        if (outputsToRemove && outputsToRemove.length > 0) {
+          for (const [txHash, vout] of outputsToRemove) {
+            const [rows] = await conn.query(
+              "SELECT output_id FROM brz_token_outputs WHERE user_id = ? AND prev_tx_hash = ? AND prev_tx_vout = ?",
```
dangeross: Can we double check all uses of the `brz_token_outputs` table (postgres/mysql). I believe we should be using `id`, not `output_id`

hydra-yse: Yeah, those were exactly the cases I fixed, tests were failing before. `brz_token_outputs` is indexed by `id`, whereas `brz_token_spent_outputs` by `output_id`. I'll double check the WASM ones too.

hydra-yse: Done a5fbfe7

roeierez: @hydra-yse  Do we have a test for this? Can't rely on Ross's sharp eyes.

hydra-yse: Done b2cf9c8

roeierez (approved):

dangeross (approved):
LGTM

PR #847, opened 2026-05-05
Bindings: KMP: Add macos native bindings

hydra-yse:
This PR:
- Adds macOS x86/arm64 targets to the KMP bindings
- Improves kmp publish workflow by adding build cache
- Adds docs for setup (very similar to iOS)

Tested on macOS Sequoia 15.7 arm64


roeierez (approved):
LGTM

erdemyerebasmaz (approved):

PR #848, opened 2026-05-05
Fix Spark integration tests

hydra-yse:
Bumps spark git ref in Dockerfile to `39ecc67c0541b3c94a365e195722c30a9129600c`, which updated to frost 3.0

### DIFF
--- a/crates/spark-itest/docker/migrations.dockerfile
+++ b/crates/spark-itest/docker/migrations.dockerfile
@@ -1,4 +1,4 @@
-ARG VERSION=39ecc67c0541b3c94a365e195722c30a9129600c
+ARG VERSION=e8ceff40979d27c1c19e31899901b7b47bd591bc
 ARG REPOSITORY=https://github.com/buildonspark/spark.git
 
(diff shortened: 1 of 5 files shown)


JssDWt:
@hydra-yse Tests are failing here because it's not finding the keyshares. We have a query inside the test fixture that checks whether all operators have enough keyshares. That might need some work?

JssDWt:
I think it's worth doing this here too: https://github.com/breez/spark-sdk/blob/a9e156147ed9cc7cbd3cde5373510c85e02e96d5/crates/spark/src/services/coop_exit.rs#L401. It's not checked the same way right now, but might be in the future.

JssDWt (approved):
LGTM. 

For other reviewers: the error that led to the refund tx changes was

```
Error: Service error: service connection error: Connection error: status: InvalidArgument, message: "failed to create transfer for transfer X: unable to validate and construct bitcoin transactions: leaf X validation for transfer failed: leaf X is a zero node, zero nodes must not have a direct refund tx, transfer id: Y
```

roeierez (approved):

PR #850, opened 2026-05-05
Expose default_spark_config() for customizing SO/SSP endpoints

peterrojs:
## Summary

- Adds `default_spark_config(network) -> SparkConfig` to the Rust core, making it easier to pass in your own config.
- Re-exports it from `lib.rs` (UniFFI-exported for all bindings)
- Adds `defaultSparkConfig(network)` WASM binding in `wasm/src/sdk.rs`

## Usage

```rust
let mut spark = default_spark_config(Network::Mainnet);
spark.ssp_config.base_url = "https://staging-ssp.example.com".to_string();
let mut config = default_config(Network::Mainnet);
config.spark_config = Some(spark);
```

```typescript
const spark = defaultSparkConfig('mainnet')
spark.sspConfig.baseUrl = 'https://staging-ssp.example.com'
const config = defaultConfig('mainnet')
config.sparkConfig = spark
```

### DIFF
--- a/crates/breez-sdk/core/src/sdk/mod.rs
+++ b/crates/breez-sdk/core/src/sdk/mod.rs
@@ -200,7 +200,54 @@ pub fn default_config(network: Network) -> Config {
         },
         stable_balance_config: None,
         max_concurrent_claims: 4,
-        spark_config: None,
+        spark_config: Some(default_spark_config(network)),
+    }
+}
+
+/// Builds the default [`SparkConfig`](crate::models::SparkConfig) for the given network.
+///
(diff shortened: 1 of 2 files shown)


JssDWt (approved):
utACK

roeierez:
@peterrojs Perhaps instead of another endpoint return the default spark config inside the sdk config returned from `default_config` ?

peterrojs:
> @peterrojs Perhaps instead of another endpoint return the default spark config inside the sdk config returned from `default_config` ?

Yep, that makes more sense let me change it

peterrojs:
It should now be a bit cleaner

roeierez (approved):
LGTM

PR #851, opened 2026-05-05
Allow unsigned Spark invoices in address parsing

peterrojs:
## Summary
- Allow parsing Spark invoices without a signature, matching the behavior of Spark Signing Operators which treat signatures as optional
- When a signature is present, it is still verified as before
- This enables server-side invoice generation (e.g. on behalf of a customer) where the server has the customer's public key but not their signing key

## Context
The Spark SO `validateInvoiceFields` function only verifies signatures when present:
```go
if decoded.SparkAddress.Signature != nil {
    err := common.VerifySparkAddressSignature(...)
}
```

The SDK should follow the same contract. Without this fix, `prepare_send_payment` rejects unsigned invoices with `InvalidInput: 'invalid input'`.

### DIFF
--- a/crates/spark/src/address/mod.rs
+++ b/crates/spark/src/address/mod.rs
@@ -476,13 +476,10 @@ impl FromStr for SparkAddress {
 
         let address = SparkAddress::new(identity_public_key, network, invoice_fields);
 
-        if address.is_invoice() {
+        if address.is_invoice()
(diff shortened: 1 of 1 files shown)


JssDWt (approved):
utACK

roeierez (approved):

PR #852, opened 2026-05-05
Expose query_spark_invoices via SDK public API

peterrojs:
## Summary
- Wraps `SparkWallet::query_spark_invoices` in a new `BreezSdk::query_spark_invoices` method so language bindings can query the Spark operator for invoice status (`NotFound` / `Pending` / `Finalized` / `Returned`) and, when finalized, the underlying transfer details.
- Adds `QuerySparkInvoicesRequest/Response`, `QuerySparkInvoiceResult`, `SparkInvoiceStatus`, `SparkInvoiceTransferType` to the core models with `uniffi::Record`/`Enum` derives, and mirrors them in the WASM and Flutter interfaces per `CLAUDE.md`.
- Used for direct polling of Spark Invoices from the SO

### DIFF
--- a/crates/breez-sdk/core/src/models/mod.rs
+++ b/crates/breez-sdk/core/src/models/mod.rs
@@ -1841,3 +1841,93 @@ pub struct UnregisterWebhookRequest {
     /// The unique identifier of the webhook to unregister.
     pub webhook_id: String,
 }
+
+/// Request for querying the status of Spark invoices.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct QuerySparkInvoicesRequest {
+    /// The list of Spark invoices to query.
+    pub invoices: Vec<String>,
+}
+
+/// Result of querying a single Spark invoice.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct QuerySparkInvoiceResult {
+    /// The queried invoice string.
+    pub invoice: String,
+    /// The status of the invoice on the Spark operator.
+    pub status: SparkInvoiceStatus,
+    /// Details of the underlying transfer when the invoice has been finalized.
+    pub transfer_type: Option<SparkInvoiceTransferType>,
+}
+
+/// Status of a queried Spark invoice as reported by the Spark operator.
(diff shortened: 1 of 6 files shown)


JssDWt:
Do I understand correctly that the real issue here is that `invoice_details` is not set?

peterrojs:
Sorry, the description was a bit misleading. It's mainly for direct polling of invoices to the SO, skipping the persistence layer that `list_payments` has

JssDWt:
Can you explain a little more about your use-case? I'd like to keep the sdk interface clean of spark specifics where possible. I want to see how else we can solve it.

peterrojs:
Yeah no worries, that makes total sense. We have a flow where a Spark Invoice is issued and, once the sender pays it, we need a way to poll the SO to check whether it's been paid. This should largely be replaced by event streaming, but we'd need a fallback in case that fails. Worst case, we can implement this in our own codebase to avoid polluting the SDK

benwgold:
Hey @JssDWt-
Just explaining why this helps us here-

For our upcoming integration of Breez- clients will use Spark Invoices as a stable tracker for payment completion in and out of our wallet. Although payment succeeded events are useful, its not quite enough because there is the chance the service could experience downtime / crashes and not receive the event as expected.  Having an explicit query_spark_invoices allows services like ours to implement a fully trustable fallback mechanism for checking whether a payment in or out of the SDK was actually completed.  Making sure we don't miss a payment is critical for business logic and accounting.

Do you think there is another way the service implementing the breez sdk can be 100% sure of payment completion even taking into account potential downtime?

roeierez:
> Hey @JssDWt- Just explaining why this helps us here-
> 
> For our upcoming integration of Breez- clients will use Spark Invoices as a stable tracker for payment completion in and out of our wallet. Although payment succeeded events are useful, its not quite enough because there is the chance the service could experience downtime / crashes and not receive the event as expected. Having an explicit query_spark_invoices allows services like ours to implement a fully trustable fallback mechanism for checking whether a payment in or out of the SDK was actually completed. Making sure we don't miss a payment is critical for business logic and accounting.
> 
> Do you think there is another way the service implementing the breez sdk can be 100% sure of payment completion even taking into account potential downtime?

@benwgold thanks for explaining. We certainly need to support a reliable way of knowing if an invoice was paid.
I am curious to know if the automatic sync or manual sync method is something you considered? Or perhaps there is something that blocks you there? In our mind the polling should looks like this:
1. Sync (or wait for periodic sync event)
2. lookup for the payment in the db passing the invoice hash.

LMK what you think.

PR #853, opened 2026-05-05
Multi-tenant storage and tree/token stores

danielgranhao:
Adds multi-tenant support to all three Postgres-backed stores (storage, tree, token) so multiple SDK instances can safely share one Postgres database, scoped by the tenant's identity pubkey.

### DIFF
--- a/crates/spark-postgres/Cargo.toml
+++ b/crates/spark-postgres/Cargo.toml
@@ -11,6 +11,7 @@ workspace = true
 
 [dependencies]
 async-trait.workspace = true
+bitcoin.workspace = true
 chrono.workspace = true
 deadpool.workspace = true
 deadpool-postgres.workspace = true
@@ -19,7 +20,6 @@ macros.workspace = true
 platform-utils.workspace = true
 rustls.workspace = true
 serde.workspace = true
-sha2.workspace = true
 serde_json.workspace = true
 spark-wallet.workspace = true
 thiserror.workspace = true
--- a/crates/spark-postgres/Cargo.toml
+++ b/crates/spark-postgres/Cargo.toml
@@ -14,6 +14,7 @@ async-trait.workspace = true
 chrono.workspace = true
 deadpool.workspace = true
 deadpool-postgres.workspace = true
+hex.workspace = true
 macros.workspace = true
 platform-utils.workspace = true
 rustls.workspace = true
--- a/crates/spark-postgres/Cargo.toml
+++ b/crates/spark-postgres/Cargo.toml
@@ -19,6 +19,7 @@ macros.workspace = true
 platform-utils.workspace = true
 rustls.workspace = true
 serde.workspace = true
+sha2.workspace = true
 serde_json.workspace = true
 spark-wallet.workspace = true
 thiserror.workspace = true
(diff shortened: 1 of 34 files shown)


crates/spark-postgres/src/token_store.rs
```diff
@@ -809,31 +976,45 @@ impl PostgresTokenStore {
...
+    /// other. Same-tenant writes still serialize on the same lock.
     async fn acquire_write_lock(
+        &self,
         tx: &tokio_postgres::Transaction<'_>,
     ) -> Result<(), TokenOutputServiceError> {
         tx.execute(
-            "SELECT pg_advisory_xact_lock($1)",
-            &[&TOKEN_STORE_WRITE_LOCK_KEY],
+            "SELECT pg_advisory_xact_lock($1, $2)",
+            &[&TOKEN_STORE_LOCK_CLASSID, &self.lock_objid],
```
JssDWt: I think perhaps it might be better to use a 64 bit number derived from the pubkey here? A collision is a birthday problem, with 65k users, there's ~50% probability at least two collide. 

Ideally we would get rid of the advisory lock. It worked in the single-user scenario, but seems to get out of hand a bit here. I wonder whether we can do `SELECT … FOR UPDATE SKIP LOCKED` in try_reserve_leaves. and/or use a mutex row, like `user_lock(user_id PK)`, and do `SELECT 1 FROM user_lock WHERE user_id = $1 FOR UPDATE`

JssDWt (commented):
Looking good!

crates/spark-postgres/src/token_store.rs
roeierez: @JssDWt based on my test advisory locks here are really cheap. I tested many benches with and without and they are not a factor anymore. Payments take seconds and both reserve/unreserve takes milliseconds. Also select for update would have some consequences such as:
1. Need to implement retry
2. A slightly change in behavior where leaves would be return for the second caller that are higher than the requested amount resulting in swap (since all leaves bellow are locked by the query). 

If we delegate the select leaves algo to the db (using a slightly complicated window function) that might be a follow up step but I am really not sure it worth the complexity.

JssDWt: @roeierez So you mean we can accept the collision there? Multiple users using the same advisory lock? something like the first 8 bytes of sha256('tenant:', identity) would make a 64-bit key where the risk of collision becomes negligable. I'm not really saying we should get rid of the advisory lock, just I think locks across users shouldn't collide and looking for alternatives.

roeierez: @JssDWt I was only referring to the suggestion of moving to the "select for update".
About the collision I agree it is a good idea to make it negligable.

danielgranhao: Fixed, thanks Jesse

crates/spark-postgres/Cargo.toml
```diff
@@ -19,6 +19,7 @@ macros.workspace = true
 platform-utils.workspace = true
 rustls.workspace = true
 serde.workspace = true
+sha2.workspace = true
```
JssDWt: Let's use `bitcoin_hashes::sha256` or `bitcoin::hashes::sha256` for this, because they are already in the dependency tree.

danielgranhao: sha2 was already in the dep tree as well but I've changed to bitcoin::hashes::sha256 for consistency with rest of the codebase

roeierez (approved):
LGTM

JssDWt (approved):
LGTM

PR #856, opened 2026-05-06
Make the ConnectionManager shareable across SDK instances

JssDWt:
Sharing the ConnectionManager is an optimization that ensures there is a single grpc connection open per operator, rather than a grpc connection per operator per user. The connection manager implementation remains hidden. We just expose a wrapper type that can be shared across sdk instances.

The session manager is not shared, because sessions contain user-specific tokens. So the session manager stays per-user.


crates/breez-sdk/wasm/src/sdk_builder.rs
```diff
@@ -186,6 +187,14 @@ impl SdkBuilder {
         self
     }
 
+    #[wasm_bindgen(js_name = "withConnectionManager")]
+    pub fn with_connection_manager(mut self, connection_manager: ConnectionManager) -> Self {
+        self.builder = self
+            .builder
+            .with_connection_manager(connection_manager.inner);
+        self
+    }
```
danielgranhao: As is, this takes the manager by value and the JS reference to it becomes unusable. I've confirmed this with a small test: throws `null pointer passed to rust`

The suggestion fixes the issue.
```suggestion
    #[wasm_bindgen(js_name = "withConnectionManager")]
    pub fn with_connection_manager(mut self, connection_manager: &ConnectionManager) -> Self {
        self.builder = self
            .builder
            .with_connection_manager(connection_manager.inner.clone());
        self
    }
```

packages/flutter/rust/src/sdk_builder.rs
```diff
@@ -49,6 +49,15 @@ impl SdkBuilder {
...
     }
 
+    #[frb(sync)]
+    pub fn with_connection_manager(self, connection_manager: ConnectionManager) -> Self {
+        let builder = <breez_sdk_spark::SdkBuilder as Clone>::clone(&self.inner)
+            .with_connection_manager(connection_manager.inner);
+        Self {
+            inner: Arc::new(builder),
+        }
+    }
```
danielgranhao: Same issue here
```suggestion
    pub fn with_connection_manager(self, connection_manager: &ConnectionManager) -> Self {
        let builder = <breez_sdk_spark::SdkBuilder as Clone>::clone(&self.inner)
            .with_connection_manager(connection_manager.inner.clone());
        Self {
            inner: Arc::new(builder),
        }
    }
```

danielgranhao (changes requested):
Looks good, just an issue with Wasm and flutter

roeierez:
@JssDWt as this change is solely for high throughput backends I think we want to actually have a pool of connections per operator as I am not sure one connection will be able to handle all the traffic (streams subscriptions, etc..)
Tonic has this as builtin feature that can be wired without too much of refactoring: https://docs.rs/tonic/latest/tonic/transport/struct.Channel.html#method.balance_list

JssDWt:
> @JssDWt as this change is solely for high throughput backends I think we want to actually have a pool of connections per operator as I am not sure one connection will be able to handle all the traffic (streams subscriptions, etc..)
> Tonic has this as builtin feature that can be wired without too much of refactoring:

I'll look into that deeper

JssDWt:
@roeierez I added a balanced connection manager that adds connections based on the amount of sdk instances, but batches up to `max_tenants_per_connection` instances over 1 connection. I think it doesn't make sense to initialize the connection manager with a fixed amount of connections, so it should be dynamic. One bottleneck is MAX_CONCURRENT_STREAMS, although it seems that is unlimited in go's grpc server by default.

Some other bottlenecks may be tcp head-of-line blocking, single CPU-core writer-task, HTTP/2 flow control, load-balancer stickiness. These are reasons to increase the amount of connections.

Reasons to decrease the amount of connections are File descriptors and TCP handshakes.

Altogether I think it's a per-server reason to pool connections or not. With the current setup you can do all options:
- connection per sdk instance (do not pass connection manager to builder)
- single connection across all instances (pass connection manager with None for `max_tenants_per_connection`)
- connection per n instances

For browser and node.js this doesn't help. I added some docs to tune node.js if needed.

JssDWt:
@roeierez Replaced the dynamic connection manager with a static one.

crates/breez-sdk/breez-itest/tests/connection_manager.rs
danielgranhao, line 1: nit: Seems the 2 tests differ only in the connection manager argument. Maybe worth a ` #[rstest] #[case(None)] #[case(Some(2))]` parameterization?

danielgranhao (approved):
LGTM

roeierez (approved):

PR #860, opened 2026-05-07
feat(spark-wallet): split token output consolidation into N outputs

peterrojs:
Add `target_output_count` to `TokenOutputsOptimizationOptions` so token output auto-consolidation produces N roughly-equal outputs instead of collapsing into a single one. Default is 5.

When auto-consolidation collapses everything into one output, subsequent payments must serialize on that single output until new ones accumulate. That's bad for server-side wallets doing parallel sends. Splitting into N outputs preserves concurrency.

In `optimize_token_outputs`, for each token over threshold:
- reserve up to MAX_TOKEN_TX_INPUTS smallest outputs (unchanged)
- emit `(N - 1)` explicit `TransferTokenOutput`s of `amount / N` to the wallet's own spark address
- let `build_transfer_token_transaction` auto-add the remainder as the change output (also self-addressed) — net result is exactly N outputs
- if `amount < N`, fall back to a single output and warn

Internal "needed too many outputs" retry path passes `target=1` explicitly — that is emergency consolidation during payment retries where collapsing aggressively is the right move.

Validation: `target_output_count >= 1` and `target_output_count < min_outputs_threshold`.

spark-wallet validation tests covering the new fields are included.

### DIFF
--- a/crates/breez-sdk/core/src/models/mod.rs
+++ b/crates/breez-sdk/core/src/models/mod.rs
@@ -637,6 +637,21 @@ pub struct OptimizationConfig {
     ///
     /// Default value is 1.
     pub multiplicity: u8,
+    /// Number of token outputs to produce when token-output auto-consolidation
+    /// fires.
+    ///
+    /// Auto-consolidation triggers per-token when the wallet has more than the
+    /// configured threshold of available outputs for that token; instead of
+    /// collapsing them into a single output (which serializes subsequent
+    /// payments), the SDK splits the consolidated balance across this many
+    /// outputs of roughly equal value. Higher values preserve concurrency for
+    /// parallel sends at the cost of a slightly larger output set.
+    ///
+    /// Must be >= 1 and strictly less than the underlying
+    /// `min_outputs_threshold` (default 50).
+    ///
+    /// Default value is 5.
+    pub token_target_output_count: u32,
 }
 
 /// A stable token that can be used for automatic balance conversion.
--- a/crates/spark/src/token/token_service.rs
+++ b/crates/spark/src/token/token_service.rs
@@ -50,6 +50,52 @@ fn is_transaction_preempted_error(error: &OperatorRpcError) -> bool {
     matches!(error, OperatorRpcError::Connection(status) if status.code() == tonic::Code::Aborted)
 }
 
+/// Build the list of `TransferTokenOutput` entries that consolidation should send
+/// to the wallet's own spark address.
+///
+/// The summed input `amount` is split into `target_output_count` outputs of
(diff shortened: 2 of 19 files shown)


crates/spark/src/token/token_service.rs
```diff
@@ -50,6 +50,52 @@ fn is_transaction_preempted_error(error: &OperatorRpcError) -> bool {
...
+/// Build the list of `TransferTokenOutput` entries that consolidation should send
+/// to the wallet's own spark address.
+///
+/// The summed input `amount` is split into `target_output_count` outputs of
+/// `amount / target_output_count`. The transfer builder auto-adds any remainder
+/// as a change output to the same address, so the resulting on-chain output
+/// count is exactly `target_output_count`. If the per-output amount would round
+/// down to zero (i.e., `amount < target_output_count`), we fall back to a
+/// single output of `amount` to avoid creating zero-value outputs.
+fn build_consolidation_outputs(
```
dangeross, line 62: Is it worth adding tests for this fn?

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -637,6 +637,21 @@ pub struct OptimizationConfig {
...
+    /// collapsing them into a single output (which serializes subsequent
+    /// payments), the SDK splits the consolidated balance across this many
+    /// outputs of roughly equal value. Higher values preserve concurrency for
+    /// parallel sends at the cost of a slightly larger output set.
+    ///
+    /// Must be >= 1 and strictly less than the underlying
+    /// `min_outputs_threshold` (default 50).
+    ///
+    /// Default value is 5.
+    pub token_target_output_count: u32,
```
dangeross, line 654: Could you also add a short description to the [config](https://github.com/breez/spark-sdk/blob/main/docs/breez-sdk/src/guide/config.md#optimization-configuration) mkdocs about the added config option?

dangeross (commented):
Looking good

crates/spark/src/token/token_service.rs
peterrojs: yep, makes sense

crates/breez-sdk/core/src/models/mod.rs
peterrojs: done!

dangeross (approved):
LGTM

danielgranhao (approved):
Looks good! Just a heads up, on a separate PR we'll be rearranging the optimization config a bit on the SDK level because as is it mixes leaf only settings (like auto enabled) with this new token one. It will probably end up becoming separate config structs like we have on the spark wallet level, which also paves the way for more token settings to be exposed.

PR #861, opened 2026-05-07
feat: implement token transaction v3 broadcast flow

peterrojs:
## Summary
- Implement the V3 token transaction broadcast flow in `crates/spark/src/token/token_service.rs`, with the supporting plumbing in `spark_rpc_client.rs`, `services/models.rs`, and `spark-wallet`.
- Populate `TokenTransaction.outputs` in the V3 broadcast path so downstream consumers see the produced outputs.
- Adds a new dependency on the published [`spark-token-primitives`](https://crates.io/crates/spark-token-primitives) crate (v0.1.3) for shared token primitives.


crates/spark/src/token/token_service.rs
```diff
@@ -587,6 +590,330 @@ impl TokenService {
...
+                    self.refresh_tokens_outputs().await?;
+                    continue;
+                }
+                Err(e) => return Err(e),
+            }
+        };
+
+        // Refresh outputs from the operator instead of extracting them from the
+        // response, since FinalTokenOutput lacks the id/revocation fields needed
+        // to construct local TokenOutput entries directly.
```
dangeross: Seems the `revocation_commitment` is in `FinalTokenOutput`, is this what you're implying by "revocation fields"?

crates/spark/src/token/token_service.rs
```diff
@@ -582,9 +683,93 @@ impl TokenService {
...
+        let outputs = broadcast_response
+            .final_token_transaction
+            .map(|ft| {
+                ft.final_token_outputs
+                    .into_iter()
+                    .map(|fo| (fo, self.network).try_into())
+                    .collect::<Result<Vec<TokenOutput>, ServiceError>>()
+            })
+            .transpose()?
+            .unwrap_or_default();
```
dangeross: I guess there is no real need to validate the v3 token transaction like in v2, just because its built/broadcast in a single step. But do you think its worth doing some sanity check/logging that the final transaction outputs match the receiver outputs?

crates/spark/src/services/models.rs
```diff
@@ -775,6 +775,40 @@ impl TryFrom<(operator_rpc::spark_token::TokenOutput, Network)> for TokenOutput
...
+        let revocation_commitment = hex::encode(output.revocation_commitment);
+        let token_identifier = bech32m_encode_token_id(&partial.token_identifier, network)?;
+        let token_amount = u128::from_be_bytes(
+            partial
+                .token_amount
+                .try_into()
+                .map_err(|_| ServiceError::Generic("Invalid token amount".to_string()))?,
+        );
+        Ok(TokenOutput {
+            id: String::new(),
```
dangeross: I just want to check, post-broadcast when calling `query_all_token_outputs()` does it return only finalized token outputs / token outputs with the id set?

Cargo.toml
```diff
@@ -179,6 +175,7 @@ platform-utils = { path = "crates/platform-utils" }
 utils = { path = "crates/utils" }
 proc-macro2 = "1.0.97"
 prost = "0.13.4"
+prost-build = "0.13.3"
```
danielgranhao: Is this needed? Also, version differs from `prost` and `prost-types`

crates/spark/src/token/token_service.rs
```diff
@@ -521,24 +528,10 @@ impl TokenService {
...
-        self.token_output_service
-            .insert_token_outputs(&TokenOutputs {
-                metadata: reservation.token_outputs.metadata,
-                outputs,
-            })
-            .await?;
+        // Refresh outputs from the operator instead of extracting them from the
+        // response, since FinalTokenOutput lacks the id/revocation fields needed
+        // to construct local TokenOutput entries directly.
+        self.refresh_tokens_outputs().await?;
```
danielgranhao: This can become problematic for high-throughput use cases. Would it make sense to query for the tx itself to get the outputs in case it really isn't possible to add the actual id to `FinalTokenOutput`?

crates/spark/src/services/models.rs
```diff
@@ -775,6 +775,40 @@ impl TryFrom<(operator_rpc::spark_token::TokenOutput, Network)> for TokenOutput
     }
 }
 
+/// `FinalTokenOutput` carries no SO-local `id` and no `token_public_key`;
+/// those are left empty / `None`.
+impl TryFrom<(operator_rpc::spark_token::FinalTokenOutput, Network)> for TokenOutput {
```
danielgranhao: I think it might be preferable to add a separate type to represent these outputs that we don't have ids/pubkeys for. We only use the outputs in order to construct payments on the SDK level (for which ids happen not to matter, but they could matter for some future use). A separate type with just what is needed would be more transparent. Example:

`TransferOutputSummary { owner_public_key, token_identifier, token_amount }` which would then be used by `token_transaction_to_payments`

Cargo.toml
peterrojs: Leftover, removed it

crates/spark/src/services/models.rs
peterrojs: yes, it just returns finalized outputs

crates/spark/src/token/token_service.rs
peterrojs: IDs not being included and not consistent over SOs is a regression on our part, let me work on fixing it, we shouldn't be adopting it as a feature

peterrojs: Releasing the fix to prod today, otherwise removed the refresh call

crates/spark/src/services/models.rs
peterrojs: if there is an id available do we still need this?

peterrojs:
I don't believe I explained actually what is the situation now. So I completely forgot that we moved from UUIDs for token output IDs to prev_tx_hash and vout. And protos do not change the order of arrays so we can actually get the ID from the current structure without any changes

crates/breez-sdk/breez-itest/src/helpers.rs
```diff
@@ -11,6 +11,12 @@ use testcontainers_modules::postgres::Postgres;
 use tokio::sync::{OnceCell, mpsc};
 use tracing::{Instrument, debug, info};
 
+fn install_rustls_provider() {
+    rustls::crypto::ring::default_provider()
+        .install_default()
+        .ok();
+}
+
```
dangeross: We've fixed some TLS issues in the repo with panicking over provider selection (aws-lc-rs / ring).  Should be fixed and this no longer needed post-rebase

.github/workflows/docs-ci.yml
```diff
@@ -90,6 +90,10 @@ jobs:
           # crates/breez-sdk/bindings/Makefile changes.
           key: gobley-uniffi-bindgen-v0.3.7-with-fixes-${{ runner.os }}
 
+      - name: Ensure Rust toolchain active
+        if: runner.os == 'macOS'
+        run: rustup default stable
+
```
dangeross: There seems to be issues in the CI at the moment across several PRs involving rustup. I would ignore

```
error: error: unexpected argument 'xtask' found

Usage: rustup-init[EXE] [OPTIONS]
```

crates/spark/src/token/token_service.rs
```diff
@@ -570,25 +573,6 @@ impl TokenService {
...
-                prev_tx_vout: vout as u32,
-            })
-            .collect::<Vec<_>>();
-        self.token_output_service
-            .insert_token_outputs(&TokenOutputs {
-                metadata: reservation.token_outputs.metadata,
-                outputs,
-            })
-            .await?;
-
```
dangeross, line 660: I think now that we're no longer relying on `refresh_tokens_outputs()`, we should revert this change to update the token store inline

crates/spark/src/token/token_service.rs
peterrojs: added!

dangeross (commented):
It seems like the [breez-itest is throwing an error because there is a mismatch](https://github.com/breez/spark-sdk/actions/runs/25926901575/job/76210475204?pr=861#step:4:5706) in the expected number of payments (each send being inserted twice? One from the transfer_tokens call, one from the sync). I think the issue might be the returned TokenTransaction from the transfer_tokens() calls contains a partial hash, not the final hash

crates/breez-sdk/breez-itest/src/helpers.rs
peterrojs: **Applied.** Removed `install_rustls_provider()` and the `rustls` dependency from breez-itest — the provider panic is now handled upstream.

_Review response generated using `/pr-review-respond` Claude command._

crates/spark/src/token/token_service.rs
peterrojs: **Applied.** Restored the inline `insert_token_outputs()` call after a successful broadcast — our change outputs are now immediately available in the local token store without waiting for the next sync cycle. The loop now breaks with `(token_transaction, reservation)` again, matching the V2 pattern.

_Review response generated using `/pr-review-respond` Claude command._

peterrojs:
Regarding the breez-itest payment duplication: the inline `insert_token_outputs()` fix should address this. When `transfer_tokens` returned without updating the local token store, the next sync would re-fetch and re-insert all outputs from scratch, potentially creating conflicting payment entries.

On the partial vs final hash question: the `partial_token_transaction_hash` from `spark_token_primitives` is the only hash algorithm defined for V3 transactions — there is no `hash_final_token_transaction` in the library. My reading is that this IS the canonical hash for V3 (it is what the wallet signs and what the SO indexes by), which should make the payment IDs consistent between `transfer_tokens` and sync.

If there is still a duplication after the inline store fix, the issue may be that the SO is computing a different canonical hash for V3 transactions than what `spark_token_primitives` returns. In that case the cleanest fix would be to add a `token_transaction_hash` field to `BroadcastTransactionResponse` so the wallet gets the authoritative hash directly from the broadcast response. Happy to implement that if confirmed.

_Review response generated using `/pr-review-respond` Claude command._

.github/workflows/docs-ci.yml
peterrojs: makes sense

crates/spark/src/services/models.rs
danielgranhao: No need, thanks

crates/spark/src/token/token_service.rs
```diff
@@ -602,27 +604,134 @@ impl TokenService {
...
+        .map_err(|_| ServiceError::Generic("client_created_timestamp overflows i64".to_string()))?;
+
+        let result = spark_token_primitives::construct_partial_transfer_transaction(
+            spark_token_primitives::TransferBuildRequest {
+                identity_public_key: identity_public_key_bytes.clone(),
+                selected_outputs,
+                receiver_outputs: prim_receiver_outputs,
+                operator_identity_public_keys: self.get_operator_identity_public_keys()?,
+                network: u32::try_from(self.network.to_proto_network() as i32)
+                    .expect("network proto value is non-negative"),
```
dangeross: Can we replace the expect with a ServiceError?

dangeross (commented):
Just this one nit then I think we're good. I tested locally and all tests pass

crates/spark/src/token/token_service.rs
peterrojs: done!

crates/spark/src/token/token_service.rs
```diff
@@ -602,27 +604,135 @@ impl TokenService {
...
+            })
+            .collect::<Vec<_>>();
+
+        let partial_txid = hex::encode(&result.partial_token_transaction_hash);
 
         if let Some(observer) = &self.transfer_observer {
             observer
                 .before_send_token(
-                    &txid,
+                    &partial_txid,
```
dangeross, line 799: My only concern here now is that we're emitting the partial txid to the payment observer, as the final tx hash is not available until after broadcast. The consequence is the payment observer cannot track the payment post broadcast (can still reject it). We could emit it after the broadcast with the final txid, but it is then not "before send". 

Wdyt @roeierez?

An option might be adding an after_send_token fn to the trait, where we provide the partial_txid and final_txid, so observers can correlate the txids

roeierez: Right. I see we do have the full transaction bellow before we broadcast right?

roeierez (approved):
LGTM

crates/spark/src/token/token_service.rs
peterrojs: closing the loop, synced offline and agreed to add `after_send_token` and also to rename `tx_id` in `before_send_token` to `partial_tx_id`

crates/breez-sdk/core/src/models/payment_observer.rs
danielgranhao: This is worth updating

crates/spark/src/token/token_service.rs
```diff
@@ -700,9 +810,132 @@ impl TokenService {
...
+        )
+        .map_err(|e| {
+            ServiceError::Generic(format!(
+                "Failed to compute final token transaction hash for tx {partial_txid}: {e}"
+            ))
+        })?;
+        let txid = hex::encode(&final_tx_hash);
+
+        if let Some(observer) = &self.transfer_observer {
+            observer.after_send_token(&partial_txid, &txid).await?;
```
danielgranhao: Maybe log and continue instead of propagating error? If we got here we've broadcasted the tx. Returning an error will result in reserved token outputs being wrongly put back on the available pool.

danielgranhao (commented):
Looking good. Final comments from my side

crates/breez-sdk/core/src/models/payment_observer.rs
peterrojs: done

crates/spark/src/token/token_service.rs
peterrojs: done

crates/spark/src/token/token_service.rs
```diff
@@ -671,27 +673,135 @@ impl TokenService {
...
+
+        // SO sorts inputs by vout before validating owner signatures; match that here.
+        let mut inputs = inputs;
+        inputs.sort_by_key(|o| o.prev_tx_vout);
+
+        let mut receiver_outputs = receiver_outputs;
+        let inputs_amount = inputs.iter().map(|o| o.output.token_amount).sum::<u128>();
+        let outputs_amount = receiver_outputs.iter().map(|o| o.amount).sum::<u128>();
+        if inputs_amount > outputs_amount {
+            receiver_outputs.push(TransferTokenOutput {
```
danielgranhao, line 696: The change output is pushed onto `receiver_outputs` *before* the `before_send_token` call below, so observers now see it as an extra self-payment (send 100 → observer gets `{recipient, 100}` + `{self, change}`). On `main` change was added to a clone and stayed out of the observer list. It's also a bit inconsistent with the rest of the pipeline, which treats change as non-payable (`utils/token.rs` `token_transaction_to_payments`), and `ProvisionalPaymentDetails::Token` has no field to tell change apart. We should filter change before

crates/breez-sdk/core/src/models/payment_observer.rs
```diff
@@ -37,6 +37,13 @@ pub enum ProvisionalPaymentDetails {
     },
 }
 
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct PaymentIdUpdate {
+    pub partial_tx_id: String,
+    pub final_tx_id: String,
```
danielgranhao: `before_send` advertises payment ids as `{partial_txid}:{index}`, but `PaymentIdUpdate` only carries the bare tx ids. So to apply the update, an integrator has to prefix-replace and assume the provisional output order lines up with the final tx's vout order, which nothing here guarantees. Could we carry the per-output mapping?

crates/spark/src/token/token_service.rs
```diff
@@ -671,27 +673,135 @@ impl TokenService {
             return Err(ServiceError::NeededTooManyOutputs);
         }
 
-        let partial_tx = self
-            .build_transfer_token_transaction(inputs.clone(), receiver_outputs.clone())
-            .await?;
-        let final_tx = self.start_transaction(partial_tx).await?;
-        let txid = hex::encode(final_tx.compute_hash(false)?);
+        let identity_public_key = self.signer.get_identity_public_key().await?;
```
danielgranhao: nit: `get_identity_public_key()` is fetched here on every attempt and again in `transfer_tokens` after the loop. It's a stable value, so we could fetch once and reuse it to save the extra signer round-trip

danielgranhao (commented):
Sorry for the back and forth, looks like opus 4.8 was able to find a couple more things that are worth improving :)

peterrojs:
> Sorry for the back and forth, looks like opus 4.8 was able to find a couple more things that are worth improving :)

no worries!

crates/spark/src/token/token_service.rs
peterrojs: done!

crates/spark/src/token/token_service.rs
peterrojs: done!

crates/breez-sdk/core/src/models/payment_observer.rs
peterrojs: done, `PaymentIdUpdate` carries the full per-output ids now. still leaning on index to vout order internally (change appended last, SO preserves order)

danielgranhao (approved):
LGTM

PR #863, opened 2026-05-07
Share SSP connection between SDK instances (and replace bitreq with reqwest)

JssDWt:
First of all, replace bitreq with reqwest. bitreq only supports HTTP/1.1. See https://github.com/rust-bitcoin/corepc/issues/587. We need better performance for requests to the SSP. They need to be fully parallel.

Second, similar to #856, make the SSP connection shareable between sdk instances. For this one we have a single reqwest client, rather than a multiplexed connection like in the SO connection. This is because reqwest doesn't support multiple loadbalanced http/2 connections to the same host at the moment. A single client should cover most cases.


crates/breez-sdk/core/src/connection_manager.rs
```diff
@@ -0,0 +1,39 @@
...
+///
+/// Pass the returned `Arc<SspConnectionManager>` to
+/// [`SdkBuilder::with_ssp_connection_manager`](crate::SdkBuilder::with_ssp_connection_manager)
+/// when building each SDK instance that should share the underlying HTTP
+/// connection pool.
+#[cfg_attr(feature = "uniffi", uniffi::export)]
+#[allow(clippy::needless_pass_by_value)]
+pub fn new_ssp_connection_manager(user_agent: Option<String>) -> Arc<SspConnectionManager> {
+    Arc::new(SspConnectionManager {
+        client: create_http_client(user_agent.as_deref()),
```
dangeross: Should we set a default user agent here like in `GraphQLClient::new`?

crates/breez-sdk/core/src/connection_manager.rs
```diff
@@ -0,0 +1,39 @@
...
+/// of how many wallets are loaded — useful for multi-tenant servers running
+/// many SDK instances.
+///
+/// # Caveats
+///
+/// - The user-agent of the first SDK to construct this manager is reused for
+///   all subsequent instances. This is rarely a problem since SDK instances
+///   in one process typically share a build version.
+/// - Connections close when the last `Arc<SspConnectionManager>` is dropped.
+///   `BreezSdk::disconnect` does not close them.
```
dangeross, line 23: Is it worth adding a docs comment (along with with_ssp_connection_manager docs)?

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -298,6 +301,23 @@ impl SdkBuilder {
...
+    /// `SdkBuilder` whose SSP traffic should share a single underlying
+    /// `reqwest::Client` (and its HTTP/2 connection pool). Useful for
+    /// multi-tenant servers running many SDK instances in one process.
+    ///
+    /// If not set, each SDK instance constructs its own internal HTTP client.
+    #[must_use]
+    pub fn with_ssp_connection_manager(
+        mut self,
+        manager: Arc<crate::SspConnectionManager>,
+    ) -> Self {
```
danielgranhao, line 316: Also need to expose this in uniffi (`bindings.rs`). 

Also flutter is missing both this and connection manager constructor

crates/platform-utils/src/http/client.rs
```diff
@@ -1,22 +1,35 @@
...
             builder = builder.user_agent(ua);
         }
+        #[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
+        {
+            builder = builder
+                .pool_max_idle_per_host(usize::MAX)
+                .pool_idle_timeout(Duration::from_secs(90))
+                .tcp_keepalive(Some(Duration::from_mins(1)))
+                .tcp_nodelay(true);
+        }
```
danielgranhao, line 33: Did you consider `http2_keep_alive_interval` + `http2_keep_alive_while_idle`? Seems like TCP keepalive alone won't survive intermediaries that reap idle HTTP/2 flows which may cause the first request after a long idle gap to hang on a half-open connection.

danielgranhao (changes requested):
Looks good! Worth adding a test or adjusting the tests introduced in #856 to also share the SSP connection manager?

danielgranhao (approved):
LGTM

dangeross (approved):
LGTM

PR #864, opened 2026-05-07
Match mysql isolation level to postgres (READ COMMITED).

roeierez:
This matches the mysql tx isloation level to the same as in postgress.


.claude/worktrees/vigilant-noyce-9bdb12
dangeross, line 1: I think this leaked into the PR

dangeross (commented):
Do we need to apply this to the SDK `MysqlStorage` implementation also? I see some default `TxOpts` used
```
.start_transaction(TxOpts::default())
```

If yes, should we consider adding it to the js/wasm implementation also? I think we need to add in `_withTransaction()`:
```
await conn.query("SET TRANSACTION ISOLATION LEVEL READ COMMITTED");
```

crates/breez-sdk/wasm/js/mysql-token-store/index.cjs
```diff
@@ -520,6 +522,7 @@ class MysqlTokenStore {
     try {
       const conn = await this.pool.getConnection();
       try {
+        await conn.query("SET TRANSACTION ISOLATION LEVEL READ COMMITTED");
         await conn.beginTransaction();
 
         await this._upsertMetadata(conn, tokenOutputs.metadata);
```
danielgranhao, line 524: Hm... Should `insertTokenOutputs ` have been using `_withTransaction` in the first place? I know it's not related to this PR, but maybe worth fixing

danielgranhao (approved):
Looks good, just raising a prexisting issue - can be fixed now or separately

crates/breez-sdk/wasm/js/mysql-token-store/index.cjs
roeierez: You are right. I fixed that and another query to use the wrapper.

dangeross (approved):

PR #865, opened 2026-05-08
feat: add mysql config to not use foreign keys

rloomba:


dangeross:
I'm closing this in favor of #873 which builds on this PR, integrating it into the MySQL connection pool and adding the config also to Wasm

PR #866, opened 2026-05-08
feat: allow prefixing of database tables

rloomba:


crates/spark-storage/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+            .unwrap_or_else(|| identifier.to_string())
+    }
+
+    /// Applies the configured prefix to known SDK storage identifiers in SQL.
+    ///
+    /// The rewriter is intentionally narrow: it only rewrites exact identifiers
+    /// from the SDK storage allowlist and skips string literals, so JSON paths
+    /// and user data are left alone.
+    #[must_use]
+    pub fn sql(&self, sql: &str) -> String {
```
JssDWt: I am leaning towards preferring format! statements on every query with the prefix instead. Just to avoid having to maintain the STORAGE_IDENTIFIERS list.

roeierez:
Thanks @rloomba!
To make this less intrusive, we're considering adding a namespace prefix ("brz_") to our table names. For example:
payments → brz_payments
leaves → brz_leaves
etc.
I think this should effectively address your use case, and it's a worthwhile change regardless. What do you think? Let us know if this works for you.

rloomba:
> Thanks @rloomba!
> 
> To make this less intrusive, we're considering adding a namespace prefix ("brz_") to our table names. For example:
> 
> payments → brz_payments
> 
> leaves → brz_leaves
> 
> etc.
> 
> I think this should effectively address your use case, and it's a worthwhile change regardless. What do you think? Let us know if this works for you.

That works!

PR #867, opened 2026-05-08
Share Postgres / MySQL connection pools across SDK instances

dangeross:
For multi-tenant server deployments (one Breez SDK per user), the SDK
forced one connection pool per `BreezSdk` — a server hosting N users opened
N independent pools to the same database. This PR replaces the
`with_postgres_backend(config)` / `with_mysql_backend(config)` API with a
shareable wrapper:

- New `create_postgres_connection_pool(config) -> Arc<PostgresConnectionPool>`
  and `create_mysql_connection_pool(config) -> Arc<MysqlConnectionPool>`.
- New `SdkBuilder::with_postgres_connection_pool(pool)` /
  `with_mysql_connection_pool(pool)` — pass the same `Arc` to multiple
  builders to share connections across SDKs.
- Per-tenant scoping (rows isolated by seed identity) is preserved — the
  pool is shared, the data isn't.
- `BreezSdk::disconnect()` does **not** close a shared pool; the pool's
  lifecycle belongs to the integrator and ends when the last `Arc` drops.

**Demonstration**

A new bench `pool-share-perf` builds 32 SDK instances (different seeds /
tenants) against a single Postgres database, once with each instance
getting its own pool, once with all instances sharing one pool. Snapshots
`pg_stat_activity` from a separate connection at each lifecycle step.

| Metric | Separate (32 pools) | Shared (1 pool) | Win |
|---|---|---|---|
| **Peak DB connections** | **96** | **8** | **12× fewer** |
| Idle DB connections (after init) | 96 | 8 | 12× fewer |
| SDK init time (sequential, 32 instances) | 492ms | 148ms | 3.3× faster |
| Concurrent `get_info()` (all 32 in parallel) | 1.47ms | 0.95ms | ~35% faster |
| Connections after `disconnect()` | 0 | 8 | (pool still held by integrator) |
| Connections after pool dropped | 0 | 0 | (pool actually closes) |

**Observations**

- **96 → 8 connections.** Each SDK opens 3 stores (storage / tree
  store / token store) — separate-pool, 32 SDKs × 3 stores = 96 idle
  connections held even with no queries. Shared-pool needs at most
  `max_pool_size = 8` total. With default `max_pool_size = 32`, separate
  mode would want 32 × 3 × 32 = 3072 connections, far past Postgres'
  default `max_connections = 100`. *That's the failure shared-pool fixes.*
- **3.3× faster init** is migration amortization — each separate pool runs
  schema migrations independently; the shared pool runs them once and the
  rest no-op.
- **Lifecycle is correct.** Separate-mode → 0 connections after
  `disconnect()`. Shared-mode → 8 connections persist after `disconnect()`
  because the integrator still holds the `Arc`. After explicitly dropping
  the pool, both go to 0. Confirms the docstring claim.
- **Concurrent-query latency** moved by ~0.5ms; both modes have plenty of
  available connections at this scale, so contention is negligible. The
  metric matters more at higher concurrency.

The existing `concurrent-perf` bench was also re-run end-to-end (4 SDKs
sharing a pool, 100 regtest payments): **100/100 success, throughput
within 1% of baseline, p99 latency improved by 3.25s.**

### DIFF
--- a/crates/breez-sdk/core/src/sdk_builder.rs
+++ b/crates/breez-sdk/core/src/sdk_builder.rs
@@ -65,9 +65,9 @@ pub struct SdkBuilder {
     storage_dir: Option<String>,
     storage: Option<Arc<dyn Storage>>,
     #[cfg(feature = "postgres")]
-    postgres_backend_config: Option<crate::persist::postgres::PostgresStorageConfig>,
+    postgres_pool: Option<Arc<crate::persist::postgres::PostgresConnectionPool>>,
     #[cfg(feature = "mysql")]
-    mysql_backend_config: Option<crate::persist::mysql::MysqlStorageConfig>,
+    mysql_pool: Option<Arc<crate::persist::mysql::MysqlConnectionPool>>,
     chain_service: Option<Arc<dyn BitcoinChainService>>,
     fiat_service: Option<Arc<dyn FiatService>>,
     lnurl_client: Option<Arc<dyn platform_utils::HttpClient>>,
@@ -98,9 +98,9 @@ impl SdkBuilder {
             storage_dir: None,
             storage: None,
             #[cfg(feature = "postgres")]
-            postgres_backend_config: None,
+            postgres_pool: None,
             #[cfg(feature = "mysql")]
-            mysql_backend_config: None,
+            mysql_pool: None,
             chain_service: None,
             fiat_service: None,
             lnurl_client: None,
@@ -124,9 +124,9 @@ impl SdkBuilder {
             storage_dir: None,
             storage: None,
             #[cfg(feature = "postgres")]
-            postgres_backend_config: None,
+            postgres_pool: None,
             #[cfg(feature = "mysql")]
-            mysql_backend_config: None,
+            mysql_pool: None,
             chain_service: None,
             fiat_service: None,
             lnurl_client: None,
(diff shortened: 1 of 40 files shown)


danielgranhao (approved):
Looks great

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -180,28 +180,43 @@ impl SdkBuilder {
...
+    /// Construct the pool once via
+    /// [`create_postgres_connection_pool`](crate::create_postgres_connection_pool) and pass the same
+    /// `Arc` to multiple `SdkBuilder` instances to share connections across
+    /// SDKs. Per-tenant scoping is derived from each SDK's seed.
+    ///
+    /// # Arguments
+    /// - `pool`: The shared `PostgreSQL` connection pool.
     #[must_use]
     #[cfg(feature = "postgres")]
-    pub fn with_postgres_backend(
```
JssDWt, line 189: Perhaps we can keep these for backward compatibility? Make it create a pool and call with_postgres_connection_pool? (same for mysql)

JssDWt (commented):
Looking good!

crates/breez-sdk/core/src/sdk_builder.rs
dangeross: I've re-added them, with a caveat that the return signature changes to avoid a panic [bf815dd](https://github.com/breez/spark-sdk/pull/867/commits/bf815dda17115f25d32ff934050da41748cffbce)

JssDWt (approved):

roeierez (approved):
LGTM

PR #868, opened 2026-05-08
expose session manager

JssDWt:
Previously we just had the in-memory session manager. This PR adds the option to run with a shared, persistent session manager. It adds postgres and mysql implementations.

The PR also makes a bit of a refactor on how the session manager works. There are different headers for the SO and SSP (as per #819). We don't want to put that burden into the session manager, especially when it is shared on the outside. The session manager should be very simple. It should store session tokens. That's it. So I created header providers. There is an auth header provider, and a jwt header provider. They use the session store underneith, but that logic is taken out of the clients themselves. I think this is a clean and simple structure, and it exposes well to the outside.

Perhaps we could have called the SessionManager SessionStore now. But I kept the name just to not throw away everything.


crates/breez-sdk/wasm/src/sdk_builder.rs
```diff
@@ -336,6 +351,14 @@ impl SdkBuilder {
                         .await?;
                 let token_store = Arc::new(WasmTokenStore::new(token_store_js));
                 self.builder = self.builder.with_token_output_store(token_store);
+
+                let session_manager_js =
```
roeierez: Seems that we override the user supplied session manager here and bellow?

crates/spark/src/ssp/graphql/client.rs
roeierez, line 165: Do we invalidate the cache here or the retry loop will use the same cached token?

roeierez (changes requested):
Looks good and clean.

crates/spark/src/ssp/graphql/client.rs
JssDWt: We don't. This is the way it works on main too. Shall we address that in a separate PR?

crates/breez-sdk/wasm/src/sdk_builder.rs
JssDWt: We now track whether the session manager was user-supplied already.

crates/spark/src/ssp/graphql/client.rs
roeierez: Yes. we might need an invalidate/delete method.

roeierez (approved):

PR #869, opened 2026-05-08
Kotlin multi-tenant server bench

danielgranhao:
## Summary

Adds `crates/breez-sdk/breez-bench/kotlin/` — an HTTP server that wraps the
Breez SDK plus an open-loop load generator, used to estimate per-pod RPS
capacity for partners running the SDK server-side at multi-tenant scale.

- **Server** (`Main.kt`, Ktor): three endpoints (`/info`, `/send`, `/receive`)
  that do `connect → op → disconnect` per request against a shared MySQL
  backend, with a per-`userId` mutex so same-user requests serialize while
  different users run in parallel.
- **Load generator** (`LoadGen.kt`): open-loop dispatcher (fires at target RPS
  regardless of completion) with a `--max-in-flight` cap so backpressure
  surfaces as `dropped=true` rows instead of being hidden by closed-loop
  self-clocking.
- **Sweep driver** (`scripts/sweep.sh`): restarts the server fresh per step
  (cold-start cost is part of the baseline), runs idempotent
  `fund` + `seed-senders` pre-steps, sweeps a list of target RPS values, then
  invokes the aggregator.
- **Aggregator** (`scripts/aggregate.py`, stdlib only): per-op p50/p95/p99
  (client + server views), 1Hz process metrics, error categorization, and a
  headline `max_safe_rps` derived as the highest swept RPS where
  client-side p99(send) is < 2× the lowest stable p99(send).
- **Closed-loop funding**: `/send` always targets a treasurer wallet; sender
  pool drains, treasurer fills, system total stays constant. Workload-sized
  top-up computed from sweep config; faucet hit only when the treasurer
  drops below threshold (wallets persist across runs via a fixed
  `MASTER_SECRET`).
- **Docs**: `README.md` (how to run), `DESIGN.md`

This is the **worst-case baseline** — every request pays cold-start (SDK
build + initial sync + operator TLS handshakes). SDK-level work in flight
upstream (shared MySQL pool, shared operator connections) and an in-process
SDK pool layer would both improve these numbers without harness changes.

~~No production code touched; bench lives in its own subdirectory and is not
wired into the workspace build.~~ -> I've bundled a fix for a chain service sharing issue that caused a panic on Uniffi integrations.


dangeross (commented):
Reviewed the whole bench in detail (DESIGN.md, README.md, all five Kotlin files, both Python scripts, sweep.sh, build.gradle.kts, Makefile). Overall this is a well-thought-out, self-contained harness with strong design rationale; the open-loop dispatcher, per-step server restart, idempotent fund/seed pre-steps, treasurer-address caching, and two-views latency separation are all the right calls. Below are findings ordered by impact. Nothing here is a blocker for a draft PR — most of these are paper cuts.

## Worth fixing

**1. `fund` mode requires `FAUCET_USERNAME`/`FAUCET_PASSWORD` even on a no-op run** — `Main.kt:360-361`

```kotlin
val targetSats = opts["target-sats"]?.toLongOrNull() ?: 5_000_000L

System.getenv("FAUCET_USERNAME") ?: error("FAUCET_USERNAME env var is required")
System.getenv("FAUCET_PASSWORD") ?: error("FAUCET_PASSWORD env var is required")
```

These env-var checks run before the cached-balance fast-path at lines 378-383. README claims "subsequent sessions just need `make run`" because closed-loop funding keeps wallets populated, but the fund step (which `sweep.sh` runs unconditionally) will refuse to start without faucet creds even when the treasurer is already at-or-above target and no faucet call would happen. Move the env-var checks below the fast-path so a no-op fund is genuinely a no-op.

**2. Loadgen drain timeout is misleading** — `LoadGen.kt:336-345`

```kotlin
println("[loadgen] dispatch done; draining ${inFlight.get()} in-flight requests (max 60s)")
val drainEndNs = System.nanoTime() + 60_000_000_000L
while (inFlight.get() > 0 && System.nanoTime() < drainEndNs) {
    delay(100)
}
if (inFlight.get() > 0) {
    System.err.println("[loadgen] drain warning: ${inFlight.get()} requests still in flight after 60s")
}
progressJob.cancel()
```

After the 60s warning fires, `coroutineScope` still won't return until every in-flight `launch(Dispatchers.IO)` child finishes — which can be another 60s per straggler (HttpRequest timeout is also 60s). The "max 60s" message reads like an upper bound on total drain time but isn't. Either reword the warning ("…still in flight after 60s; will block until they complete or the per-request timeout fires") or actually cancel the scope on timeout.

**3. Hand-rolled JSON for `/send` body** — `LoadGen.kt:158-168`

```kotlin
val body = """{"destination":"$destination","amountSats":$amountSats}"""
```

`destination` is the treasurer's Spark address so input is controlled, but the comment "encode minimal JSON by hand to avoid coupling LoadGen to the server's `@Serializable` type" doesn't really apply here — `kotlinx.serialization.json.buildJsonObject` is already in use in `Faucet.kt`, doesn't couple to the server, and is the same line count. Using it makes a future refactor (e.g. someone adds an apostrophe or backslash to a destination) a non-issue.

**4. `synchronized(this@runBlocking)` in `seedSenders`** — `Main.kt:516-521`

```kotlin
synchronized(this@runBlocking) {
    when (outcome) {
        SeedOutcome.FUNDED  -> fundedCount++
        SeedOutcome.SKIPPED -> skippedCount++
    }
}
```

Using a `BlockingCoroutine` instance as a monitor object works but is non-idiomatic and inconsistent with how `LoadGen.kt` handles the same pattern (`AtomicLong` for `dispatched`/`completed`/`errors`). Two `AtomicInteger`s for `fundedCount`/`skippedCount` would be a cleaner fit and remove the surprise of locking on the runBlocking scope.

## Minor

**5. `MysqlConnPoller.toJdbcUrl()` drops query parameters** — `Metrics.kt:51, 62-77`

`parseMysqlUrl` extracts host/port/db/user/pass and rebuilds `jdbc:mysql://host:port/db`, silently discarding any query string from the input URL. Bench-controlled localhost MySQL doesn't need flags today, but if someone passes `mysql://…/db?useSSL=false&serverTimezone=UTC` the SDK would honor it and the JDBC sampler wouldn't, which could surface as a connect failure in `metrics.jsonl` only. Worth either preserving the query string or noting the limitation in a comment.

**6. `Faucet.kt` doesn't check HTTP status before parsing JSON** — `Faucet.kt:80-90`

A non-200 with a non-JSON body throws inside `parseToJsonElement` and gets swallowed into the retry loop with a confusing exception message. Cheap to add `if (resp.statusCode() !in 200..299) throw RuntimeException("Faucet HTTP ${resp.statusCode()}: ${resp.body().take(500)}")` before parsing.

**7. Faucet retry comment off-by-one** — `Faucet.kt:43`

`/** Retries with exponential backoff up to 3 times. */` but `maxAttempts = 4` (1 initial + 3 retries). Either reading is defensible; just make the comment match the literal.

**8. Treasurer address cache has no version stamp** — `sweep.sh:183-186`

```bash
secret_hash() {
    printf '%s' "$MASTER_SECRET" | shasum -a 256 | cut -c1-16
}
TREASURER_ADDR_CACHE="$ROOT_DIR/out/.cache/treasurer-spark-addr.$(secret_hash).txt"
```

Cache is keyed only by master secret. If the SDK ever changes how Spark addresses are derived (or address format evolves), stale caches would silently point to a wrong destination on every subsequent run. Including SDK version / a binding fingerprint in the cache key would be defensive. Low likelihood, easy mitigation.

**9. `shasum -a 256` portability** — `sweep.sh:184`

Works on macOS (perl shasum) and most Linux distros, but `sha256sum` is the more reliable choice on Linux containers / minimal images. Either-or fallback would be more robust.

**10. JDK 17 requirement undocumented** — README + Makefile

`build.gradle.kts:46` pins `jvmToolchain(17)` but README's Quickstart doesn't mention a JDK requirement; first-time users on a host without one will get a Gradle error. One line in the prerequisites would help.

**11. Unconditional `parseMysqlUrl` parsing duplication** — `Main.kt:600` and `SharedHandlers.create` both parse `mysqlUrl`. Cosmetic; passing already-parsed `MysqlUrlParts` through would dedupe.

**12. `recycleTimeoutSecs = 300UL`** — `Main.kt:146`

Comment is excellent; just flag that 300s is a magic number — would be nice as a `private const val` so it's easier to tune from one spot if TIME_WAIT churn reappears.

## Nits

- `DESIGN.md` line 89: "Sync forced — see below" — the "below" reference matches the next bullet, fine, but adding "see below in this table" or removing the dangling reference would read cleaner.
- `Main.kt:516`: `synchronized(this@runBlocking)` (see #4) — also the lambda is inside `coroutineScope { ... }`, so `this@runBlocking` reaches *out* of the inner scope; readers will momentarily wonder why.
- `LoadGen.kt:103-105`: `OpSampler.sample` falls back to `ops.last()` if no bucket matches due to FP rounding; correct, but a one-line comment would save a reader the "why is this needed?" beat.
- `aggregate.py:122-128`: `per_op_latency` filters out errors and dropped, then by missing fields. `requests.jsonl` rows always have `duration_ms` (server populates in `finally`), so the `d is None` branch is dead for server-side; harmless.
- `Metrics.kt:255`: macOS `remoteTcpSocketCount = -1` is documented in the class comment; `RESULTS.md` and `DESIGN.md` say "real measurements should be taken on Linux", consistent.
- `compute_funding.py:14-17`: `parse_duration_secs` and `sweep.sh:55-63` `parse_duration_secs` accept different-but-overlapping forms (Python doesn't accept `H`/`M`/`S` upper-case; bash version does). Inconsequential but worth aligning.

## Things I checked that look fine

- `withMysqlConnectionPool` / `withSspConnectionManager` / `withConnectionManager` / `withChainService` all exist on `SdkBuilder` (`bindings.rs:107-189`, `sdk_builder.rs:226-394`). The bench's transport-sharing model is sound.
- Hardcoded chain creds (`spark-sdk` / `mCMk1JqlBNtetUNy`) in `SharedHandlers.create` mirror the same defaults already in `crates/breez-sdk/core/src/sdk_builder.rs:532-533` and `crates/internal/src/config.rs:20-21`. Not a new disclosure.
- Open-loop dispatch math: `nextDispatchNs += intervalNs` correctly preserves cadence under late ticks; the `delay(sleepMs) else yield()` switch handles sub-ms waits without burning CPU.
- `stream_filtered` correctly captures `${PIPESTATUS[0]}` (the gradle command, not `tee` or `grep`), so a "no matching lines" grep return doesn't shadow a real failure.
- `stop_server` covers the gradle-wrapper-orphans-JVM-child case via `lsof -t -iTCP:$PORT -sTCP:LISTEN`, which is the right shape; the 15s grace is calibrated to the 10s `JsonlWriter.close()` join.
- `aggregate_step` correctly excludes `dropped=true` rows from error totals (loadgen-side over-cap, not a server failure) — `aggregate.py:191-194`.
- `derive_p99_doubling` correctly requires ≥30 send samples for "stable", and the headline is documented as "client-side p99(send) < 2× lowest stable" in both DESIGN.md and the README.

Solid harness. The fund-mode env-var check (#1) and the misleading drain timeout (#2) are the two most worth addressing before sharing numbers; everything else is polish.

---
_Generated by [Claude Code](https://claude.ai/code)_

roeierez:
@danielgranhao I have been running this bench for the last week and it looks great to me as a first step. Do you have other things in mind before moving it out of draft?

docs/breez-sdk/src/guide/customizing.md
```diff
@@ -94,7 +96,16 @@ The SDK provides a default Bitcoin Chain Service implementation. If you want to
...
 
-Construct one via {{#name new_rest_chain_service}} and pass it to each {{#name SdkBuilder}} via {{#name with_chain_service}}. All SDK instances sharing the chain service must be configured for the same network.
+Construct one via {{#name new_rest_chain_service}} and pass the returned handle to each {{#name SdkBuilder}} via {{#name with_chain_service}}. All SDK instances sharing the chain service must be configured for the same network.
+
+{{#tabs sdk_building:with-shared-rest-chain-service}}
+
+<div class="warning">
+<h4>Developer note</h4>
+
+In Rust, {{#name new_rest_chain_service}} returns an `Arc<BitcoinChainServiceHandle>`; call `.inner()` on it to extract the `Arc<dyn BitcoinChainService>` that {{#name with_chain_service}} accepts. All non-Rust bindings hand the handle to {{#name with_chain_service}} directly.
```
dangeross: Can we make the Rust API consistent with the UniFFI one?

crates/breez-sdk/core/src/chain/mod.rs
```diff
@@ -71,30 +71,97 @@ pub struct RecommendedFees {
...
+/// Only built when the `uniffi` feature is enabled. Rust callers should
+/// use [`BitcoinChainServiceHandle::new`] directly.
+#[cfg(feature = "uniffi")]
+#[uniffi::export]
+#[must_use]
+pub fn wrap_chain_service(
+    chain_service: Arc<dyn BitcoinChainService>,
+) -> Arc<BitcoinChainServiceHandle> {
+    BitcoinChainServiceHandle::new(chain_service)
+}
```
dangeross: I'm trying to think of ways to avoid exposing differences in the API (and wondering if the differences are necessary). The changes to `customizing.md` I feel are exposing to much implementation detail.

docs/breez-sdk/src/guide/customizing.md
```diff
@@ -78,6 +78,8 @@ The MySQL tree store can use the same or a separate MySQL database as the MySQL
 
 The SDK provides a default Bitcoin Chain Service implementation. If you want to use your own, you can provide it either by using [With REST Chain Service](#with-rest-chain-service) or by implementing the Bitcoin Chain Service interface.
 
+In the UniFFI bindings (Kotlin, Swift, Python, Go, React Native, C#) {{#name with_chain_service}} takes an opaque `BitcoinChainServiceHandle` rather than the trait directly. Pass a foreign-language implementation of `BitcoinChainService` through {{#name wrap_chain_service}} first to obtain a handle. Rust and WASM callers pass the trait/interface directly.
+
```
dangeross: Feels like unnecessary detail

crates/breez-sdk/core/src/chain/mod.rs
```diff
@@ -71,30 +71,97 @@ pub struct RecommendedFees {
...
+/// every method call through `UniFFI` to the foreign impl — that's correct,
+/// since the impl really is foreign — but uses the same single-method
+/// handle-based API as the Rust-built case.
+///
+/// Only built when the `uniffi` feature is enabled. Rust callers should
+/// use [`BitcoinChainServiceHandle::new`] directly.
+#[cfg(feature = "uniffi")]
+#[uniffi::export]
+#[must_use]
+pub fn wrap_chain_service(
```
dangeross: Could this just be named `new_chain_service()`?

crates/breez-sdk/core/src/chain/rest_client.rs
```diff
@@ -92,15 +116,52 @@ impl RestClientChainService {
...
+                api_type,
+            }),
+            // Captured here so each trait-method body can re-enter the
+            // surrounding runtime even when invoked from a `UniFFI`
+            // foreign-callback proxy that polls outside any tokio context.
+            // Callers reach this constructor from within an async path
+            // (`new_rest_chain_service`, `SdkBuilder::build`, etc.), so
+            // `Handle::current()` is guaranteed to find a runtime.
+            #[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
+            runtime_handle: tokio::runtime::Handle::current(),
```
dangeross, line 135: I think there is a potential issue in Rust/Flutter using this in `SdkBuilder::with_rest_chain_service` because the fn not async it's not in a tokio runtime. Should work ok for UniFFI because `SdkBuilder::with_rest_chain_service` is async there

danielgranhao: Right! Fixed by deferring construction to `build()`, which is async

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -277,10 +280,22 @@ impl SdkBuilder {
...
+    /// The chain service is constructed lazily inside [`build`](Self::build),
+    /// so this method is safe to call from sync contexts. Constructing the
+    /// service eagerly here would panic if the caller isn't already on a
+    /// tokio runtime — the inner `RestClientChainService` captures
+    /// [`tokio::runtime::Handle::current`] at construction time so it can
+    /// dispatch chain calls back onto the runtime after `UniFFI`'s
+    /// foreign-callback proxy strips the runtime context. See
+    /// [`crate::chain::rest_client::RestClientChainService`] for the full
+    /// rationale.
+    ///
```
dangeross: I think too much implementation detail for surface docs

danielgranhao: removed

roeierez (approved):
LGTM

dangeross (approved):
LGTM

PR #872, opened 2026-05-08
fix: align token invoice signing hash with spark-token-primitives

peterrojs:
There was a difference in hashing amount-less invoices, `TokensPayment { amount: None } ` hashed `sha256([0x00])` because `u128::to_unpadded_be_bytes(0)` returns `[0]`, while everyone else hashes `sha256(empty)`.

This PR migrates the hashing to spark-token-primitives for token invoices

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -196,7 +196,7 @@ serde_with = "3.13.0"
 shlex = "1.3.0"
 spark = { path = "crates/spark", default-features = false }
 spark-postgres = { path = "crates/spark-postgres" }
-spark-token-primitives = "0.1.1"
+spark-token-primitives = "0.1.2"
 spark-wallet = { path = "crates/spark-wallet", default-features = false }
 syn = "2.0.105"
 test-log = { version = "0.2.18", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -196,6 +196,7 @@ serde_with = "3.13.0"
 shlex = "1.3.0"
 spark = { path = "crates/spark", default-features = false }
 spark-postgres = { path = "crates/spark-postgres" }
+spark-token-primitives = "0.1.1"
 spark-wallet = { path = "crates/spark-wallet", default-features = false }
 syn = "2.0.105"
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -198,8 +195,8 @@ serde-wasm-bindgen = "0.6.5"
 serde_with = "3.13.0"
 shlex = "1.3.0"
 spark = { path = "crates/spark", default-features = false }
-spark-mysql = { path = "crates/spark-mysql" }
 spark-postgres = { path = "crates/spark-postgres" }
+spark-token-primitives = "0.1.1"
```
danielgranhao: nit: we can align with the locked version: `0.1.2`

danielgranhao (approved):
LGTM

Cargo.toml
peterrojs: yep, bumped it

crates/spark/src/address/mod.rs
```diff
@@ -333,72 +401,41 @@ impl SparkAddress {
...
+
+        if let SparkAddressPaymentType::TokensPayment(payment) = payment_type {
+            return Ok(self
+                .prepare_token_invoice_via_primitives(payment)?
+                .spark_invoice_hash);
+        }
+
+        let SparkAddressPaymentType::SatsPayment(payment) = payment_type else {
+            unreachable!("token payment handled above");
+        };
```
dangeross: I would feel more comfortable doing an exhaustive match here. What do you think?
```suggestion
        let payment = match invoice_fields.payment_type.as_ref() {
            Some(SparkAddressPaymentType::TokensPayment(payment)) => {
                return Ok(self
                    .prepare_token_invoice_via_primitives(payment)?
                    .spark_invoice_hash);
            }
            Some(SparkAddressPaymentType::SatsPayment(payment)) => payment,
            None => return Err(AddressError::Other("No payment type".to_string())),
        };
```

peterrojs: yeah that's much better

dangeross (approved):
LGTM

PR #873, opened 2026-05-08
Add MySQL config to not use foreign keys

dangeross:
Builds on #865 to integrate it into the MySQL connection pool and add the config to the Wasm bindings


crates/spark-mysql/src/token_store.rs
```diff
@@ -1376,6 +1381,47 @@ fn build_placeholders(n: usize) -> String {
     s
 }
 
+fn token_outputs_create_table_sql(foreign_key_mode: MysqlForeignKeyMode) -> &'static str {
+    match foreign_key_mode {
+        MysqlForeignKeyMode::Enforced => {
```
roeierez: NIT: I think we can create the table without foreign keys and then conditionally add the foreign key statements?

roeierez (approved):
Approved with a non blocking NIT.

hydra-yse (approved):
utACK

PR #876, opened 2026-05-11
Fix user_settings reinit tests on Postgres/MySQL itest harness

dangeross:
Tests `test_01_spark_private_mode_user_setting` and `test_02_stable_balance_user_setting` are failing under `make breez-itest-pg-tree-store` and `make breez-itest-mysql-tree-store`. After updating a setting, dropping the SDK, and rebuilding it via `ReinitializableSdkInstance`, `get_user_settings()` returned config defaults instead of the persisted value.

Root cause was in the itest harness, not the SDK: `apply_storage` allocates a fresh `ts_N` database on every `build_sdk()` call (via `TREE_STORE_DB_COUNTER` / `MYSQL_TREE_STORE_DB_COUNTER`). On SQLite the `storage_dir` is reused so persistence worked, but on PG/MySQL the rebuilt SDK landed in a brand-new empty database.

### DIFF
--- a/crates/breez-sdk/breez-itest/src/helpers.rs
+++ b/crates/breez-sdk/breez-itest/src/helpers.rs
@@ -92,29 +92,69 @@ async fn get_mysql_tree_store_base_url() -> Option<&'static str> {
     Some(&shared.base_url)
 }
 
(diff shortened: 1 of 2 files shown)


roeierez (approved):
LGTM

JssDWt (approved):
LGTM

PR #877, opened 2026-05-11
feat: add ability to external define database schemas

rloomba:
Adds a `schema_managed_externally` / `schemaManagedExternally` option for Postgres and MySQL SDK storage configs.

When enabled, Breez trusts that the embedding service owns and applies the database schema, so initialization skips SDK-managed migrations entirely, including writes to schema migration tables. The flag is captured on shared connection pools and threaded through storage, tree store, token store, session manager, and WASM Node backends so service-managed schemas work without fake migration-version rows.

### DIFF
--- a/crates/breez-sdk/core/src/persist/mysql/base.rs
+++ b/crates/breez-sdk/core/src/persist/mysql/base.rs
@@ -34,6 +34,11 @@ pub struct MysqlStorageConfig {
     /// Only used when the connection string requests TLS
     /// (`ssl-mode=verify_ca` or `ssl-mode=verify_identity`).
     pub root_ca_pem: Option<String>,
+
+    /// If true, the SDK trusts that the database schema is managed by the
+    /// embedding service and skips all migrations, including writes to the
+    /// schema migrations tables.
+    pub schema_managed_externally: bool,
 }
 
 impl From<MysqlStorageConfig> for spark_mysql::MysqlStorageConfig {
@@ -43,6 +48,7 @@ impl From<MysqlStorageConfig> for spark_mysql::MysqlStorageConfig {
             max_pool_size: config.max_pool_size,
             recycle_timeout_secs: config.recycle_timeout_secs,
             root_ca_pem: config.root_ca_pem,
+            schema_managed_externally: config.schema_managed_externally,
         }
     }
 }
@@ -54,6 +60,7 @@ impl From<spark_mysql::MysqlStorageConfig> for MysqlStorageConfig {
             max_pool_size: config.max_pool_size,
             recycle_timeout_secs: config.recycle_timeout_secs,
             root_ca_pem: config.root_ca_pem,
+            schema_managed_externally: config.schema_managed_externally,
         }
     }
 }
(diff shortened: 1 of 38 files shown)


crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -628,30 +628,42 @@ impl SdkBuilder {
...
                         identity,
+                        true,
                     )
                     .await
-                    .map_err(|e| SdkError::Generic(e.to_string()))?,
-                ));
+                } else {
+                    crate::persist::postgres::PostgresStorage::new_with_pool(pool.clone(), identity)
+                        .await
+                }
```
roeierez: Since we now introduced `new_with_pool_and_schema_management` function that takes the new boolean we don't need the `new_with_pool` anymore.
Perhaps just adding the boolean to the existing function instead of introducing another one?

crates/spark-postgres/src/tree_store.rs
```diff
@@ -1300,6 +1319,25 @@ pub async fn create_postgres_tree_store_from_pool(
     ))
 }
 
+/// Creates a `PostgresTreeStore` instance from an existing connection pool.
+///
+/// If `schema_managed_externally` is true, skips SDK-managed schema
+/// migrations and trusts that all required tables, columns, and indexes exist.
+pub async fn create_postgres_tree_store_from_pool_with_schema_management(
```
roeierez: Same here, I think better to extend the existing function (`create_postgress_tree_store_from_pool`) by adding the boolean flag. It will minimize the api surface.

crates/breez-sdk/core/src/persist/mysql/base.rs
```diff
@@ -34,6 +34,11 @@ pub struct MysqlStorageConfig {
     /// Only used when the connection string requests TLS
     /// (`ssl-mode=verify_ca` or `ssl-mode=verify_identity`).
     pub root_ca_pem: Option<String>,
+
+    /// If true, the SDK trusts that the database schema is managed by the
+    /// embedding service and skips all migrations, including writes to the
+    /// schema migrations tables.
+    pub schema_managed_externally: bool,
```
roeierez: Nit: I perhaps we can be here more specific with the field naming? `run_migration` ?

roeierez (changes requested):
Looks good and certainly a valid use case. Dropped some small comments mainly around naming.

roeierez:
@rloomba I created a PR on top of your branch that address my own feedback in case you want to use it: https://github.com/rloomba/spark-sdk/pull/2

rloomba:
> @rloomba I created a PR on top of your branch that address my own feedback in case you want to use it: [rloomba#2](https://github.com/rloomba/spark-sdk/pull/2)

amazing thanks! merged.

roeierez:
@rloomba I created another [PR](https://github.com/rloomba/spark-sdk/pull/3) to fix the format issues and test.
I am going to add you as collaborator to the repo so you can create branches here and CI will run. Also easier to manage.

roeierez (approved):
LGTM

docs/breez-sdk/src/guide/customizing.md
```diff
@@ -32,6 +32,8 @@ When using the SDK Builder, you either have to provide a Storage implementation
 
 The SDK includes a PostgreSQL backend as an alternative to file-based storage. Construct a connection pool once with {{#name create_postgres_connection_pool}} and pass it to the builder via {{#name with_postgres_connection_pool}} — this configures PostgreSQL for all stores (storage, tree store, and token store), which is suitable for server-side deployments with horizontal scaling. The same pool can be shared across multiple `SdkBuilder` instances; per-tenant scoping (rows isolated by seed identity) is preserved.
 
+If your service owns the database schema and applies SDK-compatible migrations externally, set `run_migration`/`runMigration` to `false` on the storage config before creating the pool. The SDK will trust the existing schema and skip all migration runs, including writes to schema migration tables.
```
dangeross, line 35: So it adapts to the selected language casing
```suggestion
If your service owns the database schema and applies SDK-compatible migrations externally, set {{#name run_migration}} to `false` on the storage config before creating the pool. The SDK will trust the existing schema and skip all migration runs, including writes to schema migration tables.
```

dangeross (commented):
Looks good. Just a nit

dangeross (approved):

PR #879, opened 2026-05-12
Install rustls provider in GrpcClient::new

dangeross:
`BreezServer::new` calls `GrpcClient::new` during SDK build, before the operator connection manager (which already installs Ring) ever runs. Tonic's `ClientTlsConfig::with_webpki_roots` builds a `rustls::ClientConfig` via `ClientConfig::builder()`, which panics under the workspace's dual-provider state. Install Ring once at the GrpcClient boundary so the auto-detect has a default to find.

Adds a `tokio::test` that constructs `GrpcClient` lazily as a regression guard.


dangeross:
Superseded by #880

PR #880, opened 2026-05-12
Use only ring as crypto provider for rustls

roeierez:
Upgrade to mysql_async 0.26 that allows declaring only ring as default provider.
This enables us remove all the rustls initialization code and also allows rust callers to pick their own provider.


dangeross (approved):
LGTM

PR #883, opened 2026-05-14
Add Postgres/MySQL `brz_` prefixes

dangeross:
Closes the issue raised in [#866](https://github.com/breez/spark-sdk/pull/866):
SDK tables like `payments`, `settings`, `contacts` collide with customer
schemas when integrating into a shared database. 
This PR uses a fixed `brz_` prefix instead of the configurable approach
proposed in #866 — no public API surface change, no config knob.

- **Postgres + MySQL only.** SQLite (mobile/desktop) and IndexedDB (browser)
  are isolated per-user databases with no collision risk and keep their
  current names.
- All SDK-owned identifiers prefixed: tables, indexes (`idx_*` → `brz_idx_*`),
  primary-key constraints (`<table>_pkey` → `brz_<table>_pkey`), and foreign
  keys (`fk_*` → `brz_fk_*`).
- Migration tracking tables also prefixed (`schema_migrations` →
  `brz_schema_migrations`), and serve as the canary for the bootstrap rename.

**Existing-deployment upgrade**

`run_migrations` now takes an optional `SchemaRenames` argument. On startup:
- Canary check probes for the legacy unprefixed migrations table.
- If present, renames every table / index / constraint to its `brz_*` form,
  preserving data and the migrations version row, under the same advisory
  lock (Postgres) or `GET_LOCK` (MySQL) used by the migration runner — one
  critical section, one transaction (Postgres) / session (MySQL).
- If absent, the rename block short-circuits; fresh DBs and already-upgraded
  DBs pay only the canary probe.

### DIFF
--- a/crates/breez-sdk/core/src/persist/mysql/storage.rs
+++ b/crates/breez-sdk/core/src/persist/mysql/storage.rs
@@ -92,6 +92,48 @@ const SCHEMA_RENAMES: SchemaRenames<'static> = SchemaRenames {
             "idx_sync_incoming_user_revision",
             "brz_idx_sync_incoming_user_revision",
         ),
+        // Pre-multi-tenant indexes (still present on version < 16 DBs).
+        // The multi-tenant migration drops these via the post-rename names.
+        (
+            "brz_payments",
+            "idx_payments_timestamp",
+            "brz_idx_payments_timestamp",
+        ),
+        (
+            "brz_payments",
+            "idx_payments_payment_type",
+            "brz_idx_payments_payment_type",
+        ),
+        (
+            "brz_payments",
+            "idx_payments_status",
+            "brz_idx_payments_status",
+        ),
+        (
+            "brz_payment_metadata",
+            "idx_payment_metadata_parent",
+            "brz_idx_payment_metadata_parent",
+        ),
+        (
+            "brz_payment_details_lightning",
+            "idx_payment_details_lightning_invoice",
+            "brz_idx_payment_details_lightning_invoice",
+        ),
+        (
+            "brz_payment_details_lightning",
+            "idx_payment_details_lightning_payment_hash",
+            "brz_idx_payment_details_lightning_payment_hash",
+        ),
+        (
+            "brz_sync_outgoing",
+            "idx_sync_outgoing_data_id_record_type",
+            "brz_idx_sync_outgoing_data_id_record_type",
+        ),
+        (
+            "brz_sync_incoming",
+            "idx_sync_incoming_revision",
+            "brz_idx_sync_incoming_revision",
+        ),
     ],
     foreign_keys: &[],
 };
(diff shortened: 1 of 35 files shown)


crates/breez-sdk/core/src/persist/postgres/storage.rs
```diff
@@ -29,10 +29,108 @@ use crate::{
...
+        ("payment_details_token", "brz_payment_details_token"),
+        ("payment_details_spark", "brz_payment_details_spark"),
+        ("lnurl_receive_metadata", "brz_lnurl_receive_metadata"),
+        ("sync_revision", "brz_sync_revision"),
+        ("sync_outgoing", "brz_sync_outgoing"),
+        ("sync_state", "brz_sync_state"),
+        ("sync_incoming", "brz_sync_incoming"),
+        ("contacts", "brz_contacts"),
+    ],
+    indexes: &[
```
roeierez, line 56: Are we missing here the indexes from before the multi user migration (idx_tree_leaves_available). If users already ran the multi user migration these will stay orphan.

roeierez (changes requested):
Other than one comment looks good.

roeierez (approved):

crates/breez-sdk/core/src/persist/mysql/storage.rs
```diff
@@ -26,11 +26,117 @@ use crate::{
...
 
-const MIGRATIONS_TABLE: &str = "schema_migrations";
+const MIGRATIONS_TABLE: &str = "brz_schema_migrations";
+
+/// Pre-prefix rename map for upgrading core persist deployments.
+const SCHEMA_RENAMES: SchemaRenames<'static> = SchemaRenames {
+    old_migrations_table: "schema_migrations",
+    new_migrations_table: MIGRATIONS_TABLE,
+    tables: &[
+        ("payments", "brz_payments"),
```
JssDWt, line 40: ~~The reason we have to prefix these tables in mysql is because people may have tables with the same name already. That means we cannot rename the table? Should we just rewrite the migrations so mysql support starts after this change is included?~~

JssDWt: Never mind, checking whether the migrations table exists I think is pretty good.

crates/spark-mysql/src/migrations.rs
```diff
@@ -145,23 +145,160 @@ impl Migration {
...
+        }
+        let sql = format!("ALTER TABLE `{table}` RENAME INDEX `{old}` TO `{new}`");
+        conn.query_drop(&sql).await.map_err(|e| {
+            MysqlError::Database(format!(
+                "Failed to rename index {old} -> {new} on {table}: {e}"
+            ))
+        })?;
+    }
+
+    for fk in renames.foreign_keys {
```
JssDWt, line 223: The drop + add here isn't atomic — `DROP FOREIGN KEY` and `ADD CONSTRAINT` auto-commit separately, so a crash between them silently drops the FK on replay. Fix: combine into one `ALTER TABLE t DROP FOREIGN KEY x, ADD CONSTRAINT y …`, which is atomic on MySQL 8.0+.

hydra-yse (approved):
Besides @JssDWt's comment, looks good.
To note: I think we should open a follow-up PR to add the prefix and conversion also to Sqlite/IndexedDB, that way we save on maintenance costs by not having to distinguish between storage implementations every time. 

Not strictly necessary atm, just good to have. @roeierez @JssDWt Wdyt?

crates/spark-mysql/src/migrations.rs
dangeross: Added [683cba8](https://github.com/breez/spark-sdk/pull/883/commits/683cba8f3b0beb8f05d4cc17c17eacb16957df0f)

JssDWt (approved):
LGTM.

@hydra-yse I think I agree with renaming sqlite and indexeddb too.

erdemyerebasmaz (approved):

PR #885, opened 2026-05-15
Add server-mode SDK profile (background_tasks_enabled=false)

roeierez:
This PR adds a master `background_tasks_enabled` flag and a server preset that disables background work by default.
Server mode keeps SDK instances lightweight by skipping per-instance background tasks like sync loops, real-time sync, wallet processors, recovery, private-mode init, and refund handling. JWT still initializes once at startup and should be refactored out in a following PR.
Runtime behavior is now handled through RuntimeProfile implementations instead of scattered flag checks. It also allows us to put all the client logic in one place and ensure no time consuming tasks affect the server runtime. 
Existing config values remain visible, but background work is gated at startup.

Public API adjustments for the no-loop world:
- sync_wallet routes through RuntimeProfile::run_user_sync, which in server mode calls sync_wallet_internal inline (no coordinator dispatch, since the receiver isn't running).
- claim_deposit no longer fires an automatic follow-up sync; in client mode the TransferClaimed event-driven path already refreshes balances and triggers a WalletState sync.
- get_info reads balance directly from the spark wallet in server mode rather than from the cached `account_info` that no background task maintains. `ensure_synced` is a no-op in this mode.
- ensure_spark_private_mode_initialized short-circuits in server mode to avoid an indexed storage read per request on ephemeral instances; partners apply Config::private_enabled_default via update_user_settings on a one-time setup pass.
- New BreezSdk::refund_pending_conversions surfaces the refund pass explicitly via a new TokenConverter::refund_pending trait method, since the periodic refunder no longer runs in server mode.
- StableBalance signals completed conversions via a RuntimeEvent on the EventEmitter; the client runtime listens and triggers sync, while the server runtime ignores it.

In summary all background tasks and work related to client side is not routed via the runtime/client module and anything that should affect high throughput  should use the "runtime" encapsulation.

### DIFF
--- a/crates/breez-sdk/breez-bench/src/bin/concurrent_perf.rs
+++ b/crates/breez-sdk/breez-bench/src/bin/concurrent_perf.rs
@@ -21,7 +21,6 @@ use breez_sdk_spark::{
     default_server_config,
 };
 
-use breez_bench::events::wait_for_claimed_event;
 use breez_bench::stats::DurationStats;
 
 #[derive(Parser, Debug)]
@@ -98,6 +97,7 @@ struct PaymentResult {
 
 struct BenchSdkInstance {
     sdk: BreezSdk,
+    #[allow(dead_code)]
     events: mpsc::Receiver<SdkEvent>,
     #[allow(dead_code)]
     temp_dir: Option<TempDir>,
@@ -900,14 +900,29 @@ async fn fund_via_faucet(
             let extra = needed.clamp(FAUCET_MIN_PER_CALL, FAUCET_MAX_PER_CALL);
             let txid = faucet.fund_address(&deposit_address, extra).await?;
             info!("Top-up faucet txid: {}", txid);
-            wait_for_claimed_event(&mut sdk_instance.events, 240).await?;
-            sdk_instance.sdk.sync_wallet(SyncWalletRequest {}).await?;
-            let after = sdk_instance
-                .sdk
-                .get_info(GetInfoRequest {
-                    ensure_synced: Some(false),
-                })
-                .await?;
+
+            let claim_start = Instant::now();
+            let claim_timeout = Duration::from_secs(240);
+            let balance_before = info.balance_sats;
+            let after = loop {
+                sdk_instance.sdk.sync_wallet(SyncWalletRequest {}).await?;
+                let snap = sdk_instance
+                    .sdk
+                    .get_info(GetInfoRequest {
+                        ensure_synced: Some(false),
+                    })
+                    .await?;
+                if snap.balance_sats > balance_before {
+                    break snap;
+                }
+                if claim_start.elapsed() >= claim_timeout {
+                    bail!(
+                        "Timeout waiting for top-up to land (balance still {} sats)",
+                        snap.balance_sats
+                    );
+                }
+                tokio::time::sleep(Duration::from_secs(2)).await;
+            };
             if after.balance_sats >= min_required {
                 info!("Funded after top-up: {} sats", after.balance_sats);
                 return Ok(());
--- a/crates/breez-sdk/breez-bench/src/bin/concurrent_perf.rs
+++ b/crates/breez-sdk/breez-bench/src/bin/concurrent_perf.rs
@@ -17,11 +17,11 @@ use breez_sdk_spark::{
     BreezSdk, GetInfoRequest, MysqlConnectionPool, Network, PostgresConnectionPool,
     PrepareSendPaymentRequest, ReceivePaymentMethod, ReceivePaymentRequest, SdkEvent,
     SendPaymentRequest, SyncWalletRequest, create_mysql_connection_pool,
-    create_postgres_connection_pool, default_config, default_mysql_storage_config,
-    default_postgres_storage_config,
+    create_postgres_connection_pool, default_mysql_storage_config, default_postgres_storage_config,
+    default_server_config,
 };
 
-use breez_bench::events::{wait_for_claimed_event, wait_for_synced_event};
+use breez_bench::events::wait_for_claimed_event;
 use breez_bench::stats::DurationStats;
 
 #[derive(Parser, Debug)]
@@ -197,7 +197,7 @@ async fn main() -> Result<()> {
     );
 
     info!("Initializing sender and receiver SDKs...");
-    let (mut sender, mut receiver, sender_seed, sender_backend) = initialize_sdk_pair(
+    let (mut sender, receiver, sender_seed, sender_backend) = initialize_sdk_pair(
         args.no_auto_optimize,
         args.pre_optimize,
         args.sender_postgres.clone(),
@@ -206,11 +206,6 @@ async fn main() -> Result<()> {
         args.receiver_mysql.clone(),
     )
     .await?;
-
-    info!("Waiting for sender sync...");
-    wait_for_synced_event(&mut sender.events, 120).await?;
-    info!("Waiting for receiver sync...");
-    wait_for_synced_event(&mut receiver.events, 120).await?;
 
     info!(
         "Funding sender with {} sats (need {} sats min)...",
@@ -255,10 +250,6 @@ async fn main() -> Result<()> {
             )
             .await?;
             sender_instances.push(extra);
-        }
-        for (i, inst) in sender_instances.iter_mut().enumerate() {
-            info!("Waiting for sender instance {} initial sync", i);
-            wait_for_synced_event(&mut inst.events, 120).await?;
         }
     }
 
@@ -677,7 +668,7 @@ async fn initialize_sdk_pair(
     let mut sender_seed = [0u8; 32];
     rand::thread_rng().fill_bytes(&mut sender_seed);
 
-    let mut sender_config = default_config(Network::Regtest);
+    let mut sender_config = default_server_config(Network::Regtest);
     if no_auto_optimize || pre_optimize.is_some() {
         sender_config.optimization_config.auto_enabled = false;
     }
@@ -702,8 +693,7 @@ async fn initialize_sdk_pair(
     let mut receiver_seed = [0u8; 32];
     rand::thread_rng().fill_bytes(&mut receiver_seed);
 
-    let mut receiver_config = default_config(Network::Regtest);
-    receiver_config.optimization_config.auto_enabled = false;
+    let receiver_config = default_server_config(Network::Regtest);
     let itest_receiver = build_sdk_with_tree_store_config(
         receiver_path,
         receiver_seed,
@@ -743,7 +733,7 @@ async fn build_extra_sender(
         .tempdir()?;
     let path = dir.path().to_string_lossy().to_string();
 
-    let mut config = default_config(Network::Regtest);
+    let mut config = default_server_config(Network::Regtest);
     if no_auto_optimize || pre_optimize.is_some() {
         config.optimization_config.auto_enabled = false;
     }
@@ -846,15 +836,29 @@ async fn fund_via_faucet(
         let txid = faucet.fund_address(&deposit_address, chunk).await?;
         info!("Faucet chunk #{} txid: {}", chunk_idx, txid);
 
-        wait_for_claimed_event(&mut sdk_instance.events, 240).await?;
-
-        sdk_instance.sdk.sync_wallet(SyncWalletRequest {}).await?;
-        let after = sdk_instance
-            .sdk
-            .get_info(GetInfoRequest {
-                ensure_synced: Some(false),
-            })
-            .await?;
+        let claim_start = Instant::now();
+        let claim_timeout = Duration::from_secs(240);
+        let balance_before = info.balance_sats;
+        let after = loop {
+            sdk_instance.sdk.sync_wallet(SyncWalletRequest {}).await?;
+            let snap = sdk_instance
+                .sdk
+                .get_info(GetInfoRequest {
+                    ensure_synced: Some(false),
+                })
+                .await?;
+            if snap.balance_sats > balance_before {
+                break snap;
+            }
+            if claim_start.elapsed() >= claim_timeout {
+                bail!(
+                    "Timeout waiting for chunk #{} to land (balance still {} sats)",
+                    chunk_idx,
+                    snap.balance_sats
+                );
+            }
+            tokio::time::sleep(Duration::from_secs(2)).await;
+        };
         info!(
             "Balance after chunk #{}: {} sats",
             chunk_idx, after.balance_sats
--- a/crates/breez-sdk/breez-itest/src/helpers.rs
+++ b/crates/breez-sdk/breez-itest/src/helpers.rs
@@ -1768,7 +1768,6 @@ pub async fn build_sdk_with_postgres_server_mode(
     config.api_key = None;
     config.lnurl_domain = None;
     config.prefer_spark_over_lightning = true;
-    config.sync_interval_secs = 5;
     config.real_time_sync_server_url = None;
     config.optimization_config.auto_enabled = false;
 
@@ -1812,7 +1811,6 @@ pub async fn build_sdk_with_mysql_server_mode(
     config.api_key = None;
     config.lnurl_domain = None;
     config.prefer_spark_over_lightning = true;
-    config.sync_interval_secs = 5;
     config.real_time_sync_server_url = None;
     config.optimization_config.auto_enabled = false;
 
--- a/crates/breez-sdk/breez-itest/src/helpers.rs
+++ b/crates/breez-sdk/breez-itest/src/helpers.rs
@@ -741,11 +741,10 @@ pub async fn build_sdk_with_tree_store_config(
     let event_listener = Box::new(ChannelEventListener { tx });
     let _listener_id = sdk.add_event_listener(event_listener).await;
 
-    // Ensure initial sync completes
-    let _ = sdk
-        .get_info(GetInfoRequest {
-            ensure_synced: Some(true),
-        })
+    // Force an initial full sync. Works in both modes (auto sync on or off)
+    // because `sync_wallet` drives the coordinator directly rather than
+    // waiting on the auto-sync initial-synced watcher.
+    sdk.sync_wallet(breez_sdk_spark::SyncWalletRequest {})
         .await?;
 
     Ok(SdkInstance {
--- a/crates/breez-sdk/breez-itest/src/helpers.rs
+++ b/crates/breez-sdk/breez-itest/src/helpers.rs
@@ -1100,6 +1100,42 @@ async fn ensure_funded_inner(sdk_instance: &mut SdkInstance, min_balance: u64) -
     Ok(())
 }
 
+/// Server-mode variant of [`ensure_funded`].
+///
+/// Identical contract, but routes through `receive_and_fund(.., must_be_claimer=false)`
+/// so the wait loop never depends on a `ClaimedDeposits` SDK event. Server-mode
+/// SDKs don't run the spark-wallet `BackgroundProcessor`, so the event would
+/// never fire on its own; `wait_for_balance` (which polls `sync_wallet` +
+/// `get_info`) is the only mechanism that works in both modes.
+pub async fn ensure_funded_via_polling(
+    sdk_instance: &mut SdkInstance,
+    min_balance: u64,
+) -> Result<()> {
+    let span = sdk_instance.span.clone();
+    return ensure_funded_via_polling_inner(sdk_instance, min_balance)
+        .instrument(span)
+        .await;
+}
+
+async fn ensure_funded_via_polling_inner(
+    sdk_instance: &mut SdkInstance,
+    min_balance: u64,
+) -> Result<()> {
+    sdk_instance.sdk.sync_wallet(SyncWalletRequest {}).await?;
+    let info = sdk_instance
+        .sdk
+        .get_info(GetInfoRequest {
+            ensure_synced: Some(false),
+        })
+        .await?;
+    if info.balance_sats < min_balance {
+        let needed = min_balance - info.balance_sats;
+        info!("Funding wallet via faucet (polling): need {} sats", needed);
+        receive_and_fund(sdk_instance, needed.clamp(10000, 50000), false).await?;
+    }
+    Ok(())
+}
+
 /// Get a deposit address and fund it from the faucet in one operation
 ///
 /// This helper generates a deposit address, funds it, and waits for the claim event.
@@ -1705,6 +1741,98 @@ pub async fn build_sdk_with_mysql(
             ensure_synced: Some(true),
         })
         .await?;
+
+    Ok(SdkInstance {
+        sdk,
+        events: rx,
+        span: tracing::Span::current(),
+        temp_dir: None,
+        data_sync_fixture: None,
+        lnurl_fixture: None,
+    })
+}
+
+/// Server-mode variant of [`build_sdk_with_postgres`].
+///
+/// Same wiring (shared PG pool, shared seed, regtest, no LNURL/RT-sync), but
+/// the config comes from `default_server_config` so the SDK runs with
+/// `background_tasks_enabled=false`: no periodic-sync loop, no spark-wallet
+/// `BackgroundProcessor`, no event-driven payment refresh. Callers must drive
+/// every reconciliation via `sync_wallet` (or use polling helpers like
+/// [`ensure_funded_via_polling`] / [`wait_for_balance`]).
+pub async fn build_sdk_with_postgres_server_mode(
+    connection_string: &str,
+    seed_bytes: [u8; 32],
+) -> Result<SdkInstance> {
+    let mut config = breez_sdk_spark::default_server_config(breez_sdk_spark::Network::Regtest);
+    config.api_key = None;
+    config.lnurl_domain = None;
+    config.prefer_spark_over_lightning = true;
+    config.sync_interval_secs = 5;
+    config.real_time_sync_server_url = None;
+    config.optimization_config.auto_enabled = false;
+
+    let seed = breez_sdk_spark::Seed::Entropy(seed_bytes.to_vec());
+
+    let postgres_config =
+        breez_sdk_spark::default_postgres_storage_config(connection_string.to_string());
+    let pool = breez_sdk_spark::create_postgres_connection_pool(&postgres_config)?;
+
+    let sdk = breez_sdk_spark::SdkBuilder::new(config, seed)
+        .with_postgres_connection_pool(pool)
+        .build()
+        .await?;
+
+    let (tx, rx) = mpsc::channel(100);
+    let event_listener = Box::new(ChannelEventListener { tx });
+    let _listener_id = sdk.add_event_listener(event_listener).await;
+
+    // `ensure_synced=true` is a no-op in server mode; issue an explicit
+    // sync_wallet so the initial tree-store hydrate completes before
+    // returning.
+    sdk.sync_wallet(SyncWalletRequest {}).await?;
+
+    Ok(SdkInstance {
+        sdk,
+        events: rx,
+        span: tracing::Span::current(),
+        temp_dir: None,
+        data_sync_fixture: None,
+        lnurl_fixture: None,
+    })
+}
+
+/// Server-mode variant of [`build_sdk_with_mysql`]. See
+/// [`build_sdk_with_postgres_server_mode`] for the rationale.
+pub async fn build_sdk_with_mysql_server_mode(
+    connection_string: &str,
+    seed_bytes: [u8; 32],
+) -> Result<SdkInstance> {
+    let mut config = breez_sdk_spark::default_server_config(breez_sdk_spark::Network::Regtest);
+    config.api_key = None;
+    config.lnurl_domain = None;
+    config.prefer_spark_over_lightning = true;
+    config.sync_interval_secs = 5;
+    config.real_time_sync_server_url = None;
+    config.optimization_config.auto_enabled = false;
+
+    let seed = breez_sdk_spark::Seed::Entropy(seed_bytes.to_vec());
+
+    let mut mysql_config =
+        breez_sdk_spark::default_mysql_storage_config(connection_string.to_string());
+    mysql_config.max_pool_size = 30;
+    let pool = breez_sdk_spark::create_mysql_connection_pool(&mysql_config)?;
+
+    let sdk = breez_sdk_spark::SdkBuilder::new(config, seed)
+        .with_mysql_connection_pool(pool)
+        .build()
+        .await?;
+
+    let (tx, rx) = mpsc::channel(100);
+    let event_listener = Box::new(ChannelEventListener { tx });
+    let _listener_id = sdk.add_event_listener(event_listener).await;
+
+    sdk.sync_wallet(SyncWalletRequest {}).await?;
 
     Ok(SdkInstance {
         sdk,
--- a/crates/breez-sdk/breez-itest/src/helpers.rs
+++ b/crates/breez-sdk/breez-itest/src/helpers.rs
@@ -1786,9 +1786,9 @@ pub async fn build_sdk_with_postgres_server_mode(
     let event_listener = Box::new(ChannelEventListener { tx });
     let _listener_id = sdk.add_event_listener(event_listener).await;
 
-    // `ensure_synced=true` is a no-op in server mode; issue an explicit
-    // sync_wallet so the initial tree-store hydrate completes before
-    // returning.
+    // `ensure_synced=true` is rejected when `background_tasks_enabled` is
+    // false; issue an explicit sync_wallet so the initial tree-store hydrate
+    // completes before returning.
     sdk.sync_wallet(SyncWalletRequest {}).await?;
 
     Ok(SdkInstance {
--- a/crates/breez-sdk/core/src/events.rs
+++ b/crates/breez-sdk/core/src/events.rs
@@ -10,7 +10,7 @@ use tokio::sync::{Mutex, RwLock, broadcast};
 use tracing::info;
 use uuid::Uuid;
 
-use crate::{DepositInfo, LightningAddressInfo, Payment};
+use crate::{DepositInfo, LightningAddressInfo, Payment, sdk::RuntimeEvent};
 
 /// Events emitted by the SDK
 #[allow(clippy::large_enum_variant)]
@@ -45,15 +45,6 @@ pub enum SdkEvent {
     NewDeposits {
         new_deposits: Vec<DepositInfo>,
     },
-}
-
-/// Internal runtime events consumed by the selected runtime profile.
-///
-/// These events are not forwarded to external SDK listeners; they are only used
-/// to decouple non-runtime-owned modules from runtime-specific background behavior.
-#[derive(Debug, Clone)]
-pub(crate) enum RuntimeEvent {
-    StableBalanceConversionCompleted,
 }
 
 impl SdkEvent {
--- a/crates/breez-sdk/core/src/events.rs
+++ b/crates/breez-sdk/core/src/events.rs
@@ -6,7 +6,7 @@ use std::{
 
 use platform_utils::time::Instant;
 use serde::Serialize;
-use tokio::sync::{Mutex, RwLock, broadcast};
+use tokio::sync::{Mutex, RwLock};
 use tracing::info;
 use uuid::Uuid;
 
@@ -185,7 +185,7 @@ pub struct EventEmitter {
     has_real_time_sync: bool,
     rtsync_failed: AtomicBool,
     listener_index: AtomicU64,
-    runtime_event_sender: broadcast::Sender<RuntimeEvent>,
+    runtime_event_handlers: RwLock<Vec<Box<dyn RuntimeEventHandler>>>,
     /// Internal listeners see ALL events before middleware processing
     internal_listeners: RwLock<BTreeMap<String, Box<dyn EventListener>>>,
     /// Middleware chain that can transform/suppress events
@@ -195,15 +195,19 @@ pub struct EventEmitter {
     synced_event_buffer: Mutex<Option<InternalSyncedEvent>>,
 }
 
+#[macros::async_trait]
+pub(crate) trait RuntimeEventHandler: Send + Sync {
+    async fn handle(&self, event: RuntimeEvent);
+}
+
 impl EventEmitter {
     /// Create a new event emitter
     pub fn new(has_real_time_sync: bool) -> Self {
-        let (runtime_event_sender, _) = broadcast::channel(256);
         Self {
             has_real_time_sync,
             rtsync_failed: AtomicBool::new(false),
             listener_index: AtomicU64::new(0),
-            runtime_event_sender,
+            runtime_event_handlers: RwLock::new(Vec::new()),
             internal_listeners: RwLock::new(BTreeMap::new()),
             middleware: RwLock::new(Vec::new()),
             external_listeners: RwLock::new(BTreeMap::new()),
@@ -268,12 +272,16 @@ impl EventEmitter {
         mw.push(middleware);
     }
 
-    pub(crate) fn subscribe_runtime_events(&self) -> broadcast::Receiver<RuntimeEvent> {
-        self.runtime_event_sender.subscribe()
+    pub(crate) async fn add_runtime_event_handler(&self, handler: Box<dyn RuntimeEventHandler>) {
+        let mut handlers = self.runtime_event_handlers.write().await;
+        handlers.push(handler);
     }
 
-    pub(crate) fn emit_runtime_event(&self, event: RuntimeEvent) {
-        let _ = self.runtime_event_sender.send(event);
+    pub(crate) async fn emit_runtime_event(&self, event: RuntimeEvent) {
+        let handlers = self.runtime_event_handlers.read().await;
+        for handler in handlers.iter() {
+            handler.handle(event.clone()).await;
+        }
     }
 
     /// Emit an event through the three-phase pipeline:
--- a/crates/breez-sdk/core/src/events.rs
+++ b/crates/breez-sdk/core/src/events.rs
@@ -6,7 +6,7 @@ use std::{
 
 use platform_utils::time::Instant;
 use serde::Serialize;
-use tokio::sync::{Mutex, RwLock};
+use tokio::sync::{Mutex, RwLock, broadcast};
 use tracing::info;
 use uuid::Uuid;
 
@@ -45,6 +45,15 @@ pub enum SdkEvent {
     NewDeposits {
         new_deposits: Vec<DepositInfo>,
     },
+}
+
+/// Internal runtime events consumed by the selected runtime profile.
+///
+/// These events are not forwarded to external SDK listeners; they are only used
+/// to decouple non-runtime-owned modules from runtime-specific background behavior.
+#[derive(Debug, Clone)]
+pub(crate) enum RuntimeEvent {
+    StableBalanceConversionCompleted,
 }
 
 impl SdkEvent {
@@ -185,6 +194,7 @@ pub struct EventEmitter {
     has_real_time_sync: bool,
     rtsync_failed: AtomicBool,
     listener_index: AtomicU64,
+    runtime_event_sender: broadcast::Sender<RuntimeEvent>,
     /// Internal listeners see ALL events before middleware processing
     internal_listeners: RwLock<BTreeMap<String, Box<dyn EventListener>>>,
     /// Middleware chain that can transform/suppress events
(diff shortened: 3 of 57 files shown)


crates/breez-sdk/core/src/sdk/deposits.rs
```diff
@@ -44,9 +44,6 @@ impl BreezSdk {
                 self.storage
                     .delete_deposit(detailed_utxo.txid.to_string(), detailed_utxo.vout)
                     .await?;
-                self.sync_coordinator
-                    .trigger_sync_no_wait(SyncType::WalletState, true)
-                    .await;
```
dangeross, line 49: Does removing the sync here effect the balance update in client mode? Do we maybe need to emit PaymentSucceeded so its picked up by the BalanceWatcher?

roeierez: Good point. I am trying to minimize the sync also in client mode but updating the balance indeed is needed. PaymentSucceed indeed sound like it was missing from the first place. And it will do the job here.

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+
+If you're building a mobile or desktop wallet, stay on the default ([client mode](initializing.md)) — server mode disables features your app relies on.
+
+## Selecting server mode
+
+Build the config with {{#name default_server_config}} instead of {{#name default_config}}:
+
+{{#tabs sdk_building:init-sdk-server}}
+
+`default_server_config` returns the same `Config` as `default_config` with [`background_tasks_enabled`](./config.md#background-tasks-enabled) set to `false`. All other fields are unchanged; the runtime profile prevents the background services from being started even when their per-field configuration (e.g. [`real_time_sync_server_url`](./config.md#real-time-sync-server-url), [`optimization_config.auto_enabled`](./config.md#optimization-configuration)) is left at its default.
```
dangeross:
```suggestion
{{#name default_server_config}} returns the same `Config` as {{#name default_server_config}} with [{{#name background_tasks_enabled}}](./config.md#background-tasks-enabled) set to `false`. All other fields are unchanged; the runtime profile prevents the background services from being started even when their per-field configuration (e.g. [{{#name real_time_sync_server_url}}](./config.md#real-time-sync-server-url), [{{#name optimization_config.auto_enabled}}](./config.md#optimization-configuration)) is left at its default.
```

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+
+Server mode usually pairs with **shared infrastructure** across SDK instances. See [Customizing the SDK](customizing.md) and the [Shared infrastructure](#shared-infrastructure) section below for the exact wiring.
+
+## What server mode turns off
+
+None of the following per-instance background work is started when `background_tasks_enabled` is `false`:
```
dangeross:
```suggestion
None of the following per-instance background work is started when {{#name background_tasks_enabled}} is `false`:
```

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+
+- **Periodic sync loop** — the SDK does not auto-sync with the Spark network.
+- **Real-time sync client** — no WebSocket subscription to the [real-time sync server](./config.md#real-time-sync-server-url).
+- **Spark wallet background processor** — no operator-event subscription, leaf optimizer, or token-output optimizer.
+- **Lightning-address recovery** — the SDK does not refresh the registered lightning address on startup.
+- **Spark private-mode init** — the [`private_enabled_default`](./config.md#private-mode-enabled-by-default) preset is **not** applied automatically on first startup; you must opt in once via {{#name update_user_settings}} (see [User settings](user_settings.md)).
```
dangeross:
```suggestion
- **Spark private-mode init** — the [{{#name private_enabled_default}}](./config.md#private-mode-enabled-by-default) preset is **not** applied automatically on first startup; you must opt in once via {{#name update_user_settings}} (see [User settings](user_settings.md)).
```

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+- **Flashnet conversion refunder** — no periodic refund pass for failed token conversions.
+- **Stable Balance** — Stable Balance is not supported in server mode because its conversion worker is a background service. Do not rely on automatic Bitcoin-to-token conversion on receive, activation/deactivation conversion, or other Stable Balance background behavior in this profile.
+
+Explicit operations such as `sync_wallet`, `claim_deposit`, `list_unclaimed_deposits`, `refund_deposit`, and `refund_pending_conversions` continue to work and are the intended entry points in this mode.
```
dangeross:
```suggestion
Explicit operations such as {{#name sync_wallet}}, {{#name claim_deposit}}, {{#name list_unclaimed_deposits}}, {{#name refund_deposit}}, and {{#name refund_pending_conversions}} continue to work and are the intended entry points in this mode.
```

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+Because nothing runs in the background, the user is responsible for calling the operations that the SDK would otherwise schedule itself. In practice there are only three things to drive, plus one one-time setup per wallet:
+
+### Sync
+
+Call {{#name sync_wallet}} **only when an external event tells you the wallet state has changed**. The two common cases:
+
+1. **A webhook fires for an incoming payment** — a Lightning receive completes, an on-chain deposit confirms, an incoming Spark transfer lands. Run `sync_wallet()` from the webhook handler so the wallet picks up the new state before downstream consumers (balance reads, payment lists, etc.) need it.
+2. **You explicitly need to reconcile state** — e.g. a periodic reconciliation job for a specific wallet, or a manual admin action. This is rare in practice; the webhook path covers the steady state.
+
+**Do not** call `sync_wallet` from user-facing request handlers (e.g. a `GET /balance` endpoint) as a precaution — it's a network round-trip to operators and is not needed if your webhooks are wired up. {{#name get_info}} reads from the local tree store directly and is the right primitive for read paths.
```
dangeross:
```suggestion
**Do not** call {{#name sync_wallet}} from user-facing request handlers (e.g. a `GET /balance` endpoint) as a precaution — it's a network round-trip to operators and is not needed if your webhooks are wired up. {{#name get_info}} reads from the local tree store directly and is the right primitive for read paths.
```

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+
+The {{#enum SdkEvent::Synced}} event pattern documented in [Listening to events](events.md) is **not available** in server mode — the SDK has no background subscriber to emit it. Treat `sync_wallet` as the synchronous primitive instead.
```
dangeross:
```suggestion
The {{#enum SdkEvent::Synced}} event pattern documented in [Listening to events](events.md) is **not available** in server mode — the SDK has no background subscriber to emit it. Treat {{#name sync_wallet}} as the synchronous primitive instead.
```

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Claiming on-chain deposits
+
+Server-mode SDKs do not run the periodic deposit detection and claim sweep that the mobile profile uses. When your webhook or chain watcher observes a relevant on-chain deposit, handle it explicitly:
+
+- Call {{#name sync_wallet}} to run the SDK's deposit sync and automatic claim logic using your configured [`max_deposit_claim_fee`](./config.md#max-deposit-claim-fee).
```
dangeross:
```suggestion
- Call {{#name sync_wallet}} to run the SDK's deposit sync and automatic claim logic using your configured [{{#name max_deposit_claim_fee}}](./config.md#max-deposit-claim-fee).
```

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+- If your backend already knows the deposit outpoint and wants to drive a specific claim, call {{#name claim_deposit}} for that `txid`/`vout`.
+
+The standard claim flow documented in [Claiming on-chain deposits](onchain_claims.md) applies.
+
+### Stable Balance
+
+Stable Balance is not available in server mode. The feature depends on the client runtime's background conversion worker, so server-mode SDKs will not automatically convert received Bitcoin to the active stable token and will not process Stable Balance activation/deactivation conversions in the background.
+
+If [`stable_balance_config`](./config.md#stable-balance-configuration) is set while using server mode, SDK initialization fails with an invalid input error. Explicit token conversion flows used by payment APIs can still be used, but do not configure Stable Balance for a server-mode deployment.
```
dangeross:
```suggestion
If [{{#name stable_balance_config}}](./config.md#stable-balance-configuration) is set while using server mode, SDK initialization fails with an invalid input error. Explicit token conversion flows used by payment APIs can still be used, but do not configure Stable Balance for a server-mode deployment.
```

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -587,6 +587,13 @@ pub struct Config {
     ///
     /// If set to true, the Spark private mode will be enabled on the first initialization of the SDK.
     /// If set to false, no changes will be made to the Spark private mode.
+    ///
+    /// This default is only auto-applied when
+    /// [`background_tasks_enabled`](Self::background_tasks_enabled) is `true`.
+    /// In server mode the SDK does not touch the Spark private mode on
+    /// startup; partners must call
+    /// `update_user_settings({ spark_private_mode_enabled: Some(...) })`
+    /// explicitly on a one-time setup pass.
```
JssDWt: I wonder how this comment exposes to other languages, with the references and rust-specific language (same in other places). 

Also "partner" I don't think is an appropriate term here (also used in other places).

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Token conversion refunds
+
+**Only relevant if your deployment uses [token conversions](token_conversion.md).** If you don't issue or convert tokens, skip this section.
+
+The flashnet conversion refunder doesn't run in the background in server mode. If you do use tokens, run {{#name refund_pending_conversions}} from your own periodic scheduler (once per minute is a reasonable default) so failed conversions are refunded promptly.
+
+### One-time setup: Spark private mode
+
+The client-mode SDK applies [`private_enabled_default`](./config.md#private-mode-enabled-by-default) on first startup. Server-mode SDKs do not — each per-request SDK would otherwise pay a redundant storage read to check the flag. At provisioning time (when a new wallet is first registered), call {{#name update_user_settings}} with `spark_private_mode_enabled: true`. See [User settings](user_settings.md).
```
dangeross:
```suggestion
The client-mode SDK applies [{{#name private_enabled_default}}](./config.md#private-mode-enabled-by-default) on first startup. Server-mode SDKs do not — each per-request SDK would otherwise pay a redundant storage read to check the flag. At provisioning time (when a new wallet is first registered), call {{#name update_user_settings}} with {{#name spark_private_mode_enabled}} set to `true`. See [User settings](user_settings.md).
```

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Event delivery via webhooks
+
+Without the background processor, the SDK doesn't emit `PaymentSucceeded` / `PaymentPending` / `ClaimedDeposits` events from operator activity. Deliver those signals through webhooks at your own infrastructure instead:
+
+- [Managing webhooks](webhooks.md) describes the supported event types and registration flow.
+- [Lightning Address payment notifications](lnurl_webhooks.md) covers the LNURL server's webhook for incoming LNURL payments.
+
+A typical pipeline: webhook arrives → webhook handler builds a per-request SDK, calls `sync_wallet()` or the relevant explicit operation (e.g. `claim_deposit`), disconnects.
```
dangeross:
```suggestion
A typical pipeline: webhook arrives → webhook handler builds a per-request SDK, calls {{#name sync_wallet}} or the relevant explicit operation (e.g. {{#name claim_deposit}}), disconnects.
```

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Lifecycle pattern
+
+There are three distinct shapes for a server-mode interaction, depending on what triggered it.
+
+### User-facing request handlers
+
+Generate an invoice, send a payment, list history, etc. **Do not call `sync_wallet()` here** — operations that read from local storage (`get_info`, `list_payments`, etc.) do not need a defensive sync, and a network round-trip to operators on every request adds latency without changing the answer.
```
dangeross:
```suggestion
Generate an invoice, send a payment, list history, etc. **Do not call {{#name sync_wallet}} here** — operations that read from local storage ({{#name get_info}}, {{#name list_payments}}, etc.) do not need a defensive sync, and a network round-trip to operators on every request adds latency without changing the answer.
```

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+    response out
+```
+
+{{#tabs sdk_building:server-mode-request-handler}}
+
+### Webhook handlers and reconciliation jobs
+
+Anything driven by an external signal that the wallet state changed. The exact operation depends on the trigger — they're not chained together in the same handler.
+
+- **Incoming Lightning / Spark transfer webhook** — call `sync_wallet()` so downstream reads see the new payment:
```
dangeross:
```suggestion
- **Incoming Lightning / Spark transfer webhook** — call {{#name sync_wallet}} so downstream reads see the new payment:
```

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+
+```text
+    webhook in  →  build SDK  →  sync_wallet()  →  disconnect()
+```
+
+- **On-chain deposit webhook** (or chain watcher) — call `sync_wallet()` to run the deposit sync and automatic claim sweep, or call `claim_deposit` if you want to claim a known outpoint explicitly:
```
dangeross:
```suggestion
- **On-chain deposit webhook** (or chain watcher) — call {{#name sync_wallet}} to run the deposit sync and automatic claim sweep, or call {{#name claim_deposit}} if you want to claim a known outpoint explicitly:
```

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+      ↓
+    disconnect()
+```
+
+{{#tabs sdk_building:server-mode-provisioning}}
+
+### A few notes
+
+- **Building is cheap when infrastructure is shared.** With the shared chain service, MySQL/Postgres pool, and SSP/Connection Managers configured ([see below](#shared-infrastructure)), each per-request SDK reuses HTTP/2 connections, DB pool slots, and gRPC channels — there's no per-request handshake to operators.
+- **Always disconnect.** Even though no background loops are running, `disconnect()` flushes outstanding storage writes and is the documented lifecycle exit. See [Disconnecting](initializing.md#disconnecting).
```
dangeross:
```suggestion
- **Always disconnect.** Even though no background loops are running, calling {{#name disconnect}} flushes outstanding storage writes and is the documented lifecycle exit. See [Disconnecting](initializing.md#disconnecting).
```

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+
+A server-mode deployment normally pairs the profile with shared resources across every per-request SDK. Each of the following is documented in [Customizing the SDK](customizing.md):
+
+- [PostgreSQL Connection Pool](customizing.md#with-postgres-connection-pool) — shared DB pool for storage, tree, and token stores.
+- [MySQL Connection Pool](customizing.md#with-mysql-connection-pool) — same for MySQL.
+- [Shared REST Chain Service](customizing.md#with-shared-rest-chain-service) — one pooled HTTP client instead of one per SDK.
+- [SSP Connection Manager](customizing.md#with-ssp-connection-manager) — share the SSP HTTP client across SDKs.
+- [Connection Manager](customizing.md#with-connection-manager) — share gRPC channels to the Spark operators across SDKs.
+
+Pair `default_server_config` with all of these shared resources — sharing the DB pool, chain service, SSP HTTP client, and gRPC channels across SDKs is the intended deployment shape.
```
dangeross:
```suggestion
Pair {{#name default_server_config}} with all of these shared resources — sharing the DB pool, chain service, SSP HTTP client, and gRPC channels across SDKs is the intended deployment shape.
```

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Driving the `background_tasks_enabled` field directly
```
dangeross:
```suggestion
## Driving the {{#name background_tasks_enabled}} field directly
```

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+
+`default_server_config` is the recommended entry point. If you need to flip the flag on an existing config built another way, see [Background tasks enabled](./config.md#background-tasks-enabled).
```
dangeross:
```suggestion
{{#name default_server_config}} is the recommended entry point. If you need to flip the flag on an existing config built another way, see [Background tasks enabled](./config.md#background-tasks-enabled).
```

docs/breez-sdk/src/SUMMARY.md
```diff
@@ -52,6 +52,7 @@
   - [Conditional Payments](guide/htlcs.md)
   - [Using an External Signer](guide/external_signer.md)
   - [Managing webhooks](guide/webhooks.md)
+  - [Server mode](guide/server_mode.md)
```
dangeross, line 55: Also needs adding to the `guide/advanced.md` page

docs/breez-sdk/src/guide/config.md
```diff
@@ -29,6 +29,23 @@ The synchronization process is used to detect some payment status updates that a
 
 A shorter synchronization interval provides more responsive detection of payment updates but increases resource usage and may trigger API rate limits. The default interval balances responsiveness with resource efficiency for most use cases.
 
+## Background tasks enabled
+
+Master switch for all per-instance background tasks. Defaults to `true`, which is the right choice for mobile and single-instance deployments — the SDK runs its periodic sync, real-time sync client, lightning-address recovery, spark private-mode init, leaf and token-output optimizers, the spark-wallet background processor, and the flashnet conversion refunder.
+
+Set to `false` for multi-tenant server deployments where the SDK is built per request and the host orchestrates sync, claiming, and event delivery (typically via webhooks) explicitly. No background work is started; explicit operations such as `sync_wallet`, `claim_deposit`, `list_unclaimed_deposits`, `refund_deposit`, and `refund_pending_conversions` continue to work and are the intended entry points in this mode.
```
dangeross:
```suggestion
Set to `false` for multi-tenant server deployments where the SDK is built per request and the host orchestrates sync, claiming, and event delivery (typically via webhooks) explicitly. No background work is started; explicit operations such as {{#name sync_wallet}}, {{#name claim_deposit}}, {{#name list_unclaimed_deposits}}, {{#name refund_deposit}}, and {{#name refund_pending_conversions}} continue to work and are the intended entry points in this mode.
```

docs/breez-sdk/src/guide/config.md
```diff
@@ -29,6 +29,23 @@ The synchronization process is used to detect some payment status updates that a
...
+
+The recommended way to opt into server mode is via {{#name default_server_config}}, which returns the same `Config` as {{#name default_config}} with this flag flipped off. See [Server mode](./server_mode.md) for the full profile, lifecycle pattern, and shared-infrastructure wiring. Configuring this field directly is supported if you build your `Config` another way:
+
+{{#tabs config:config-background-tasks}}
+
+<div class="warning">
+<h4>Developer note</h4>
+
+When this flag is `false`, related per-field options such as [`real_time_sync_server_url`](#real-time-sync-server-url) and [`optimization_config.auto_enabled`](#optimization-configuration) retain their configured values but the corresponding background services are not started. Their values stay visible on the `Config` so your code reads what it set, but they have no runtime effect in server mode.
```
dangeross:
```suggestion
When this flag is `false`, related per-field options such as [{{#name real_time_sync_server_url}}](#real-time-sync-server-url) and [{{#name optimization_config.auto_enabled}}](#optimization-configuration) retain their configured values but the corresponding background services are not started. Their values stay visible on the `Config` so your code reads what it set, but they have no runtime effect in server mode.
```

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -613,6 +620,34 @@ pub struct Config {
...
+    /// `list_unclaimed_deposits`, `refund_deposit`,
+    /// `refund_pending_conversions`, leaf/token optimization, etc.) work in
+    /// both modes.
+    ///
+    /// Stable Balance is not supported when this is `false`, because its
+    /// conversion worker is a background service.
+    ///
+    /// The SDK reads this once while building its internal runtime profile.
+    /// Feature code should use the runtime's sync/state/lifecycle services
+    /// rather than branching on this field directly.
```
JssDWt: I don't think this comment is meant to be externalized

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -1025,6 +1060,12 @@ pub struct Credentials {
 #[derive(Debug, Clone)]
 #[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
 pub struct GetInfoRequest {
+    /// In client mode, when `Some(true)` the call waits for the initial Full
+    /// sync to complete before returning.
+    ///
+    /// In server mode this flag has no effect — `get_info` reads balance
```
JssDWt: client mode and server mode don't exist. There is only a flag `background_tasks_enabled`

crates/breez-sdk/core/src/sdk/runtime/server.rs
```diff
@@ -0,0 +1,54 @@
...
+        _request: GetInfoRequest,
+    ) -> Result<GetInfoResponse, SdkError> {
+        let balance_sats = sdk.spark_wallet.get_balance().await?;
+        let token_balances = sdk
+            .spark_wallet
+            .get_token_balances()
+            .await?
+            .into_iter()
+            .map(|(k, v)| (k, v.into()))
+            .collect();
```
JssDWt: These could go in parallel

hydra-yse (approved):
Besides @dangeross's comment, looks good

crates/breez-sdk/core/src/sdk/runtime/server.rs
```diff
@@ -0,0 +1,54 @@
...
+        let request = SyncRequest::fire_and_forget(sync_type, force);
+        sdk.sync_wallet_internal(&request).await
+    }
+
+    async fn get_info(
+        &self,
+        sdk: &BreezSdk,
+        _request: GetInfoRequest,
+    ) -> Result<GetInfoResponse, SdkError> {
+        let balance_sats = sdk.spark_wallet.get_balance().await?;
```
JssDWt: I wonder whether we could optimize this call by computing the balance in the database. (can do that another time)

dangeross (commented):
Docs read well, just some small nits to make it less Rusty

crates/breez-sdk/core/src/sdk/runtime/client.rs
```diff
@@ -0,0 +1,418 @@
...
+    let span = tracing::Span::current();
+
+    tokio::spawn(
+        async move {
+            event_emitter
+                .add_internal_listener(Box::new(listener))
+                .await;
+        }
+        .instrument(span),
+    );
```
JssDWt: Does this have to be spawned? Adding the listener is a very short operation.

crates/breez-sdk/core/src/events.rs
```diff
@@ -47,6 +47,15 @@ pub enum SdkEvent {
...
 
+/// Internal runtime events consumed by the selected runtime profile.
+///
+/// These events are not forwarded to external SDK listeners; they are only used
+/// to decouple non-runtime-owned modules from runtime-specific background behavior.
+#[derive(Debug, Clone)]
+pub(crate) enum RuntimeEvent {
+    StableBalanceConversionCompleted,
+}
+
```
dangeross: Does it belong more in the runtime module?

JssDWt (commented):
I'm in the middle of review. I'll commit these comments now because I'm distracted at the moment.

crates/breez-sdk/core/src/sdk/runtime/server.rs
roeierez: I think we do that in postgres/mysql right?

crates/breez-sdk/breez-itest/src/helpers.rs
```diff
@@ -1716,3 +1751,95 @@ pub async fn build_sdk_with_mysql(
...
+/// [`ensure_funded_via_polling`] / [`wait_for_balance`]).
+pub async fn build_sdk_with_postgres_server_mode(
+    connection_string: &str,
+    seed_bytes: [u8; 32],
+) -> Result<SdkInstance> {
+    let mut config = breez_sdk_spark::default_server_config(breez_sdk_spark::Network::Regtest);
+    config.api_key = None;
+    config.lnurl_domain = None;
+    config.prefer_spark_over_lightning = true;
+    config.sync_interval_secs = 5;
```
danielgranhao: nit: better not to set it because it is ignored
```suggestion
```

crates/breez-sdk/breez-itest/src/helpers.rs
```diff
@@ -1716,3 +1751,95 @@ pub async fn build_sdk_with_mysql(
...
+/// [`build_sdk_with_postgres_server_mode`] for the rationale.
+pub async fn build_sdk_with_mysql_server_mode(
+    connection_string: &str,
+    seed_bytes: [u8; 32],
+) -> Result<SdkInstance> {
+    let mut config = breez_sdk_spark::default_server_config(breez_sdk_spark::Network::Regtest);
+    config.api_key = None;
+    config.lnurl_domain = None;
+    config.prefer_spark_over_lightning = true;
+    config.sync_interval_secs = 5;
```
danielgranhao:
```suggestion
```

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -613,6 +620,34 @@ pub struct Config {
...
+    /// `refund_pending_conversions`, leaf/token optimization, etc.) work in
+    /// both modes.
+    ///
+    /// Stable Balance is not supported when this is `false`, because its
+    /// conversion worker is a background service.
+    ///
+    /// The SDK reads this once while building its internal runtime profile.
+    /// Feature code should use the runtime's sync/state/lifecycle services
+    /// rather than branching on this field directly.
+    pub background_tasks_enabled: bool,
```
danielgranhao, line 648: I don't love that several config values are silently ignored when this is false. I guess it's possible this already is the best/simplest compromise, but here are a couple of other options to consider:
- make potentially ignored values optional and validated at runtime that they are None when `background_tasks_enabled` is false
- replace this with an optional `BackgroundTasksConfig` type that includes the values that are only used when there are background tasks. Setting None disables background tasks

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -1025,6 +1060,12 @@ pub struct Credentials {
 #[derive(Debug, Clone)]
 #[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
 pub struct GetInfoRequest {
+    /// In client mode, when `Some(true)` the call waits for the initial Full
+    /// sync to complete before returning.
+    ///
+    /// In server mode this flag has no effect — `get_info` reads balance
+    /// directly from the spark wallet on every call, so the response is
+    /// always fresh.
```
danielgranhao: Would it be preferable to return an error when get_info is called with `ensure_synced` true and there are no background tasks?

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+
+Stable Balance is not available in server mode. The feature depends on the client runtime's background conversion worker, so server-mode SDKs will not automatically convert received Bitcoin to the active stable token and will not process Stable Balance activation/deactivation conversions in the background.
+
+If [{{#name stable_balance_config}}](./config.md#stable-balance-configuration) is set while using server mode, SDK initialization fails with an invalid input error. Explicit token conversion flows used by payment APIs can still be used, but do not configure Stable Balance for a server-mode deployment.
+
+### Token conversion refunds
+
+**Only relevant if your deployment uses [token conversions](token_conversion.md).** If you don't issue or convert tokens, skip this section.
+
+The flashnet conversion refunder doesn't run in the background in server mode. If you do use tokens, run {{#name refund_pending_conversions}} from your own periodic scheduler (once per minute is a reasonable default) so failed conversions are refunded promptly.
```
danielgranhao: is the idea that the server would call `refund_pending_conversions` every minute for all wallets all the time?

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+
+```text
+    new wallet registered
+      ↓
+    build SDK (default_server_config + shared infra)
+      ↓
+    update_user_settings({ spark_private_mode_enabled: true })
+      ↓
+    disconnect()
+```
```
danielgranhao, line 139: Is this needed? Currently uses rust names. Maybe the snippet below is enough?

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -991,4 +992,41 @@ mod tests {
             );
         }
     }
+
+    // tokio::test requires `rt`, which the workspace only enables on
+    // non-wasm targets. Gate this test (and its imports) accordingly so
+    // `cargo clippy --target wasm32-unknown-unknown --all-targets` stays
+    // clean.
+    #[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
+    #[tokio::test]
```
danielgranhao: nit: Can we use the `async_test_all` macro instead? or `async_test_not_wasm` if there's a reason not to test on wasm?

danielgranhao (commented):
Also still mid-review. Mainly went over API/docs for now

crates/breez-sdk/core/src/sdk/runtime/server.rs
```diff
@@ -0,0 +1,54 @@
...
+            .collect();
+
+        Ok(GetInfoResponse {
+            identity_pubkey: sdk.spark_wallet.get_identity_public_key().to_string(),
+            balance_sats,
+            token_balances,
+        })
+    }
+
+    async fn ensure_spark_private_mode_initialized(&self, _sdk: &BreezSdk) -> Result<(), SdkError> {
```
danielgranhao: nit: could we rename method to clarify it doesn't always ensure?
```suggestion
    async fn maybe_ensure_spark_private_mode_initialized(&self, _sdk: &BreezSdk) -> Result<(), SdkError> {
```

crates/breez-sdk/core/src/sdk/payments.rs
```diff
@@ -823,16 +837,6 @@ impl BreezSdk {
...
-            ConversionType::FromBitcoin
-        ) {
-            self.sync_coordinator
-                .trigger_sync_no_wait(SyncType::WalletState, true)
-                .await;
-        }
-
         // Wait for the received conversion payment to complete
         let payment = self
             .wait_for_payment(
```
danielgranhao, line 838: IIUC this will always time out in server mode. There will be no background processor processing token tx events, so the incoming token payment will not be detected

crates/breez-sdk/core/src/sdk/deposits.rs
roeierez: In client mode we have the TransferredClaim event actually that trigers balance update so I think it is not an issue.

docs/breez-sdk/src/guide/server_mode.md
roeierez: Addressed

docs/breez-sdk/src/guide/server_mode.md
roeierez: Addressed

docs/breez-sdk/src/guide/server_mode.md
roeierez: Addressed

docs/breez-sdk/src/guide/server_mode.md
roeierez: Addressed

docs/breez-sdk/src/guide/server_mode.md
roeierez: Addressed

docs/breez-sdk/src/guide/server_mode.md
roeierez: Addressed

docs/breez-sdk/src/guide/server_mode.md
roeierez: Addressed

docs/breez-sdk/src/guide/server_mode.md
roeierez: Addressed

docs/breez-sdk/src/guide/server_mode.md
roeierez: Addressed

docs/breez-sdk/src/guide/server_mode.md
roeierez: Addressed

docs/breez-sdk/src/guide/server_mode.md
roeierez: Addressed

docs/breez-sdk/src/guide/server_mode.md
roeierez: Addressed

docs/breez-sdk/src/guide/server_mode.md
roeierez: Addressed

docs/breez-sdk/src/guide/server_mode.md
roeierez: Addressed

docs/breez-sdk/src/guide/server_mode.md
roeierez: Addressed

docs/breez-sdk/src/guide/server_mode.md
roeierez: Addressed

docs/breez-sdk/src/SUMMARY.md
roeierez: Addressed

docs/breez-sdk/src/guide/config.md
roeierez: Addressed

docs/breez-sdk/src/guide/config.md
roeierez: Addressed

crates/breez-sdk/core/src/events.rs
roeierez: Addressed

crates/breez-sdk/core/src/sdk/deposits.rs
dangeross: Just a lag is all, calling get_info won't reflect the actual balance until the event updates it

crates/breez-sdk/core/src/models/mod.rs
roeierez: Fixed

crates/breez-sdk/core/src/models/mod.rs
roeierez: Fixed

crates/breez-sdk/core/src/models/mod.rs
roeierez: Right, Fixed

crates/breez-sdk/core/src/sdk/runtime/server.rs
roeierez: Fixed

crates/breez-sdk/core/src/sdk/runtime/server.rs
roeierez: Already like this. using a query for db storages.

crates/breez-sdk/core/src/sdk/runtime/client.rs
roeierez: Fixed, made the init_sdk async.

crates/breez-sdk/breez-itest/src/helpers.rs
roeierez: Removed

crates/breez-sdk/breez-itest/src/helpers.rs
roeierez: Removed

docs/breez-sdk/src/guide/server_mode.md
roeierez: Not practical indeed. I changed the comment which in general says it is the user responsibility to track pending payments and run this on wallets in risk.

docs/breez-sdk/src/guide/server_mode.md
roeierez: I find it helpful but I did change the rust names to more pseudo.

crates/breez-sdk/core/src/sdk_builder.rs
roeierez: sure, fixed.

crates/breez-sdk/core/src/sdk/deposits.rs
roeierez: True! Better to trigger event. Then client runtime syncs in respond.

roeierez: Done

crates/breez-sdk/core/src/models/mod.rs
roeierez: I agree. Now return error.

crates/breez-sdk/core/src/models/mod.rs
roeierez: @danielgranhao  I agree not ideal but not sure I want to change the interface atm.
How about when background_tasks_enabled=false return error if: 
1. `real_time_sync_server_url` is not None.
2. `optimization_config.auto_enabled`=true
3. `stable_balance_config` is not None (already the case).

crates/breez-sdk/core/src/sdk/runtime/server.rs
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    async fn run_user_sync(
+        &self,
+        sdk: &BreezSdk,
+        sync_type: SyncType,
+        force: bool,
+    ) -> Result<(), SdkError> {
+        let request = SyncRequest::fire_and_forget(sync_type, force);
+        sdk.sync_wallet_internal(&request).await
```
JssDWt: I suppose it's possible that users call sync multiple times in parallel. I think it's not made for that yet. Should we have mechanisms in place to cascade syncs, rather than having them run in parallel?

JssDWt: I find the sync_wallet_internal call here out-of-place. It's especially the `reply` field on the `SyncRequest` that appears to have a purpose, but is not used inside sync_wallet_internal at all. 

Together with my other comment, I wonder whether this call should go through the sync coordinator instead?

JssDWt (commented):
Comments round 2!

crates/breez-sdk/breez-bench/src/bin/concurrent_perf.rs
```diff
@@ -846,15 +836,29 @@ async fn fund_via_faucet(
...
-        let after = sdk_instance
-            .sdk
-            .get_info(GetInfoRequest {
-                ensure_synced: Some(false),
-            })
-            .await?;
+        let claim_start = Instant::now();
+        let claim_timeout = Duration::from_secs(240);
+        let balance_before = info.balance_sats;
+        let after = loop {
```
danielgranhao, line 842: This main funding loop moved to `sync_wallet` + balance polling, but the final top-up fallback (~line 903) still calls `wait_for_claimed_event(.., 240)` - on server mode this will block and fail in the end if we ever got there

crates/breez-sdk/core/src/models/mod.rs
roeierez: Done

crates/breez-sdk/core/src/sdk/runtime/server.rs
roeierez: Yes it is possible but unlikely as in server mode instances are built on demand and disconnects immediately after doing some work (typically sync). The mechanism on the client we adde is mostly to mitigate our internal sync requests. I think it is OK to say it is the user responsibility and no harm is done if multiple requests are invoked in parallel.

crates/breez-sdk/core/src/sdk/deposits.rs
```diff
@@ -44,8 +47,7 @@ impl BreezSdk {
                 self.storage
                     .delete_deposit(detailed_utxo.txid.to_string(), detailed_utxo.vout)
                     .await?;
-                self.sync_coordinator
-                    .trigger_sync_no_wait(SyncType::WalletState, true)
+                get_payment_and_emit_event(&self.storage, &self.event_emitter, payment.clone())
```
danielgranhao: In client mode, won't this potentially result in duplicate events due to background event handling also calling `get_payment_and_emit_event`?

crates/breez-sdk/core/src/sdk/runtime/server.rs
roeierez: That's a good point. I agree the reply is not related to here so I moved out the reply from  the sync_wallet_internal to the client runtime where it belongs.
I do think on server mode we should attempt to keep things simple and not consolidate attempts to sync like we do on the client.

crates/breez-sdk/core/src/sdk/runtime/server.rs
roeierez: Done

crates/breez-sdk/core/src/sdk/payments.rs
roeierez: As discussed, will wait for a fix from Ross.

crates/breez-sdk/breez-bench/src/bin/concurrent_perf.rs
roeierez: Fixed

crates/breez-sdk/core/src/sdk/deposits.rs
roeierez: Used internal runtime event here client/server runtimes react differently so no duplicate event.

crates/breez-sdk/core/src/sdk/mod.rs
```diff
@@ -202,9 +206,50 @@ pub fn default_config(network: Network) -> Config {
...
+///
+/// One-time setup that the SDK normally applies automatically — notably
+/// `private_enabled_default` — is NOT applied in this mode. Drive setup
+/// explicitly via `update_user_settings` (and any other relevant APIs) so
+/// ephemeral per-request SDK instances incur no implicit setup overhead.
+///
+/// `get_info` reads balance directly from the spark wallet in this mode
+/// rather than from the background-maintained storage cache, so balance is
+/// always fresh and `ensure_synced=true` is rejected with an invalid-input
+/// error.
```
danielgranhao: "always fresh" is IMO misleading
```suggestion
/// rather than from the background-maintained storage cache, so balance 
/// reflects the latest local sync and `ensure_synced=true` is rejected with 
/// an invalid-input error
```

danielgranhao (approved):
LGTM - just a rustdoc suggestion

docs/breez-sdk/src/guide/server_mode.md
```diff
@@ -0,0 +1,165 @@
...
+
+Server mode usually pairs with **shared infrastructure** across SDK instances. See [Customizing the SDK](customizing.md) and the [Shared infrastructure](#shared-infrastructure) section below for the exact wiring.
+
+## What server mode turns off
+
+None of the following per-instance background work is started when {{#name background_tasks_enabled}} is `false`:
+
+- **Periodic sync loop** — the SDK does not auto-sync with the Spark network.
+- **Real-time sync client** — no WebSocket subscription to the [real-time sync server](./config.md#real-time-sync-server-url).
+- **Spark wallet background processor** — no operator-event subscription, leaf optimizer, or token-output optimizer.
```
dangeross, line 32: Do we need a section below about driving the leaf/token-output optimization? Do we need an API to expose running the token-output optimizer?

roeierez: We have the start_leaf_optimization API. Do you think we need anything else?

dangeross: An API to run the token-output optimization, but it doesn't necessarily need to be in this PR

roeierez: Yes let's iterate on that on a separate PR.

JssDWt (approved):
I think we're good.

I want to flag the (almost) cyclic function calls from sdk to client back to sdk. I like call trees to be cleanly down only. That could get a refactor, because the ClientRuntime technically doesn't need the sdk. It only uses underlying functions, which happen to be on sdk right now, but probably shouldn't be.

!! Note CI is failing on both format and the Breez integration tests (needs fixing) !!

dangeross (approved):
LGTM

PR #886, opened 2026-05-18
isolate jwt + expose context

JssDWt:
Introduces `SdkContext` — a single opaque handle bundling every process-shareable resource — and consolidates the SDK's HTTP/gRPC clients and DB pools behind it.

- **`SdkContext` + `new_shared_sdk_context(SdkContextConfig)`**: opaque, factory-built, passed to `SdkBuilder::with_shared_context(...)`. Auto-defaulted in `build()` when omitted.
- **Consolidates inside the context**:
  - One `Arc<dyn HttpClient>` reused for SSP GraphQL, chain service, LNURL, LNURL server, JWT refresh, Flashnet.
  - One `Arc<BreezServer>` (Breez backend gRPC).
  - One `Arc<dyn ConnectionManager>` (Spark operator gRPC) — directly the spark-wallet type, no more SDK-side wrapper.
  - One `Transport` for the realtime sync server (default URL); per-SDK fallback only if the URL is overridden.
  - Optional `Arc<PostgresConnectionPool>` / `Arc<MysqlConnectionPool>` built from `PostgresStorageConfig` / `MysqlStorageConfig`.
- **Removed builder methods**: `with_ssp_connection_manager`, `with_connection_manager`, `with_session_manager` (kept `#[doc(hidden)]` for WASM JS-storage bridging).
- **Retained for backward continuity**: `with_postgres_connection_pool` / `with_mysql_connection_pool` (+ deprecated `with_*_backend`); conflict with a context-supplied pool returns `SdkError::Generic("multiple … pools configured")`.
- **Deferred `with_rest_chain_service`**: now stores params on the builder and constructs the service in `build()` using `context.http_client`.
- **JWT provider** (`jwt_header_provider.rs`): self-managed background refresh task (drop-on-shutdown via `oneshot`), exponential backoff with 5-min cap, in-memory cache stores `(token, exp)` so `headers()` never re-parses the JWT.

- Removed `SspConnectionManager`, `new_ssp_connection_manager`, the SDK-level `ConnectionManager` wrapper, `new_connection_manager`, and the corresponding Flutter/WASM wrappers + `connection_manager.rs` modules across all bindings.

Note: Will conflict with #885 a bit, but that should be easily rebased.


packages/flutter/rust/src/sdk_builder.rs
```diff
@@ -64,19 +64,16 @@ impl SdkBuilder {
...
-        let builder = <breez_sdk_spark::SdkBuilder as Clone>::clone(&self.inner)
-            .with_ssp_connection_manager(manager.inner);
-        Self {
-            inner: Arc::new(builder),
-        }
-    }
-
-    #[frb(sync)]
-    pub fn with_connection_manager(self, connection_manager: &ConnectionManager) -> Self {
+    pub fn with_context(self, context: &SdkContext) -> Self {
```
roeierez: NIT: we need a naming that will help users understand when they would need this. perhaps with_shared_context? And also SdkSharedContext as parameter?

roeierez (commented):
Will review soon. I like it. Dropped a nit that jumped to my eyes.

docs/breez-sdk/src/guide/customizing.md
```diff
@@ -136,46 +135,30 @@ By implementing the Payment Observer interface you can be notified before a paym
...
-    <a class="header" href="#with-connection-manager">With Connection Manager</a>
-    <a class="tag" target="_blank" href="https://breez.github.io/spark-sdk/breez_sdk_spark/struct.SdkBuilder.html#method.with_connection_manager">API docs</a>
-</h2>
-
-A Connection Manager owns the gRPC channels to the Spark operators. By default each SDK instance builds its own. Server processes hosting many wallets at once can share a single Connection Manager between every SDK, so they reuse the same channels instead of each opening a fresh set.
-
-Construct one via {{#name new_connection_manager}} and pass it to each {{#name SdkBuilder}} via {{#name with_connection_manager}}. Connections close when the last reference to the Connection Manager is dropped; calling {{#name disconnect}} on an SDK instance does not affect them.
-
-The `connections_per_operator` argument controls how many connections the manager opens to each operator:
+The `connections_per_operator` setting on {{#name SdkContextConfig}} controls how many gRPC connections the context opens to each Spark operator:
```
dangeross:
```suggestion
The {{#name connections_per_operator}} setting on {{#name SdkContextConfig}} controls how many gRPC connections the context opens to each Spark operator:
```

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -271,6 +291,17 @@ impl SdkBuilder {
         Ok(self.with_mysql_connection_pool(pool))
     }
 
+    /// Internal: plumb a pre-built session manager into the builder. Used by
+    /// the WASM bindings to inject a JS-side DB-backed session manager. Not
+    /// part of the public Rust API surface — Rust integrators should rely on
+    /// the session manager derived from the [`SdkContext`]'s DB pool.
+    #[doc(hidden)]
```
dangeross: I guess a tradeoff instead of going down the `#[cfg(target_family = "wasm")]` route and a cascade of clippy errors

packages/flutter/rust/src/sdk_builder.rs
JssDWt: Done. I left SdkContext as is, but changed to `new_shared_sdk_context` and `with_shared_context`.

crates/breez-sdk/core/src/sdk_builder.rs
JssDWt: `#[cfg(target_family = "wasm")]` works fine. Updated.

docs/breez-sdk/src/guide/customizing.md
JssDWt: Done. Also made an attempt to improve claude in this regard.

crates/breez-sdk/core/src/sdk_context.rs
```diff
@@ -0,0 +1,137 @@
...
+    #[cfg(feature = "postgres")]
+    #[cfg_attr(feature = "uniffi", uniffi(default = None))]
+    pub postgres_config: Option<PostgresStorageConfig>,
+
+    /// `MySQL` backend configuration. When set, the context builds a shared
+    /// connection pool and SDKs constructed with this context store their
+    /// data in `MySQL`.
+    #[cfg(feature = "mysql")]
+    #[cfg_attr(feature = "uniffi", uniffi(default = None))]
+    pub mysql_config: Option<MysqlStorageConfig>,
```
roeierez, line 92: perhaps either enum or validate that both not set?

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -818,12 +845,27 @@ impl SdkBuilder {
         let inner_session_manager: Arc<dyn spark_wallet::SessionManager> = Arc::new(
             crate::session_manager::CachingSessionManager::new(inner_session_manager),
         );
-        let partner_headers = Arc::new(BreezPartnerHeaderProvider::new());
+        let jwt_header_provider: Option<Arc<BreezJwtHeaderProvider>> =
+            if matches!(self.config.network, Network::Mainnet)
+                && let Some(api_key) = self.config.api_key.clone()
+            {
+                Some(BreezJwtHeaderProvider::new(
```
roeierez: Doesn't this mean every sdk instance has its own provider? Own refresh task? Isn't this suppose to be shared as part of the SdkContext with a single task globally?

roeierez (changes requested):

crates/breez-sdk/core/src/sdk_builder.rs
JssDWt: Right, that was the whole point of the PR wasn't it, lol. Fixed

crates/breez-sdk/core/src/sdk_context.rs
JssDWt: Do we want to enforce use of shared postgres or mysql? If it is not set here, we use normal storage resolution as before in the sdk.

roeierez: That's seems fine to me , only concerned about if both are set together. 
Ah.. but that was the behavior before you are saying? If that's the case then it doesn't have to be in this PR.

roeierez (approved):
LGTM

crates/breez-sdk/core/src/sdk_context.rs
JssDWt: Yes it will then fail on build just like before.

JssDWt: Added an additional check for multiple storage configurations so we don't initialize any pool

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -826,12 +850,15 @@ impl SdkBuilder {
...
             spark_wallet::WalletBuilder::new(spark_wallet_config, spark_signer)
                 .with_cancellation_token(shutdown_sender.subscribe())
-                .with_session_manager(inner_session_manager)
-                .with_so_extra_header_provider(partner_headers.clone());
+                .with_session_manager(inner_session_manager);
+        if let Some(provider) = &context.jwt_header_provider {
+            wallet_builder = wallet_builder.with_so_extra_header_provider(
+                Arc::clone(provider) as Arc<dyn spark_wallet::HeaderProvider>
+            );
+        }
```
dangeross, line 891: I'm wondering if we need to validate that the context's jwt_header_provider is provided if it's Mainnet/API key set. Otherwise its just skipped over

JssDWt: We control initialization of the context, so the mainnet + api key check is done in the constructor there. I did add an additional check that api key + network must be the same as the config during build.

dangeross (approved):
LGTM

PR #887, opened 2026-05-19
Dedup payment events and don't override terminal state

roeierez:
We are emitting events from various sources (sync, spark events, ssp).
These are not always in sync so we have to ensure:
1. We don't emit duplicate events (enforced by the storage for atomicity).
2. We don't override completed/failed payments with payments that are not in terminal state (final state).

I believe this also solve a flakiness in our integration tests such as:  https://github.com/breez/spark-sdk/actions/runs/26083908033/job/76692046270

### DIFF
--- a/crates/breez-sdk/core/src/persist/mod.rs
+++ b/crates/breez-sdk/core/src/persist/mod.rs
@@ -269,6 +269,13 @@ pub struct PaymentMetadata {
     pub conversion_status: Option<ConversionStatus>,
 }
 
+#[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
+pub(crate) fn parse_payment_status(value: &str) -> Result<PaymentStatus, StorageError> {
+    value
+        .parse()
+        .map_err(|e| StorageError::Implementation(format!("invalid payment status: {e}")))
+}
+
 /// Trait for persistent storage
 #[cfg_attr(feature = "uniffi", uniffi::export(with_foreign))]
 #[async_trait]
@@ -290,16 +297,15 @@ pub trait Storage: Send + Sync {
         request: StorageListPaymentsRequest,
     ) -> Result<Vec<Payment>, StorageError>;
 
-    /// Inserts a payment into storage
+    /// Inserts or updates a payment unless it would replace a terminal status.
     ///
-    /// # Arguments
+    /// Same-status updates are still persisted so details can be enriched.
     ///
-    /// * `payment` - The payment to insert
-    ///
-    /// # Returns
-    ///
-    /// Success or a `StorageError`
-    async fn insert_payment(&self, payment: Payment) -> Result<(), StorageError>;
+    /// Returns `true` if the caller should emit a payment event (new payment was
+    /// inserted, or status transitioned). Returns `false` for redundant
+    /// same-status updates and for rejected updates against a terminal stored
+    /// status
+    async fn apply_payment_update(&self, payment: Payment) -> Result<bool, StorageError>;
 
     /// Inserts payment metadata into storage
     ///
--- a/crates/breez-sdk/core/src/persist/mysql/storage.rs
+++ b/crates/breez-sdk/core/src/persist/mysql/storage.rs
@@ -669,7 +669,7 @@ impl MysqlStorage {
         engine.input(identity);
         engine.input(payment_id.as_bytes());
         let digest = sha256::Hash::from_engine(engine);
-        format!("brz_payment_{}", &digest.to_string()[..32])
+        digest.to_string()[..32].to_string()
     }
 
     async fn get_payment_status_in_tx(
--- a/crates/breez-sdk/core/src/persist/mysql/storage.rs
+++ b/crates/breez-sdk/core/src/persist/mysql/storage.rs
@@ -6,7 +6,7 @@
 
 use std::collections::HashMap;
 
-use bitcoin::hashes::{Hash, sha256};
+use bitcoin::hashes::{Hash, HashEngine, sha256};
 use macros::async_trait;
 use mysql_async::prelude::*;
 use mysql_async::{Params, Pool, Row, Transaction, Value};
@@ -664,10 +664,11 @@ fn from_json_string_opt<T: serde::de::DeserializeOwned>(
 
 impl MysqlStorage {
     fn payment_update_lock_name(identity: &[u8], payment_id: &str) -> String {
-        let mut lock_key = Vec::with_capacity(identity.len().saturating_add(payment_id.len()));
-        lock_key.extend_from_slice(identity);
-        lock_key.extend_from_slice(payment_id.as_bytes());
-        let digest = sha256::Hash::hash(&lock_key);
+        let mut engine = sha256::Hash::engine();
+        engine.input(b"brz_payment_update");
+        engine.input(identity);
+        engine.input(payment_id.as_bytes());
+        let digest = sha256::Hash::from_engine(engine);
         format!("brz_payment_{}", &digest.to_string()[..32])
     }
 
--- a/crates/breez-sdk/core/src/persist/mysql/storage.rs
+++ b/crates/breez-sdk/core/src/persist/mysql/storage.rs
@@ -6,20 +6,22 @@
 
 use std::collections::HashMap;
 
+use bitcoin::hashes::{Hash, sha256};
 use macros::async_trait;
 use mysql_async::prelude::*;
-use mysql_async::{Params, Pool, Row, Value};
+use mysql_async::{Params, Pool, Row, Transaction, Value};
 use spark_mysql::{mysql_async, tx_opts};
 use tracing::warn;
 
 use crate::{
     AssetFilter, Contact, ConversionDetails, ConversionInfo, ConversionStatus, DepositInfo,
     ListContactsRequest, LnurlPayInfo, LnurlReceiveMetadata, LnurlWithdrawInfo, PaymentDetails,
-    PaymentMethod, SparkHtlcDetails, SparkHtlcStatus,
+    PaymentMethod, PaymentStatus, SparkHtlcDetails, SparkHtlcStatus,
     error::DepositClaimError,
     persist::{
         Payment, PaymentMetadata, SetLnurlMetadataItem, Storage, StorageError,
         StorageListPaymentsRequest, StoragePaymentDetailsFilter, UpdateDepositPayload,
+        parse_payment_status,
     },
     sync_storage::{
         IncomingChange, OutgoingChange, Record, RecordChange, RecordId, UnversionedRecordChange,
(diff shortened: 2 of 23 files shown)


crates/breez-sdk/core/src/persist/tests.rs
```diff
@@ -2629,6 +2653,73 @@ pub async fn test_payment_details_update_persistence(storage: Box<dyn Storage>)
...
+    let stored_payment = storage.get_payment_by_id(payment.id.clone()).await.unwrap();
+    assert_eq!(stored_payment.status, PaymentStatus::Completed);
+    assert_eq!(stored_payment.amount, 15_000);
+
+    let mut conflicting_final = payment.clone();
+    conflicting_final.status = PaymentStatus::Failed;
+    storage
+        .apply_payment_update(conflicting_final)
+        .await
+        .unwrap();
```
dangeross, line 2698: should we assert the payment update outcome?

crates/breez-sdk/core/src/persist/mod.rs
```diff
@@ -290,16 +389,14 @@ pub trait Storage: Send + Sync {
...
-    /// # Returns
-    ///
-    /// Success or a `StorageError`
-    async fn insert_payment(&self, payment: Payment) -> Result<(), StorageError>;
+    /// Same-status updates are still persisted so details can be enriched without
+    /// emitting duplicate payment events.
+    async fn apply_payment_update(
+        &self,
+        payment: Payment,
+    ) -> Result<PaymentUpdateOutcome, StorageError>;
```
danielgranhao: Have you considered just a conditional upsert? Something like:
```
ON CONFLICT ... DO UPDATE ... WHERE NOT (status IN ('completed','failed') AND EXCLUDED.status != status) + RETURNING (xmax = 0)
```

We would enforce the guard atomically, and callers get a bool to know if the status changed.

Seems like it would be simpler. No need for advisory lock, etc.

roeierez: I think I didn't find a way to do it in mysql and wanted the code to work the same on postgress

danielgranhao: I see, indeed not as clean on mysql. Still:
- I think mysql can do it via UPDATE-then-INSERT in one tx - affectedRows from the UPDATE tells you if the guard rejected, and the unique constraint handles the new-row race. No advisory lock needed
- Even if we keep advisory locks, do callers need the 4 variant enum? Seems like it's just used for should_emit_event so we could directly return the bool. 

But as it is, it works, just raising the potential to simplify

roeierez: > * I think mysql can do it via UPDATE-then-INSERT in one tx - affectedRows from the UPDATE tells you if the guard rejected, and the unique constraint handles the new-row race. No advisory lock needed

Not sure... update first means we don't have a way to lock an existing row which means in between this row may be added. Also probably 99% of the cases would succeed on insert so I think it adds the round trip anyway. Also worth mentioning that the lock use the payment id so very highly chances it doesn't create any contention or affect overall payments inserts. 



> * Even if we keep advisory locks, do callers need the 4 variant enum? Seems like it's just used for should_emit_event so we could directly return the bool.

I will take a deeper look and see if we can simplify that, thanks.

roeierez: @danielgranhao thanks for the suggestion. I simplified the storage and got rid of the enum

crates/breez-sdk/wasm/js/mysql-storage/index.cjs
```diff
@@ -360,130 +364,229 @@ class MysqlStorage {
...
-            payment.method ? JSON.stringify(payment.method) : null,
-            withdrawTxId,
-            depositTxId,
-            spark,
-          ]
+      await conn.query("SET TRANSACTION ISOLATION LEVEL READ COMMITTED");
+      await conn.beginTransaction();
+      try {
+        const [rows] = await conn.query(
+          "SELECT status FROM brz_payments WHERE user_id = ? AND id = ?",
```
dangeross: The Rust implementation does a "FOR UPDATE" also here. Should we align for consistency?

crates/breez-sdk/core/src/sdk/payments.rs
```diff
@@ -1542,14 +1544,14 @@ impl BreezSdk {
...
-                            };
-                            tokio::time::sleep(sleep_time).await;
+                                let sleep_time = if i < 5 {
+                                    Duration::from_secs(1)
+                                } else {
+                                    Duration::from_secs(i.into())
+                                };
+                                tokio::time::sleep(sleep_time).await;
+                            }
                         }
```
danielgranhao, line 1555: Is this change intentional? After this change, SSP fetches that return Err/Ok(None) or conversion failures result in the next attempt starting immediately

crates/breez-sdk/core/src/utils/payments.rs
```diff
@@ -1,12 +1,52 @@
...
+    payment: Payment,
+    should_emit: bool,
+) -> bool {
+    if should_emit {
+        get_payment_and_emit_event(storage, event_emitter, payment).await;
+        true
+    } else {
+        false
+    }
+}
```
danielgranhao: nit: could be inlined

crates/breez-sdk/core/src/utils/payments.rs
```diff
@@ -1,12 +1,52 @@
...
+/// Emit a payment event based on whether storage indicated the update should
+/// produce one (new payment inserted or status transitioned).
+pub(crate) async fn emit_payment_update(
+    storage: &Arc<dyn Storage>,
+    event_emitter: &EventEmitter,
+    payment: Payment,
+    should_emit: bool,
+) -> bool {
+    if should_emit {
+        get_payment_and_emit_event(storage, event_emitter, payment).await;
```
danielgranhao: I think here we need to pass the expected payment or payment status and only emit if the payment we get from storage has same status. Otherwise we may still emit duplicate `PaymentSucceeded` events if theres a race for setting pending then succeeded shortly after. 

Also, even after fixing that, I think it may still be possible for the events to get emitted out of order (e.g. PaymentSucceeded followed by PaymentPending). To fix this we could add version/sequence column to payment row and return it when doing an update.

danielgranhao (commented):
Mainly a concern about duplicate events still being possible. Also, CI is failing on fmt and wasm tests

crates/breez-sdk/core/src/sdk/payments.rs
```diff
@@ -1542,14 +1544,14 @@ impl BreezSdk {
...
-                                Duration::from_secs(i.into())
-                            };
-                            tokio::time::sleep(sleep_time).await;
+                                let sleep_time = if i < 5 {
+                                    Duration::from_secs(1)
+                                } else {
+                                    Duration::from_secs(i.into())
+                                };
+                                tokio::time::sleep(sleep_time).await;
+                            }
```
dangeross, line 1547: Did you see an issue here? Better to exhaust the polling fast if there are parsing issues?

crates/breez-sdk/wasm/js/postgres-storage/index.cjs
```diff
@@ -353,130 +355,176 @@ class PostgresStorage {
...
 
+  _paymentUpdateLockKey(paymentId) {
+    return crypto
+      .createHash("sha256")
+      .update("brz_payment_update")
+      .update(this.identity)
+      .update(Buffer.from(paymentId))
+      .digest()
+      .readBigInt64BE(0)
+      .toString();
```
dangeross, line 415: Very minor, mysql and postgres do slightly inconsistent hashing, postgres includes the prefix in the hash, mysql prepends it. Just want to point it out
```
    return `brz_payment_${crypto
      .createHash("sha256")
      .update(lockKey)
      .digest("hex")
      .slice(0, 32)}`;
```

dangeross (commented):
Looking good

crates/breez-sdk/core/src/utils/payments.rs
roeierez: > Also, even after fixing that, I think it may still be possible for the events to get emitted out of order (e.g. PaymentSucceeded followed by PaymentPending). To fix this we could add version/sequence column to payment row and return it when doing an update.

@danielgranhao  How do you see that can happen? We only update the event if it is not in terminal state so if there is already a payment completed in the db update (and emit) should not happen.

danielgranhao: After discussing offline we agreed these issues are possible but should be rare. Given the impact is non critical, we will accept it for now

crates/breez-sdk/core/src/sdk/payments.rs
roeierez: Good catch! rebase issue. fixed.

crates/breez-sdk/core/src/utils/payments.rs
roeierez: Agree. Fixed

crates/breez-sdk/core/src/persist/tests.rs
roeierez: No enum anymore

crates/breez-sdk/wasm/js/mysql-storage/index.cjs
roeierez: fixed

crates/breez-sdk/core/src/sdk/payments.rs
roeierez: That was a rebase issue. Addressed.

crates/breez-sdk/wasm/js/postgres-storage/index.cjs
roeierez: True, made them consistent.

danielgranhao (approved):
LGTM

crates/breez-sdk/wasm/js/mysql-storage/index.cjs
```diff
@@ -360,130 +364,230 @@ class MysqlStorage {
...
+  }
+
+  _paymentUpdateLockName(paymentId) {
+    return `brz_payment_${crypto
+      .createHash("sha256")
+      .update("brz_payment_update")
+      .update(this.identity)
+      .update(Buffer.from(paymentId))
+      .digest("hex")
+      .slice(0, 32)}`;
```
dangeross: Still has the prepend

crates/breez-sdk/core/src/persist/mysql/storage.rs
```diff
@@ -659,6 +662,173 @@ fn from_json_string_opt<T: serde::de::DeserializeOwned>(
...
 }
 
+impl MysqlStorage {
+    fn payment_update_lock_name(identity: &[u8], payment_id: &str) -> String {
+        let mut engine = sha256::Hash::engine();
+        engine.input(b"brz_payment_update");
+        engine.input(identity);
+        engine.input(payment_id.as_bytes());
+        let digest = sha256::Hash::from_engine(engine);
+        format!("brz_payment_{}", &digest.to_string()[..32])
```
dangeross: Same here

crates/breez-sdk/wasm/js/mysql-storage/index.cjs
roeierez: Fixed, sorry.

crates/breez-sdk/core/src/persist/mysql/storage.rs
roeierez: Fixed

dangeross (approved):
LGTM

PR #889, opened 2026-05-19
Refactor `wait_for_payment` to be polling only

dangeross:
In server mode (`background_tasks_enabled=false`), `wait_for_payment` blocked for its full timeout because no spark-wallet event subscription exists to fire `PaymentSucceeded`. This broke:

- `complete_conversion_and_send` — errored with "Timeout waiting for conversion to complete"
- `lnurl_withdraw` — returned `payment: None` after a full timeout

### What this PR does

Replaces the event-listener `wait_for_payment` with a single polling `wait_for_incoming_payment` implementation that works in both modes — operator/SSP query loop with targeted local-store updates on terminal status:

- Spark transfer (`PaymentId` as `TransferId`) → `list_transfers` poll → `SparkWallet::process_transfer` claims the leaves into the tree-store.
- Token tx (`PaymentId` as `{hash}:{vout}`) → `get_token_transactions_by_hashes` poll → `SparkWallet::process_token_transaction` updates the local output store.
- Inbound Lightning (new `LightningReceive { invoice, ssp_id }` variant) → `fetch_lightning_receive_payment` poll → once the SSP populates `transfer_id`, claims the resulting Spark transfer.

Two new `pub` methods on `SparkWallet` (`process_transfer`, `process_token_transaction`) wrap the same non-emitting claim/process layer the background processor already uses, so the polled path and the event-driven path are idempotent under concurrency.

Extracted `BreezSdk::finalize_payment` for the shared post-claim work: apply cached payment metadata, sync LNURL metadata, persist, refresh balances, emit `PaymentSucceeded` on first observation. Called from `wait_for_payment`, `handle_wallet_event::TransferClaimed`, and `process_token_transaction_event`.

`lnurl_withdraw` SSP id is plumbed via an inner `receive_bolt11_invoice` fn and the new `LightningReceive` variant.

Added `--server-mode` to the CLI for manual mainnet validation.

Follow-up to #885 (timeout flagged by @danielgranhao).

### Test plan

- New regtest itest `lightning_send_server_mode.rs` covers the polling path end-to-end.
- New `conversion_send_server_mode.rs` (ignored — needs regtest token liquidity).
- Manual mainnet via `cargo run -p cli -- --server-mode --network mainnet`:
  - sats → token ✓
  - token → sats ✓
  - lnurl-withdraw ✓

### DIFF
--- a/crates/breez-sdk/core/src/sdk/payments.rs
+++ b/crates/breez-sdk/core/src/sdk/payments.rs
@@ -5,8 +5,7 @@ use platform_utils::tokio;
 use spark_wallet::{ExitSpeed, SparkAddress, TransferId, TransferTokenOutput};
 use spark_wallet::{InvoiceDescription, Preimage};
 use std::str::FromStr;
-use tokio::sync::{mpsc, oneshot};
-use tokio::time::timeout;
+use tokio::sync::oneshot;
 use tracing::{Instrument, error, info, warn};
 
 use crate::{
@@ -35,10 +34,7 @@ use crate::{
     },
 };
 
-use super::{
-    BreezSdk,
-    helpers::{InternalEventListener, get_deposit_address, is_payment_match},
-};
+use super::{BreezSdk, helpers::get_deposit_address};
 
 #[cfg_attr(feature = "uniffi", uniffi::export(async_runtime = "tokio"))]
 #[allow(clippy::needless_pass_by_value)]
@@ -837,7 +833,9 @@ impl BreezSdk {
     ) -> Result<SendPaymentResponse, SdkError> {
         // Wait for the received conversion payment to complete
         let payment = self
+            .runtime
             .wait_for_payment(
+                self,
                 WaitForPaymentIdentifier::PaymentId(
                     conversion_response.received_payment_id.clone(),
                 ),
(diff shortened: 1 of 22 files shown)


crates/breez-sdk/core/src/sdk/runtime/server.rs
```diff
@@ -67,6 +84,153 @@ impl RuntimeProfile for ServerRuntime {
...
+    sdk: &BreezSdk,
+    invoice: &str,
+) -> Result<Option<Payment>, SdkError> {
+    // `force=true` bypasses the "synced recently" skip in
+    // sync_wallet_internal — without it, every probe after the first
+    // becomes a no-op inside the cache window and we re-read stale
+    // storage. `Wallet | WalletState` is the minimal scope that runs
+    // claim_pending_transfers and writes Payment records to storage so
+    // get_payment_by_invoice can see the incoming send.
+    sdk.sync_wallet_internal(SyncType::Wallet | SyncType::WalletState, true)
```
roeierez: Do you think better to use spark query_invoices endpoint instead of full sync?

crates/breez-sdk/core/src/sdk/runtime/server.rs
```diff
@@ -67,6 +84,153 @@ impl RuntimeProfile for ServerRuntime {
...
+                    .await;
+            }
+        };
+
+        // PaymentId probes only confirm operator-side status; trigger a
+        // local sync so claim_pending_transfers runs and the new
+        // leaves/outputs land in the tree-store / token-output store
+        // before downstream callers (e.g. the send leg of a
+        // conversion-and-send) try to use them.
+        self.run_user_sync(sdk, SyncType::Wallet, true).await?;
```
roeierez: Why do we need to call run_user_sync on server mode? When payment completes don't we already update the leaves store?

crates/breez-sdk/core/src/sdk/runtime/client.rs
```diff
@@ -91,6 +95,56 @@ impl RuntimeProfile for ClientRuntime {
     ) -> Result<(), SdkError> {
         sdk.ensure_spark_private_mode_initialized_inner().await
     }
+
+    async fn wait_for_payment(
```
roeierez: Did you consider use the same polling implementation used in the server runtime here as well? It seems more reliable there and If it works for the server perhaps it is good here as well?

roeierez (changes requested):

crates/breez-sdk/core/src/utils/polling.rs
```diff
@@ -0,0 +1,84 @@
...
+            Ok(Some(value)) => return Ok(value),
+            Ok(None) => last_error = None,
+            Err(e) => last_error = Some(e),
+        }
+
+        let elapsed = started.elapsed();
+        if elapsed >= schedule.timeout {
+            return Err(last_error
+                .unwrap_or_else(|| SdkError::Generic("Timeout while polling".to_string())));
+        }
```
JssDWt: I think it makes sense to put this check after the sleep, because
- The sleep time will be 0 then anyway
- You wouldn't want to probe again after the final sleep? (because the timeout is hit)

crates/breez-sdk/core/src/sdk/runtime/server.rs
dangeross: The payment doesn't complete in server mode. wait_for_payment polls list_transfers, if the transfer is there then runs the sync to claim/refresh leaves. We could also manually claim and add leaves to avoid the sync. For token receives we need to refresh or add the token outputs to the store. Both of these are needed to fulfill the ongoing payment

crates/breez-sdk/core/src/sdk/runtime/server.rs
dangeross: query_spark_invoices exists, but its for spark invoices. There is a fetch_lightning_receive_payment but it keyed to the ssp_id not the invoice

crates/breez-sdk/core/src/sdk/runtime/client.rs
dangeross: I wanted to start with just fixing the server mode runtime, they discuss using it for both if it makes sense

roeierez: Makes sense.

crates/breez-sdk/core/src/sdk/payments.rs
```diff
@@ -1414,64 +1437,57 @@ impl BreezSdk {
         Ok(SendPaymentResponse { payment })
     }
 
-    pub(super) async fn wait_for_payment(
+    pub(crate) async fn wait_for_payment(
```
roeierez: NIT: I think now we can rename this to wait_for_incoming_payment perhaps?

roeierez: Not in the context of this PR, but it’s worth thinking about whether we want to expose this function. It seems useful for some use cases we’ve run into in the past.

roeierez (changes requested):

crates/breez-sdk/core/src/sdk/payments.rs
```diff
@@ -24,21 +26,27 @@ use crate::{
...
 use super::{
     BreezSdk,
-    helpers::{InternalEventListener, get_deposit_address, is_payment_match},
+    helpers::{get_deposit_address, maybe_get_payment_from_storage, update_balances},
 };
+use crate::sync::SparkSyncService;
+
+// Polling cadence for wait_for_incoming_payment.
+const WAIT_FOR_INCOMING_PAYMENT_INITIAL_DELAY_MS: u64 = 500;
+const WAIT_FOR_INCOMING_PAYMENT_MAX_DELAY_MS: u64 = 2000;
```
roeierez, line 49: Perhaps we should use a shorter max delay, at least for the first minutes where most payments should be finished. Just to get a better response time.

roeierez (changes requested):

roeierez (approved):

danielgranhao (approved):
LGTM

PR #890, opened 2026-05-19
make new_shared_sdk_context async to run on the tokio runtime

danielgranhao:
Building the shared resources transitively tokio::spawns at construction with no .await of our own: BreezServer::new -> Endpoint::connect_lazy() spawns tonic's channel buffer-worker (unconditional), and on mainnet+api_key BreezJwtHeaderProvider::new spawns its refresh task. tokio::spawn aborts outside a runtime, which is exactly the case for a sync UniFFI export called on a foreign thread.

Export it async with async_runtime = "tokio" so UniFFI polls it on the managed runtime, and propagate .await to all callers (builder, cli, itest helpers, bench bins, wasm and flutter bindings). The flutter binding drops #[frb(sync)] accordingly.

### DIFF
--- a/crates/breez-sdk/core/src/sdk_context.rs
+++ b/crates/breez-sdk/core/src/sdk_context.rs
@@ -117,9 +117,11 @@ impl SdkContextConfig {
 /// `SdkContextConfig::new(network)` yields an in-memory, single-tenant setup;
 /// supply a DB config to back the SDKs with a shared `PostgreSQL` or `MySQL`
 /// pool.
-#[cfg_attr(feature = "uniffi", uniffi::export)]
-#[allow(clippy::needless_pass_by_value)]
-pub fn new_shared_sdk_context(config: SdkContextConfig) -> Result<Arc<SdkContext>, SdkError> {
+// Async-on-tokio so UniFFI runs it on the managed runtime: building the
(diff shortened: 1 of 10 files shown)


roeierez (approved):
LGTM

crates/breez-sdk/core/src/sdk_context.rs
```diff
@@ -117,9 +117,11 @@ impl SdkContextConfig {
 /// `SdkContextConfig::new(network)` yields an in-memory, single-tenant setup;
 /// supply a DB config to back the SDKs with a shared `PostgreSQL` or `MySQL`
 /// pool.
-#[cfg_attr(feature = "uniffi", uniffi::export)]
-#[allow(clippy::needless_pass_by_value)]
-pub fn new_shared_sdk_context(config: SdkContextConfig) -> Result<Arc<SdkContext>, SdkError> {
+// Async-on-tokio so UniFFI runs it on the managed runtime: building the
+// shared resources `tokio::spawn`s internally (gRPC channel; mainnet JWT
+// task) and aborts off-runtime, despite no `.await` here.
```
JssDWt, line 122:
```suggestion
```

JssDWt (approved):
Looking good, just have a comment suggestion, because this is exposed to the outside.

PR #895, opened 2026-05-20
rename SessionManager to SessionStore

JssDWt:
The SessionManager doesn't actually manage anthing. It's just a store. Reflect that in the name.


dangeross (approved):
LGTM

roeierez (approved):

PR #896, opened 2026-05-20
core: treat disconnect() as a no-op when no shutdown receivers

danielgranhao:
Server mode (`background_tasks_enabled = false`) has no listener on the shutdown watch channel — `watch::Sender::send` returns Err and `disconnect()` reports "Failed to send shutdown signal" even though there is nothing to stop. Match that case explicitly: log at debug and return Ok(()).

### DIFF
--- a/crates/breez-sdk/core/src/sdk/api.rs
+++ b/crates/breez-sdk/core/src/sdk/api.rs
@@ -1,6 +1,6 @@
 use bitcoin::secp256k1::{PublicKey, ecdsa::Signature};
 use std::str::FromStr;
-use tracing::info;
+use tracing::{debug, info};
 
 use breez_sdk_common::buy::cashapp::CashAppProvider;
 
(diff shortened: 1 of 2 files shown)


roeierez (approved):
LGTM

crates/breez-sdk/core/src/sdk/api.rs
```diff
@@ -60,9 +60,15 @@ impl BreezSdk {
     /// Result containing either success or an `SdkError` if the background task couldn't be stopped
     pub async fn disconnect(&self) -> Result<(), SdkError> {
         info!("Disconnecting Breez SDK");
-        self.shutdown_sender
-            .send(())
-            .map_err(|_| SdkError::Generic("Failed to send shutdown signal".to_string()))?;
+        if self.shutdown_sender.send(()).is_err() {
```
roeierez, line 63: Just checking, is this error always indicate no listeners or we should be more specific here?

danielgranhao: Yes, that's the only failure mode. It doesn't necessarily mean it's expected behavior on our side. If we're on "client mode" and there are no receivers that would be unexepected but I didn't think it was worth only ignoring the err on `background_tasks_enabled = false` - I can add it if you prefer.

roeierez: I am fine with it as it is.

dangeross (approved):
LGTM. Do the docs need adjusting to reflect this for server mode also?

PR #897, opened 2026-05-20
Pin all MySQL timestamps to UTC to fix host-TZ mismatch

dangeross:
The wasm-bindgen test `token_store::tests::mysql::test_remove_token_outputs_prevents_refresh_re_add` (added in #846) fails on environments whose host timezone has a positive UTC offset (e.g. CEST). Output: `left: 2, right: 1` — a spent output is silently re-added by a refresh.

Root cause: `mysql2` defaults to `timezone: 'local'`, which serializes JS `Date` parameters as **host-local** datetime strings. The MySQL server stores them literally in `DATETIME(6)` columns (no TZ awareness) and compares them against `NOW()`, which returns the **session** TZ. When the host and the session disagree, the inequality `spent_at >= refresh_timestamp` flips and the spent marker is missed.

Postgres is unaffected because `TIMESTAMPTZ` carries TZ info. Rust `spark-mysql` tests pass only because the test container happens to be UTC; the same MySQL session bug affects production deployments where the server's `time_zone` isn't UTC, regardless of client.

### Fix

Pin every server-side timestamp to UTC explicitly, and fix the JS-only client-side serialization:

1. **Server-side**: replace every `NOW()` / `NOW(6)` / `DATE_SUB(NOW(6), …)` with `UTC_TIMESTAMP(…)` across JS `mysql-token-store`, `mysql-tree-store`, and Rust `spark-mysql::{token_store,tree_store}`. Make INSERTs that relied on `DEFAULT CURRENT_TIMESTAMP(6)` pass `UTC_TIMESTAMP(6)` explicitly.
2. **Schema migration**: new migration flips column defaults to `DEFAULT (UTC_TIMESTAMP(6))`.
3. **JS client**: `timezone: "Z"` on every `createMysqlPool` (token-store, tree-store, storage) so `Date` parameters are sent as UTC.

After this, every timestamp comparison is UTC↔UTC regardless of host TZ or MySQL session TZ.

### Regression test

New Node-level test in `crates/breez-sdk/wasm/js/mysql-token-store/tests/` that spawns child processes under `TZ=UTC | Pacific/Kiritimati (+14) | Pacific/Pago_Pago (-11) | Europe/Berlin (+2)` and runs the spent-marker scenario in each. Wired into `make wasm-test`, so any future TZ regression fails on standard CI runners (UTC) rather than only on contributor machines.

### DIFF
--- a/crates/breez-sdk/wasm/js/mysql-tree-store/migrations.cjs
+++ b/crates/breez-sdk/wasm/js/mysql-tree-store/migrations.cjs
@@ -367,6 +367,19 @@ class MysqlTreeStoreMigrationManager {
           `ALTER TABLE brz_tree_swap_status ADD PRIMARY KEY (user_id)`,
         ],
       },
+      {
+        // Pin DATETIME defaults to UTC. Server-side INSERTs already pass
+        // `UTC_TIMESTAMP(6)` explicitly; this migration makes the column
+        // default match, so any future callsite that omits the column also
+        // gets a UTC value rather than a session-TZ-dependent one.
+        name: "Pin DATETIME defaults to UTC",
+        sql: [
+          `ALTER TABLE brz_tree_reservations MODIFY COLUMN created_at DATETIME(6) NOT NULL DEFAULT (UTC_TIMESTAMP(6))`,
+          `ALTER TABLE brz_tree_leaves       MODIFY COLUMN created_at DATETIME(6) NOT NULL DEFAULT (UTC_TIMESTAMP(6))`,
+          `ALTER TABLE brz_tree_leaves       MODIFY COLUMN added_at   DATETIME(6) NOT NULL DEFAULT (UTC_TIMESTAMP(6))`,
+          `ALTER TABLE brz_tree_spent_leaves MODIFY COLUMN spent_at   DATETIME(6) NOT NULL DEFAULT (UTC_TIMESTAMP(6))`,
+        ],
+      },
     ];
   }
 }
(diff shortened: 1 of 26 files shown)


crates/breez-sdk/wasm/js/mysql-tree-store/migrations.cjs
```diff
@@ -367,6 +367,19 @@ class MysqlTreeStoreMigrationManager {
...
+        // Pin DATETIME defaults to UTC. Server-side INSERTs already pass
+        // `UTC_TIMESTAMP(6)` explicitly; this migration makes the column
+        // default match, so any future callsite that omits the column also
+        // gets a UTC value rather than a session-TZ-dependent one.
+        name: "Pin DATETIME defaults to UTC",
+        sql: [
+          `ALTER TABLE brz_tree_reservations MODIFY COLUMN created_at DATETIME(6) NOT NULL DEFAULT (UTC_TIMESTAMP(6))`,
+          `ALTER TABLE brz_tree_leaves       MODIFY COLUMN created_at DATETIME(6) NOT NULL DEFAULT (UTC_TIMESTAMP(6))`,
+          `ALTER TABLE brz_tree_leaves       MODIFY COLUMN added_at   DATETIME(6) NOT NULL DEFAULT (UTC_TIMESTAMP(6))`,
+          `ALTER TABLE brz_tree_spent_leaves MODIFY COLUMN spent_at   DATETIME(6) NOT NULL DEFAULT (UTC_TIMESTAMP(6))`,
```
roeierez: We change the default here and existing columns are not changed. I don't think it is a practical problem, just noting here.

crates/breez-sdk/wasm/js/mysql-tree-store/migrations.cjs
roeierez: applied_at also should be handled?

roeierez (changes requested):
Looks good.

danielgranhao (approved):
LGTM

roeierez (approved):

JssDWt:
@dangeross Changes look good. The last commit has some merge conflicts. (I suppose session store rename)

JssDWt (approved):
LGTM

PR #898, opened 2026-05-20
Simplify storage setup

JssDWt:
Storage setup used to be tangled: the SDK builder and context each held
branching logic for every database type. This change replaces that with a
single, clear path.

- **Storage config** — one `StorageConfig` value names which built-in storage
  to use (SQLite, Postgres, or MySQL), built via `default_storage`,
  `postgres_storage`, or `mysql_storage`.
- **Backend resolution** — the builder no longer branches per database. It
  resolves exactly one storage backend in one place. Integrators who bring
  their own storage pass a `CustomStorage` to `with_storage`.
- **SQLite feature** — SQLite is now an opt-in `sqlite` Cargo feature, matching
  `postgres` and `mysql`. None build on WASM, so WASM supplies its own storage.

The result: one clear way to choose storage, no scattered per-database logic,
and a smaller, more predictable API.


crates/breez-sdk/core/src/persist/backend/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+            }
+        }
+    }
+}
+
+/// The single storage source for one `SdkBuilder` — either a built-in backend
+/// config or a caller-supplied store set.
+#[derive(Clone)]
+pub(crate) enum StorageSetup {
+    Config(StorageConfig),
```
roeierez: If we are already introducing breaking changes why not get rid of the StorageConfig variant? So get rid of the enum and the only way is to pass CustomStorage

roeierez (commented):
Still reviewing, like that. Dropped a comment.

crates/breez-sdk/core/src/persist/backend/mod.rs
JssDWt: I was trying to minimize the amount of breaking changes, but I like the idea. I'll try that out.

JssDWt: Done

roeierez (approved):
Looks great!

danielgranhao (commented):
Looks good! I see that the Flutter build has some issues and there are also some rustdocs linking to `crate::create_storage_backend` which should be `custom_storage`

danielgranhao (approved):
LGTM

PR #901, opened 2026-05-21
Add ENV-based spark-itest backend switch

dangeross:
Adds an env-var-driven backend switch to `spark-itest`: setting `USE_POSTGRES_BACKEND=true` or `USE_MYSQL_BACKEND=true` swaps every wallet's `SessionStore`, `TreeStore`, and `TokenOutputStore` to the corresponding SQL implementation, with per-test fresh databases backed by testcontainers. Pattern copied from `breez-itest`'s tree-store switch but generalized: one env var covers every SQL-eligible spark-wallet store.

`make itest` is unchanged (in-memory); `make spark-itest-pg` and `make spark-itest-mysql` exercise the SQL backends. CI's `itest` job is now a 3-way matrix.

### Why

PR #861's V3 dedup-by-id fix (commit `794d3c3f`, "key token output pool by `(prev_tx_hash, vout)` instead of server id") landed **only in the in-memory store**. The four SQL backends still have the bug, but it slips past CI today because `spark-itest`'s only token reproducer (`test_many_outputs`) always builds wallets via `SparkWallet::connect`, which hardcodes `InMemoryTokenOutputStore::default()`. So the canonical end-to-end test never exercises the SQL paths.

This change lets `test_many_outputs` and the rest of `spark-itest` run against the SQL backends, closing that coverage gap and giving a fixture-ready home for the SQL-side fix.

### Bonus fix

The `wallets` rstest fixture had a latent race: it built Alice + Bob before subscribing to either, so the first wallet's `Synced` could fire before `subscribe_events()` ran. In-memory was fast enough to hide it; SQL wallet construction (~500ms with migrations) was slow enough to lose the race deterministically. Fixed by subscribing immediately after each `build_test_wallet`.

### Test plan

- [x] `make itest` — 9/9 pass
- [x] `USE_POSTGRES_BACKEND=true make itest` — 9/9 pass (within ~5% of in-memory time)
- [x] `USE_MYSQL_BACKEND=true make itest` — 9/9 pass (within ~15% of in-memory time)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -42,8 +42,16 @@ jobs:
         run: make fmt-check
 
   itest:
-    name: Integration test
+    name: Integration test (${{ matrix.backend.name }})
     runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.backend.allow-failure }}
+    strategy:
+      fail-fast: false
+      matrix:
+        backend:
+          - { name: in-memory, target: itest,             allow-failure: false }
+          - { name: postgres,  target: spark-itest-pg,    allow-failure: true }
+          - { name: mysql,     target: spark-itest-mysql, allow-failure: true }
     env:
       FAUCET_USERNAME: ${{ secrets.FAUCET_USERNAME }}
       FAUCET_PASSWORD: ${{ secrets.FAUCET_PASSWORD }}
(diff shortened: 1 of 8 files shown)


danielgranhao (approved):
LGTM

roeierez (approved):
Feels a lot safer now!

PR #902, opened 2026-05-22
spark-postgres: set Runtime::Tokio1 on pool builder

danielgranhao:
## Summary

`deadpool::managed::PoolBuilder::build()` synchronously rejects any pool config with a timeout unless `.runtime(...)` was set on the builder — it does not detect an ambient tokio runtime. With our defaults all three timeouts are `None`, so existing tests pass; but any caller setting `wait_timeout_secs`, `create_timeout_secs`, or `recycle_timeout_secs` via the public API hits `Failed to initialize database: Error occurred while building the pool: Timeouts require a runtime` at SDK startup.

Fix: pass `.runtime(deadpool::Runtime::Tokio1)` on all four ssl_mode arms of `create_pool`. `spark-postgres` is already structurally tokio-only (tokio-postgres I/O, UniFFI's tokio runtime), so this is the only correct value.

Added a regression test that builds a pool with `recycle_timeout_secs = Some(300)` — it fails synchronously before the fix, no Postgres server required.

## Test plan

- [x] `cargo test -p spark-postgres --lib pool::`
- [x] `cargo clippy -p spark-postgres --all-targets -- -D warnings`

### DIFF
--- a/crates/spark-postgres/src/pool.rs
+++ b/crates/spark-postgres/src/pool.rs
@@ -303,6 +303,7 @@ pub fn create_pool(config: &PostgresStorageConfig) -> Result<Pool, PostgresError
             let manager = deadpool_postgres::Manager::new(pg_config, tokio_postgres::NoTls);
             Pool::builder(manager)
                 .config(pool_config)
+                .runtime(deadpool::Runtime::Tokio1)
                 .build()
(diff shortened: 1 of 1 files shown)


dangeross (approved):
LGTM

roeierez (approved):

PR #903, opened 2026-05-22
Fix missing contacts store for users who upgraded from 0.10.0

JssDWt:
## Problem

Web wallets that were on SDK version 0.10.0 at some point never had a
"contacts" area created in their local browser database. For those
wallets, saving a contact or opening the contacts list always fails.

This happened because the step that creates the contacts storage was
added in a way that wallets coming from 0.10.0 quietly skipped, so the
storage was never set up for them. The problem is permanent — it does
not fix itself on later updates.

## Solution

Add a recovery step that creates the contacts storage for any wallet
that is missing it, the next time the wallet starts. Wallets that
already have it are unaffected.

The internal schema version was also one number behind, which meant the
recovery step (and one earlier step) would not run. That number is now
corrected, and a startup check was added so this kind of mismatch is
caught immediately in the future.

### DIFF
--- a/crates/breez-sdk/wasm/js/web-storage/index.js
+++ b/crates/breez-sdk/wasm/js/web-storage/index.js
@@ -457,6 +457,35 @@ class MigrationManager {
           }
         }
       },
+      {
+        // Recovery migration for the missing "contacts" object store.
+        //
+        // The original "Create contacts store" migration above was INSERTED into
+        // the middle of this array (at index 12) instead of being appended, in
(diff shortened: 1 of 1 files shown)


dangeross (approved):
LGTM

Maybe not for this PR but we could probably make the migrations array static and derive the current db version (passed to indexedDB.open) from MIGRATIONS.length (or MigrationManager.count()) directly

roeierez (approved):

PR #904, opened 2026-05-22
Server mode: mark SDK as initially synced at startup

dangeross:
In server mode (`background_tasks_enabled = false`), `ServerRuntime::start_sdk_services` discarded the `initial_synced_sender`, leaving the watcher stuck at `false`. `SparkSyncService::sync_payments` uses that watcher to gate `PaymentSucceeded` emission for completed receives — so any host subscribed to payment events would silently miss notifications on every sync.

### DIFF
--- a/crates/breez-sdk/core/src/sdk/runtime/server.rs
+++ b/crates/breez-sdk/core/src/sdk/runtime/server.rs
@@ -1,4 +1,5 @@
 use tokio::sync::watch;
+use tracing::error;
 
 use crate::{GetInfoRequest, GetInfoResponse, error::SdkError};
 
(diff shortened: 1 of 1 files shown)


roeierez (approved):

crates/breez-sdk/core/src/sdk/runtime/server.rs
```diff
@@ -14,11 +15,17 @@ impl RuntimeProfile for ServerRuntime {
...
-        sdk: &BreezSdk,
-        _initial_synced_sender: watch::Sender<bool>,
-    ) {
+    async fn start_sdk_services(&self, sdk: &BreezSdk, initial_synced_sender: watch::Sender<bool>) {
+        // The host drives `sync_wallet` explicitly in server mode, so mark the
+        // SDK as "initially synced" up front. `SparkSyncService::sync_payments`
+        // gates `PaymentSucceeded` emission on this watcher: without it set,
+        // completed receives inserted on sync silently skip the emit, and any
+        // event-driven flow (e.g. a host that subscribes to `PaymentSucceeded`)
+        // misses the notification.
```
JssDWt:
```suggestion
```

JssDWt: Looks good. Just this comment is a bit too verbose here. Not needed.

JssDWt (approved):

PR #905, opened 2026-05-22
skip MySQL/Postgres migration table DDL when table exists

rloomba:
MySQL and PostgreSQL both support externally managed SDK schemas, but the migration runners still emit `CREATE TABLE IF NOT EXISTS` for the migrations tracking table on startup.

That can fail for deployments where schema creation is handled out of band and runtime users do not have schema `CREATE` privileges. This PR updates both migration runners to check `information_schema.tables` before creating the migrations tracking table. If the table already exists, skip the no-op DDL and proceed to read the current migration version.

### DIFF
--- a/crates/spark-mysql/src/migrations.rs
+++ b/crates/spark-mysql/src/migrations.rs
@@ -344,16 +344,20 @@ async fn run_migrations_inner(
         .await
         .map_err(map_db_error)?;
 
-    // Create migrations table if it doesn't exist.
-    let create_table_sql = format!(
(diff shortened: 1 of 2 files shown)


JssDWt:
Hi @rloomba, there is now the `run_migration` boolean on the `MysqlStorageConfig`. When that is set to `false`, the migrations aren't executed. Does that solve this problem?

rloomba:
> Hi @rloomba, there is now the `run_migration` boolean on the `MysqlStorageConfig`. When that is set to `false`, the migrations aren't executed. Does that solve this problem?

doh, totally missed that in the release notes! Yep that will work.

PR #906, opened 2026-05-23
update protocol

JssDWt:
This change updates how the Breez SDK talks to the Spark network for three core
operations: depositing Bitcoin into Spark, receiving a transfer from another user,
and cooperatively withdrawing back to Bitcoin. Each previously took two network
round-trips with the Spark operators; we collapse them into one, matching what the
reference Spark JS SDK already uses.

### DIFF
--- a/crates/spark/src/operator/rpc/spark_rpc_client.rs
+++ b/crates/spark/src/operator/rpc/spark_rpc_client.rs
@@ -126,23 +126,6 @@ impl SparkRpcClient {
     }
 
     #[instrument(level = "info", target = "spark::operator_rpc", skip_all, fields(operator_id = self.operator_id))]
-    pub async fn finalize_transfer_with_transfer_package(
-        &self,
-        req: FinalizeTransferWithTransferPackageRequest,
-    ) -> Result<FinalizeTransferResponse> {
-        debug!(
-            "Calling finalize_transfer_with_transfer_package with request: {:?}",
-            req
-        );
-        Ok(self
-            .spark_service_client()
-            .await?
-            .finalize_transfer_with_transfer_package(req)
-            .await?
-            .into_inner())
-    }
-
-    #[instrument(level = "info", target = "spark::operator_rpc", skip_all, fields(operator_id = self.operator_id))]
     pub async fn claim_transfer(
         &self,
         req: ClaimTransferRequest,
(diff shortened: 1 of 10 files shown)


crates/spark/src/operator/rpc/spark_rpc_client.rs
dangeross: Can also be removed?

crates/spark/src/services/deposit.rs
```diff
@@ -584,201 +584,161 @@ impl DepositService {
...
 
-        // TODO: Verify the returned tx signatures
-
-        let nodes = finalize_resp
-            .nodes
-            .into_iter()
-            .map(|node| {
-                let signing_keyshare = node
-                    .signing_keyshare
-                    .ok_or(ServiceError::MissingSigningKeyshare)?;
```
dangeross, line 739: Looks like ServiceError::MissingSigningKeyshare can be removed

dangeross (commented):
Looks good

crates/spark/src/services/transfer.rs
```diff
@@ -803,134 +805,140 @@ impl TransferService {
...
-                        owner_identity_public_key: identity_public_key,
-                        leaves_to_receive,
-                    })
-                    .await;
+        let chunks = signing_commitments.chunks(leaves.len()).collect::<Vec<_>>();
+        if chunks.len() != 3 {
+            return Err(ServiceError::Generic(
+                "Not enough signing commitments returned".to_string(),
+            ));
+        }
```
dangeross: This is only checking the commitments are chunked into 3, which could lead to a panic accessing the chunks later. Should we check `signing_commitments.len() == 3 * leaves.len()` instead?

crates/spark/src/services/deposit.rs
```diff
@@ -584,201 +584,161 @@ impl DepositService {
...
+                            }),
+                            additional_inputs: Vec::new(),
+                        },
+                    ),
+                    additional_on_chain_utxos: Vec::new(),
+                },
+            )
             .await?;
 
-        // TODO: Verify the returned tx signatures
```
dangeross, line 731: I'm wondering if we should definitely do this now we've dropped verifying the signature shares verifying key. Wdyt (could be a follow up PR)?

JssDWt: Sounds good. I added it for the deposit. I will do others in a follow-up PR, after adding Turnkey integration with the higher-level signer.

crates/spark/src/services/transfer.rs
JssDWt: Good catch on that panic. I added guards around that.

dangeross (approved):
LGTM

danielgranhao (approved):
LGTM

roeierez (approved):
LGTM

PR #909, opened 2026-05-25
passkey: add bulk PRF support and simplify the passkey API surface

erdemyerebasmaz:
Reduces and simplifies the passkey API surface from #781 down to a smaller, more predictable shape.

### API surface

- `PasskeyClient` is the single public entry point (register / signIn / connectWithPasskey / checkAvailability / labels()).
- The default `PasskeyClient` constructor wires up the bundled platform PRF provider; apps that need a custom one inject it via a small builder.
- Unified `connectWithPasskey` onboarding in place of separate create/restore paths _on mobile platforms where silent passkey detection is possible_.
- Every flow (register / signIn / connectWithPasskey) returns the `PasskeyCredential` it used or created, so the app manages passkeys itself (pin a known credential via `allowCredentials`, prevent duplicate registrations via `excludeCredentials`) instead of the SDK tracking credentials for you.
- Per-call, optional derive and sign-in options (allow/exclude credentials, prefer-local) instead of mutating a singleton.
- `userName` is set per registration (defaulting to `rpName`), so an app can vary it to keep credentials distinct under Apple's `(rpId, user.name)` de-duplication.
- Simplified per-platform PRF cores: one configured instance holding the RP identity, replacing the earlier nested ceremony helpers.

<details>
<summary>Breaking changes and migration notes</summary>

- **`CredentialRegistry` removed.** The SDK no longer stores or reconciles known credential IDs. `register`, `signIn`, and `connectWithPasskey` each return a unified `credential: PasskeyCredential?` (replacing `SignInResponse.credentialId` and `ConnectWithPasskeyResponse.registeredCredential`); the app persists whatever it needs. `PasskeyClient.credentials()`, the `credentialRegistry` constructor option, and the `onRegistryError` hook are gone.
- **`RegisteredCredential` renamed to `PasskeyCredential`.** `credentialId` is always present; `userId`, `aaguid`, and `backupEligible` are populated on registration and absent on a sign-in assertion (an assertion carries no attestation).
- **Request fields renamed to WebAuthn spelling.** `allow_credential_ids` and `exclude_credential_ids` become `allowCredentials` and `excludeCredentials`, and `allowCredentials` is now accepted by `connectWithPasskey` as well. Both are optional: omit them for the simple path, set them only for the advanced cases above.
- **`signIn` is the one read path.** It replaces the separate restore and derive calls. The provider exposes only the bulk `deriveSeeds`, which returns the seeds alongside the credential ID the ceremony consumed (the single-value `deriveSeed` and the separate credential-ID observer are both gone).
- **Typed provider errors.** `PasskeyPrfError` becomes `PrfProviderError` with distinct cases for no-matching-credential, credential-already-exists, and authenticator-timed-out, so callers branch on the cause instead of parsing a message; the OS's one generic failure is distinguished by how long the prompt stayed up.
- **`user_id` is provider-owned.** It is minted per registration instead of being passed in by the host, so the SDK owns the credential's stable identifier while the host still controls the displayed `userName`.

</details>

### Bulk PRF derivation

The SDK can now derive multiple seeds in one passkey ceremony. Registering a wallet used to run a separate ceremony per sub-step (publishing the label to Nostr, then connecting to it), prompting the user each time; those now happen in a single ceremony.

### Platform-specific quirks (reviewer notes)

Small workarounds for OS behaviors that are not obvious from the diff:

- **Brief pause after creating a passkey (~800ms, same on iOS and Android).** A newly created passkey can take a moment to become usable while the OS registers it, so the SDK waits up to ~800ms after creation before its first use.
- **Telling "cancelled", "timed out", and "no passkey" apart.** For privacy the OS reports all three as the same generic failure. The SDK infers which one happened from how long the prompt was on screen: an instant failure means there was no matching passkey, a ~55-second failure means the OS timed the prompt out, and anything in between means the user dismissed it. This drives the right message and recovery action.

### Addressed from review

Behavior fixes made on this branch in response to review (all present in the final diff):

- **One credential per `deriveSeeds` call.** When a single bulk derivation splits into more than one assertion (an authenticator that drops `prf.eval.second`, or N > 2 salts), every assertion now pins to the credential the first one resolved to. All seeds from one call therefore come from the same passkey: `setup_wallet`'s Nostr identity (`seeds[0]`) and wallet seed (`seeds[1]`) can no longer bind to different credentials.
- **Post-create grace survives bridge calls (React Native + Flutter iOS).** The per-call native shims now share one module-scoped grace tracker, as Android already did, so the wait armed by `createPasskey` carries into the immediately-following `deriveSeeds` instead of being dropped, avoiding a wasted biometric prompt.
- **Label writes fan out across relays.** `store_label` now publishes to every reachable relay batch rather than only the first, restoring the read/write overlap the early-exit read relies on, so a label can't become undiscoverable after a relay flap.

### Docs

- New **Credential metadata** guide documenting the `PasskeyCredential` fields and the app-side record / pin / exclude pattern, including the AAGUID provider-display use case (mapping AAGUID to an authenticator name and backup status for a passkey-management screen). `allowCredentials` / `excludeCredentials` are framed as optional, advanced knobs: the simple path works without them.

### Misc.

- Add **Comment Style** instructions to `CLAUDE.md` to reduce the amount auto-generated code blocks that are verbose, unnecessary, and/or less informative.

### Stacking

Stacked on `pr/passkey-core` (#781), which targets `main`. Review #781 first.


crates/breez-sdk/wasm/js/passkey-prf-provider/index.js
```diff
@@ -251,112 +565,306 @@ export class PasskeyProvider {
...
+        // the historical behavior; explicit `false` opts out so the
+        // browser falls through to its standard picker (cross-device
+        // QR sign-in, hybrid transports, etc.).
+        const preferImmediate = options.preferImmediatelyAvailableCredentials !== false;
+        // Immediate mediation is privacy-gated to empty allowCredentials
+        // and only enabled when getClientCapabilities() advertises it.
+        if (preferImmediate
+            && allowCredentials.length === 0
+            && await this._supportsImmediateGet()) {
+            this._applyImmediateOption({ publicKey });
```
dangeross: Shouldn't the immediate option (uiMode/mediation) be applied to the root request?

crates/breez-sdk/wasm/src/models/passkey_prf_provider.rs
```diff
@@ -1,17 +1,89 @@
...
+                return PrfProviderError::UserTimedOut;
+            }
+            "PasskeyCredentialNotFoundError" => {
+                let message = js_sys::Reflect::get(&js_error, &JsValue::from_str("message"))
+                    .ok()
+                    .and_then(|v| v.as_string())
+                    .unwrap_or_else(|| "Credential not found".to_string());
+                return PrfProviderError::CredentialNotFound(message);
+            }
+            _ => {}
```
dangeross: It seems like Wasm doesn't support the all the error variant of the other bindings (UserCancelled, AuthenticationFailed, PrfEvaluationFailed, Configuration, PrfNotSupported). Should we at least support UserCancelled? I see a few places throwing similar
```
return new Error('User cancelled authentication')
```

packages/flutter/rust/src/passkey.rs
```diff
@@ -15,165 +17,366 @@ fn panic_message(e: Box<dyn std::any::Any + Send>) -> String {
...
+    } else if lower.contains("prfnotsupported") {
+        PrfProviderError::PrfNotSupported
+    } else if lower.contains("credentialalreadyexists") {
+        PrfProviderError::CredentialAlreadyExists(msg)
+    } else if lower.contains("configuration") {
+        PrfProviderError::Configuration(msg)
+    } else {
+        PrfProviderError::Generic(msg)
+    }
 }
```
dangeross, line 60: Seems to not map userTimedOut

docs/breez-sdk/snippets/go/passkey.go
```diff
@@ -95,19 +170,92 @@ func ListLabels() ([]string, error) {
...
+	// Route the user to the sign-in path: the OS picker will surface
+	// the existing credential and the SDK's identity cache will warm
+	// up on the assertion.
+	prfProvider := &CustomPrfProvider{}
+	passkey := breez_sdk_spark.NewPasskeyClient(prfProvider, nil, nil)
+
+	label := "personal"
+	registerResponse, err := passkey.Register(breez_sdk_spark.RegisterRequest{
+		Label: &label,
+		ExcludeCredentialIds: [][]byte{
```
dangeross: ?
```suggestion
		ExcludeCredentials: [][]byte{
```

crates/breez-sdk/wasm/src/models/passkey_prf_provider.rs
```diff
@@ -20,33 +92,285 @@ unsafe impl Sync for WasmPrfProvider {}
...
+
+    async fn take_last_observed_credential_id(&self) -> Option<Vec<u8>> {
+        if !self.js_has_method(
+            "takeLastObservedCredentialId",
+            &self.supports_take_last_observed,
+        ) {
+            return None;
+        }
+        let target: &JsValue = self.inner.as_ref();
+        let func = js_sys::Reflect::get(target, &JsValue::from_str("takeLastObservedCredentialId"))
```
dangeross: Not defined in the interface below

crates/breez-sdk/wasm/src/models/passkey_prf_provider.rs
```diff
@@ -20,33 +92,285 @@ unsafe impl Sync for WasmPrfProvider {}
...
+            .dyn_into::<Promise>()
+            .map_err(|_| {
+                PrfProviderError::Generic(
+                    "removeKnownCredentialId did not return a Promise".to_string(),
+                )
+            })?;
+        JsFuture::from(promise)
+            .await
+            .map_err(js_error_to_prf_provider_error)?;
+        Ok(())
```
dangeross: I'm not sure if its the intent but it seems each of these functions has excessive boilerplate/reflection going on. And the PrfProvider and extern "C" interface is not fully updated

packages/flutter/rust/src/passkey.rs
```diff
@@ -15,165 +17,366 @@ fn panic_message(e: Box<dyn std::any::Any + Send>) -> String {
...
-/// Callback-based implementation of `PrfProvider` for Flutter.
-///
-/// This struct wraps Dart callbacks to implement the PRF provider trait,
-/// allowing Flutter to provide the passkey PRF implementation.
+/// Wraps Dart callbacks as a [`PrfProvider`] implementation. Each
+/// callback returns a Result so Dart-side throws (e.g.
+/// `PasskeyPrfException`) propagate cleanly. The known-credential
+/// callbacks back `PasskeyClient.credentials()`; hosts without a
+/// registry can have them resolve to empty / no-op on the Dart side.
 struct CallbackPrfProvider {
```
dangeross, line 23: Missing take_last_observed_credential_id?

packages/react-native/src/passkey-prf-provider.ts
```diff
@@ -101,98 +282,364 @@ export interface PasskeyProviderOptions {
...
+  /**
+   * Take ownership of the credential ID captured by the most recent
+   * successful assertion. Returns `null` if no assertion has
+   * completed since the last call.
+   */
+  takeLastObservedCredentialId(): Uint8Array | null {
+    const v = this._lastObservedCredentialId;
+    this._lastObservedCredentialId = null;
+    return v;
   }
```
dangeross: I see in several bindings (RN, Flutter, Wasm) a kind of half implemented version of using the last observed credential id, but it somehow does not seem complete. Maybe I'm missing something? But better to remove than introduce placeholders

crates/breez-sdk/wasm/js/passkey-prf-provider/index.js
erdemyerebasmaz: This is a code residue. This is an experimental feature on Web and it will be removed altogether to reduce confusion to end-users. It will be added once the feature makes it to stable.

erdemyerebasmaz: Dropped in 3473100e7b0a12aef6e4d6dd0b71e44d44a884e2

packages/flutter/rust/src/passkey.rs
erdemyerebasmaz: Added in c343f459ef8466e89de56de5e0817ef7d584838a

docs/breez-sdk/snippets/go/passkey.go
erdemyerebasmaz: Corrected the rename residue in on Go bindings c343f459ef8466e89de56de5e0817ef7d584838a

crates/breez-sdk/wasm/src/models/passkey_prf_provider.rs
erdemyerebasmaz: Declared them in the interface and replaced the reflections 3473100e7b0a12aef6e4d6dd0b71e44d44a884e2

crates/breez-sdk/wasm/src/models/passkey_prf_provider.rs
erdemyerebasmaz: https://github.com/breez/spark-sdk/pull/909#discussion_r3304294294

It was not intended, this was due to missing items on the PrfProvider interface.

crates/breez-sdk/wasm/src/models/passkey_prf_provider.rs
erdemyerebasmaz: Added support for `UserCancelled` in c343f459ef8466e89de56de5e0817ef7d584838a

There are similar cases for `AuthenticationFailed` errors, but I'd like to stop first. WASM error variants need a bit more thought than the native bindings. WebAuthn propagates all errors as `NotAllowedError`, and we currently derive typed errors by matching error messages. This makes the error-mapping less reliable than on other platforms.  I haven't verified it yet, but I suspect the error-mapping breaks on non-English locales since the messages are browser-provided.

crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
```diff
@@ -56,84 +62,330 @@ import java.security.SecureRandom
...
+     * happens at most once per call, so prompts never grow unbounded.
      */
-    public suspend fun deriveSeedOrRegister(
-        activity: Activity,
-        salt: String,
-        rpId: String,
-        rpName: String,
-        userName: String,
-        userDisplayName: String,
+    public suspend fun deriveSeeds(
```
dangeross, line 218: **The post-create grace tracker only exists in the app wrappers, not in the shared core.**

After a passkey is registered, Android's Credential Manager needs a brief moment before the new credential's PRF is reliably usable. If a sign-in (`deriveSeeds`) runs too soon, the single assertion that's meant to cover two salts comes back with only one result, so the code runs a second assertion to get the other (`:377`) — a second biometric prompt for the user.

The `PostCreateGraceTracker` that smooths this over is implemented in the React Native and Flutter modules (`BreezSdkSparkPasskeyModule.kt:27`, `BreezSdkSparkPasskeyPlugin.kt:31`), but not here in the core or in `PasskeyProvider`. iOS does it the opposite way: the grace lives inside `PasskeyAssertionCore`, so anything built on that core inherits it.

So on Android, someone using the SDK through the UniFFI binding directly (Kotlin Multiplatform, or a native Android app) doesn't get the grace — `PasskeyProvider` registers and then derives with nothing in between, and their first sign-in right after sign-up can show the extra prompt.

A few things I wasn't sure about:
- Is it intentional that the UniFFI/KMM path skips the grace that the RN and Flutter wrappers apply?
- iOS keeps this in the core so every consumer inherits it; is there a reason Android keeps it per-wrapper instead?

packages/flutter/android/build.gradle
```diff
@@ -64,6 +64,12 @@ dependencies {
     implementation "androidx.credentials:credentials:1.3.0"
     implementation "androidx.credentials:credentials-play-services-auth:1.3.0"
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:1.8.0"
+    // KnownCredentialsStore: Block Store (Google-account synced) +
+    // EncryptedSharedPreferences (local fallback). tasks.await()
+    // bridges Play Services Tasks to coroutines.
+    implementation "androidx.security:security-crypto:1.1.0-alpha06"
```
dangeross: **These three dependencies don't appear to be used by anything in the Flutter Android tree.**

The comment says they back a `KnownCredentialsStore` (Block Store + EncryptedSharedPreferences), but I can't find that class anywhere in the repo — `KnownCredentialsStore` only shows up in doc comments (`passkey_prf_provider.rs:147`, and the WASM models doc), never as an implementation. The two Kotlin files that actually exist here (`BreezSdkSparkPasskeyPlugin.kt` and the synced `CredentialManagerPrfCore.kt`) don't import `androidx.security.crypto`, the Block Store API, or `kotlinx-coroutines-play-services`.

The React Native Android binding has the same passkey functionality and doesn't declare any of these in its `build.gradle`, so they don't look like a shared requirement either.

A couple of questions:
- Is there a `KnownCredentialsStore` implementation meant to land that these are staged for, or are they left over from an earlier approach?
- If Flutter genuinely needs them but RN doesn't, what's the difference? Otherwise this is three extra dependency trees shipped to every Flutter consumer.
- `androidx.security:security-crypto` is pinned at `1.1.0-alpha06` — is shipping an alpha here intended?

crates/breez-sdk/wasm/js/passkey-prf-provider/index.js
```diff
@@ -394,71 +845,111 @@ export class PasskeyProvider {
...
-            return new Error('Unknown WebAuthn error');
+    _mapAssertionError(error, elapsedMs, augmentNoCredHelp = false) {
+        if (!error) return new Error('Unknown WebAuthn error');
+        if (error.name === 'NotAllowedError') {
+            if (elapsedMs < NO_CRED_FAST_FAIL_MS) {
+                const msg = 'Credential not found'
+                    + (augmentNoCredHelp ? CREDENTIAL_REGISTRY_HELP_SUFFIX : '');
+                return new PasskeyCredentialNotFoundError(msg);
+            }
+            if (elapsedMs >= BIOMETRIC_TIMEOUT_MS) {
```
dangeross: **A host-configured `defaultTimeoutMs` shorter than 55s makes real timeouts look like cancels.**

`defaultTimeoutMs` is applied as the WebAuthn ceremony timeout (`publicKey.timeout`, line 624). But the cancel-vs-timeout split here is hardwired to `BIOMETRIC_TIMEOUT_MS` (55s): a `NotAllowedError` is only called a timeout once elapsed crosses 55s, otherwise it becomes a `PasskeyUserCancelledError`.

So if a host sets `defaultTimeoutMs: 30000`, the ceremony aborts around 30s with `NotAllowedError`, elapsed is ~30s, and it's reported as a user cancel even though nothing was dismissed. A host branching on the typed error would treat that timeout as a deliberate cancel (e.g. skip a retry it would otherwise do). The register path has the same pattern (`_mapError`, line 931).

The 55s value is the OS biometric-sheet inactivity window, which is a sensible ceiling, but it ignores a configured timeout that fires earlier. Should the threshold track whichever is smaller — the configured `defaultTimeoutMs` or the 55s OS window?

crates/breez-sdk/core/src/passkey/mod.rs
```diff
@@ -88,126 +94,178 @@ fn validate_label(label: &str) -> Result<(), PasskeyError> {
...
-    pub async fn get_wallet(&self, label: Option<String>) -> Result<Wallet, PasskeyError> {
-        let label = label.unwrap_or_else(|| DEFAULT_LABEL.to_string());
-        validate_label(&label)?;
-        let root_key = self.prf_provider.derive_prf_seed(label.clone()).await?;
-        let mnemonic = prf_to_mnemonic(&root_key)?;
-        Ok(Wallet {
+        // Prime the Nostr client cache with the keys derived in this
+        // ceremony so subsequent label ops don't trigger another PRF.
+        let keys = derive_nostr_keypair(&seeds[0])?;
+        let _ = self
```
dangeross: **`setup_wallet` primes the cached Nostr identity with `OnceCell::set`, which silently no-ops once the cell is populated.**

`nostr_client` caches the Nostr label-store identity, derived from the fixed `ACCOUNT_MASTER_SALT` — so it's per-credential. `setup_wallet` primes it with `let _ = self.nostr_client.set(...)` here. `OnceCell::set` returns an `Err` (discarded) when the cell is already initialized, so the prime only takes effect the first time.

On a `PasskeyClient` that's already run one successful operation (cell holds identity A's keys), a later `register` of a new credential — or a sign-in that resolves to a different credential — derives identity B's keys, but the `set` no-ops and leaves A in the cell. `register` then publishes the label through the cached A client (line 222-224), so B's label lands on A's Nostr feed. The returned wallet is still B's (the seed is derived fresh), but a later discovery sign-in for B won't find that label.

This doesn't hit the common one-identity flow, and `connect_with_passkey`'s sign-in→register fallback is safe (the failed silent sign-in returns before priming the cell). It's specifically about reusing one client across credentials.

Is a single `PasskeyClient` meant to be reused across identities, or is one instance per identity the intended contract? If reuse is in scope, the prime would need to rebind when the derived identity changes rather than no-op.

crates/breez-sdk/bindings/langs/shared/ios-passkey/Sources/PasskeyAssertionCore.swift
```diff
@@ -0,0 +1,1107 @@
...
+        controller.delegate = delegate
+        controller.presentationContextProvider = delegate
+        delegate.anchor = anchor.presentationAnchor()
+        delegate.onAssertionCredentialId = makeCaptureCallback()
+
+        let preferImmediate = preferImmediatelyAvailableCredentials
+        return try await withCheckedThrowingContinuation { continuation in
+            delegate.assertionContinuation = continuation
+            delegate.extractPrf = true
+            delegate.ceremonyStartedAt = Date()
```
dangeross: **The timer starts before the prompt is shown.** 

`ceremonyStartedAt = Date()` is set here, before `DispatchQueue.main.async { performAssertionRequest(...) }`. If the main thread is busy, the wait for that dispatch gets counted as ceremony time. A real no-credential fast-fail (meant to land under 300ms) can then measure over 300ms and come back as `.userCancelled` instead of `.credentialNotFound` — which also stops `connect_with_passkey` from falling through to register. Capturing the time inside the closure, right before `performRequests`, would scope it to the actual prompt.

crates/breez-sdk/bindings/langs/shared/ios-passkey/Sources/PasskeyAssertionCore.swift
```diff
@@ -0,0 +1,1107 @@
...
+        // Hand the freshly-minted user handle to the delegate so it can
+        // attach it to the returned `IosRegisteredCredential`. The
+        // platform never echoes `user.id` back through the credential
+        // object, so we surface the value we chose ourselves.
+        delegate.registrationUserId = resolvedUserId
+
+        let credential: IosRegisteredCredential = try await withCheckedThrowingContinuation { continuation in
+            delegate.registrationContinuation = continuation
+            delegate.extractPrf = false
+            delegate.ceremonyStartedAt = Date()
```
dangeross: **The timer starts before the prompt is shown.**

`ceremonyStartedAt = Date()` is captured here before `DispatchQueue.main.async { controller.performRequests() }`, and it's wall-clock `Date`. The elapsed feeds the same `.canceled` classification in `didCompleteWithError` (line 1026) — for registration that's the 55s cancel-vs-timeout boundary (the 300ms no-credential path isn't relevant here).

crates/breez-sdk/bindings/langs/shared/ios-passkey/Sources/PasskeyAssertionCore.swift
```diff
@@ -0,0 +1,1107 @@
...
+        let challenge = Self.randomBytes(count: 32)
+        let resolvedUserId = Self.randomBytes(count: 16)
+        // `ASAuthorizationPlatformPublicKeyCredentialProvider` only
+        // exposes a 3-arg overload (challenge:name:userID:) on the
+        // current iOS / macOS SDKs; the 4-arg displayName overload is
+        // security-key-only. Password managers (Apple Passwords, GPM)
+        // surface `user.name` as the primary label here, so we pass
+        // userDisplayName when callers want that label set, otherwise
+        // we fall back to userName.
+        _ = userDisplayName // accepted for parity with caller signatures; not consumed by the platform overload
```
dangeross: **`userDisplayName` is accepted and documented as the WebAuthn `user.displayName`, but iOS silently drops it.**

The field is plumbed through the public surface (the Swift `PasskeyProvider` doc says "Maps to the WebAuthn `user.displayName`") and stored on the core (line 408), but `register` discards it here and passes only `name: userName`. Android sets both `name` and `displayName` (`CredentialManagerPrfCore.kt:705-706`) and web sets `displayName` (`index.js:770`), so iOS is the only platform that drops it.

Separately, the inline comment just above doesn't match the code: it says "we pass userDisplayName when callers want that label set, otherwise we fall back to userName," but the code never passes `userDisplayName`. That comment should be corrected regardless of what happens to the field.

The platform provider only exposes the 3-arg `createCredentialRegistrationRequest` (no displayName), so it can't be set the same way here. Given that, should the field/doc be adjusted on iOS to say displayName has no effect, rather than accepting a value that's silently dropped?

packages/react-native/ios/BreezSdkSparkPasskey.swift
```diff
@@ -63,34 +69,36 @@ class BreezSdkSparkPasskey: NSObject {
...
         userDisplayName: String,
-        excludeCredentialIds: [String],
+        excludeCredentials: [String],
         resolve: @escaping RCTPromiseResolveBlock,
         reject: @escaping RCTPromiseRejectBlock
     ) {
-        let excludeIds: [Data] = excludeCredentialIds.compactMap { Data(base64Encoded: $0) }
+        let excludeIds: [Data] = excludeCredentials.compactMap { Data(base64Encoded: $0) }
 
+        let core = PasskeyAssertionCore(
```
dangeross, line 98: **Each call builds a fresh `PasskeyAssertionCore`, so the post-create grace doesn't survive between calls.**

After `register`, the OS needs a brief moment before the new credential's PRF is reliably usable; the core's grace tracker absorbs that by arming a deadline in `register` and waiting it out before the next assertion. But each method here builds its own core, so the instance that arms the grace in `createPasskey` (line 88) is discarded when the call returns, and the `deriveSeeds` call (line 50) builds a new core whose tracker was never armed. The wait becomes a no-op, and the immediate post-register assertion can race the PRF-readiness window — dropping the second salt and costing an extra biometric prompt.

The Swift Package's `PasskeyProvider` holds one persistent core (`PasskeyProvider.swift:59`), so it's unaffected; this is specific to the per-call shims. Would holding a single core for the module's lifetime (keyed by `rpId`) be feasible here, so the grace can span the createPasskey→deriveSeeds sequence?

packages/flutter/ios/Classes/BreezSdkSparkPasskeyPlugin.swift
```diff
@@ -20,265 +22,176 @@ public class BreezSdkSparkPasskeyPlugin: NSObject, FlutterPlugin {
...
-        } catch PasskeyError.credentialNotFound {
-            guard autoRegister else { throw PasskeyError.credentialNotFound }
+        var excludeCredentials: [Data] = []
+        if let rawIds = args["excludeCredentials"] as? [FlutterStandardTypedData] {
+            excludeCredentials = rawIds.map { $0.data }
+        } else if let base64Ids = args["excludeCredentials"] as? [String] {
+            excludeCredentials = base64Ids.compactMap { Data(base64Encoded: $0) }
+        }
 
+        let core = PasskeyAssertionCore(
```
dangeross, line 68: **Each call builds a fresh `PasskeyAssertionCore`, so the post-create grace doesn't survive between calls.**

After `register`, the OS needs a brief moment before the new credential's PRF is reliably usable; the core's grace tracker absorbs that by arming a deadline in `register` and waiting it out before the next assertion. But each handler here builds its own core, so the instance that arms the grace in `handleCreatePasskey` (line 62) is discarded when the call returns, and the `handleDeriveSeeds` call (line 112) builds a new core whose tracker was never armed. The wait becomes a no-op, and the immediate post-register assertion can race the PRF-readiness window — dropping the second salt and costing an extra biometric prompt.

The Swift Package's `PasskeyProvider` holds one persistent core (`PasskeyProvider.swift:59`), so it's unaffected; this is specific to the per-call shims. Would holding a single core for the plugin's lifetime (keyed by `rpId`) be feasible here, so the grace can span the createPasskey→deriveSeeds sequence?

crates/breez-sdk/core/src/passkey/nostr_client.rs
```diff
@@ -102,32 +95,81 @@ impl NostrSaltClient {
...
+    pub async fn store_label(&self, label: &str) -> Result<(), PasskeyError> {
+        let relays = self.read_relay_candidates();
+        let timeout = Duration::from_secs(RELAY_TIMEOUT_SECS);
+        let filter = Filter::new()
+            .author(self.keys.public_key())
+            .kind(nostr::Kind::TextNote);
+
+        let mut last_err: Option<String> = None;
+
+        for chunk in relays.chunks(2) {
```
dangeross, line 129: **`store_label` writes the label to only the first reachable relay batch, which can make it undiscoverable later.**

This loop writes to the first batch of 2 relays that's reachable, then returns (line 163). The base published labels to the full relay set (`publish_label` → `create_write_client` → all `ensure_server_relays()`); narrowing to one batch is a deliberate latency win (the doc notes it avoids the cold NIP-65 fetch). Separately, the read path `fetch_events_with_fallback` (line 416) early-exits on the first batch that returns `Ok`, even when that batch is empty.

Together they can drop a label from discovery: if batch 0 errors at write time, the label lands on batch 1; if batch 0 is reachable at a later read, the read returns `Ok([])` from batch 0 and never tries batch 1, so `list_labels` comes back empty. Under the old full-fanout write the label was on every relay, so the early-exit read always found it — so this is a durability regression versus the base, gated on a relay flap between the write and a later read.

The early-exit read code is unchanged from the base; it's the narrowed write that breaks the overlap the read relied on. Is the single-batch write meant to be paired with something that converges the relays (the background NIP-65 sync?), or should label writes fan out wider to keep read/write overlap guaranteed?

crates/breez-sdk/core/src/passkey/error.rs
```diff
@@ -44,13 +82,31 @@ pub enum PasskeyPrfError {
...
+    /// Coarse classification for the caller. Lets hosts branch on a
+    /// small, actionable enum instead of pattern-matching every
+    /// variant.
+    #[must_use]
+    pub fn kind(&self) -> ErrorKind {
+        match self {
+            Self::UserCancelled => ErrorKind::Cancel,
+            Self::UserTimedOut => ErrorKind::Timeout,
+            Self::CredentialNotFound(_) => ErrorKind::NoCredential,
+            Self::PrfNotSupported | Self::PrfEvaluationFailed(_) => ErrorKind::PrfUnsupported,
```
dangeross: **`PrfEvaluationFailed` is classified as `PrfUnsupported`, which points the host at the wrong remedy.**

`ErrorKind::PrfUnsupported` tells the host "the authenticator doesn't implement PRF; fall back or switch credential providers." But `PrfEvaluationFailed` means the opposite: PRF *is* supported and was attempted, the evaluation just produced no usable output. Android raises it only when the `prf` extension is present but `results` is missing/empty (`CredentialManagerPrfCore.kt:647, 651`); an actually-absent extension is `PrfNotSupported` (line 645). iOS maps it from `invalidResponse` (a malformed response), and the CLI raises it on HMAC computation failures — none of those mean "no PRF support."

So a transient or anomalous eval failure on a PRF-capable authenticator gets surfaced as "switch credential providers," which won't help the user. `Internal` (generic retry) looks closer to the real situation. Should `PrfEvaluationFailed` map to `Internal` (or its own kind), leaving `PrfUnsupported` for `PrfNotSupported` only?

crates/breez-sdk/core/src/passkey/passkey_client.rs
```diff
@@ -0,0 +1,638 @@
...
+    /// [`RegisterRequest::exclude_credentials`]. Ignored on the
+    /// silent sign-in attempt.
+    #[cfg_attr(feature = "uniffi", uniffi(default = []))]
+    pub exclude_credentials: Vec<Vec<u8>>,
+}
+
+/// Response from [`PasskeyClient::connect_with_passkey`].
+///
+/// `registered_credential` doubles as the path discriminator:
+/// - `Some(credential)`: a new credential was registered (silent
```
dangeross: `Some(credential)` / `None` are Rust-specific. These `uniffi::Record` docs render to Swift/Kotlin/TS/Dart, where the reader sees `RegisteredCredential?` — per the CLAUDE.md cross-language doc rule, use "present"/"absent" framing. Same at lines 74-75 (`Some(label)` / `None`).

crates/breez-sdk/core/src/passkey/passkey_client.rs
```diff
@@ -0,0 +1,638 @@
...
+    /// depends on the OS fast-failing with [`PrfProviderError::CredentialNotFound`]
+    /// when no local credential exists. Without the fast-fail, the
+    /// OS would show the cross-device picker and a user dismiss would
+    /// surface as a real `Cancel`, which is propagated unchanged
+    /// rather than triggering registration. All other errors
+    /// (`Cancel`, `Timeout`, `Configuration`, etc.) propagate
+    /// unchanged: only `CredentialNotFound` flips to the register
+    /// path.
+    ///
+    /// Mobile-only: meant for iOS 18+ / Android 9+ where
```
dangeross, line 273: "Android 9+" is wrong for Credential Manager passkeys — that's Android 14 / API 34 (the RN provider doc already says API 34).

crates/breez-sdk/core/src/passkey/passkey_client.rs
```diff
@@ -0,0 +1,638 @@
...
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct SignInResponse {
+    pub wallet: Wallet,
+    /// Empty on the fast path. Populated on discovery (or empty if
+    /// the label store was unreachable).
+    pub labels: Vec<String>,
+    /// The credential ID the user used for this sign-in, when the
+    /// underlying [`PrfProvider`] surfaces it. `None` for providers
+    /// that don't expose this signal (CLI / file-backed / hardware
```
dangeross: This attributes `None` only to "CLI / file-backed / hardware providers," but after `ef2a49b9` dropped the Flutter/RN placeholders, those platform providers also return `None`. Worth updating so RN/Flutter integrators don't expect a value here.

crates/breez-sdk/core/src/passkey/mod.rs
```diff
@@ -88,126 +94,178 @@ fn validate_label(label: &str) -> Result<(), PasskeyError> {
...
+    /// in one PRF ceremony (dual-salt where the platform supports it).
+    /// Primes the identity cache so subsequent [`Self::list_labels`] /
+    /// [`Self::store_label`] need no extra PRF prompts.
+    /// `publish_label = false` skips the Nostr write: used by
+    /// speculative cold-restore.
+    pub async fn setup_wallet(
+        &self,
+        request: SetupWalletRequest,
+    ) -> Result<WalletSetup, PasskeyError> {
+        let label = request.label.unwrap_or_else(|| self.default_label.clone());
```
dangeross, line 193: `default_label` is applied here and then validated, so `PasskeyConfig { default_label: Some("") }` fails `validate_label` on every `label = None` call. Validating at config construction would catch it once instead.

dangeross (commented):
I'm trying a new Claude skill, sorry if its a bit wordy

packages/flutter/rust/src/passkey.rs
erdemyerebasmaz: Addressed in #915

packages/react-native/src/passkey-prf-provider.ts
erdemyerebasmaz: Addressed in #915 Removed.

crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
erdemyerebasmaz: Lifted into the canonical `CredentialManagerPrfCore` in 1c563bf6 so the UniFFI/KMM `PasskeyProvider` path (and any other direct-core consumer) inherits the same cross-call grace iOS gets. The RN + Flutter Android wrappers now pass their wrapper-scoped tracker into each per-call core instead of arming/consuming themselves, so they keep working across fresh-core-per-call without duplicating the bookkeeping.

packages/flutter/android/build.gradle
erdemyerebasmaz: Dropped in 6c53d8ea: the three deps were holdovers from the retired Block Store / EncryptedSharedPreferences `KnownCredentialsStore` implementation that the `CredentialRegistry` refactor replaced.

crates/breez-sdk/wasm/js/passkey-prf-provider/index.js
erdemyerebasmaz: Fixed in 6ae46768 by extracting the cancel-vs-timeout threshold into a `_cancelVsTimeoutThresholdMs()` helper that both `_mapAssertionError` and `_mapError` call. A host-configured `defaultTimeoutMs` shorter than `BIOMETRIC_TIMEOUT_MS` now wins in both classification sites, so a real timeout near the host limit no longer misreports as a cancel.

crates/breez-sdk/core/src/passkey/mod.rs
erdemyerebasmaz: Fixed in 58bd016b by swapping the `OnceCell` for `RwLock<Option<NostrSaltClient>>`. `setup_wallet` now takes the write lock and overwrites, so a follow-up register on the same `PasskeyClient` rebinds onto the new identity's relays instead of silently no-oping onto the prior cell.

crates/breez-sdk/bindings/langs/shared/ios-passkey/Sources/PasskeyAssertionCore.swift
erdemyerebasmaz: Fixed in cf35ff7b: moved the `delegate.ceremonyStartedAt = Date()` assignment inside the `DispatchQueue.main.async` closure so the wait for the main-thread dispatch doesn't count as ceremony time. Applied to the canonical `ios-passkey` copy plus the three synced copies (swift binding, packages/flutter, packages/react-native).

crates/breez-sdk/bindings/langs/shared/ios-passkey/Sources/PasskeyAssertionCore.swift
erdemyerebasmaz: Fixed in the same cf35ff7b commit (covers both the assertion and register start-time captures, plus the three synced iOS-core copies).

crates/breez-sdk/bindings/langs/shared/ios-passkey/Sources/PasskeyAssertionCore.swift
erdemyerebasmaz: Documented the silent drop in 9026041e. The platform create-credential overload available on current iOS SDKs only takes `challenge:name:userID:`, so `userDisplayName` never reaches AuthenticationServices at registration. Kept on the constructor for cross-platform parity, with a doc note so callers aren't surprised when password managers show `user.name` as the primary label.

crates/breez-sdk/core/src/passkey/error.rs
erdemyerebasmaz: Already addressed earlier on this branch in 4939910c: `PrfEvaluationFailed` now maps to its own `ErrorKind::PrfFailed`, and `AuthenticationFailed` got its own `ErrorKind::AuthFailure` instead of both being collapsed under `PrfUnsupported`. Surfacing both here in case it helps with the review trail.

crates/breez-sdk/core/src/passkey/passkey_client.rs
erdemyerebasmaz: Fixed in e981adb2: `SignInRequest.label` and `ConnectWithPasskeyResponse.registered_credential` both use the present/absent framing now so the docs render correctly in Swift / Kotlin / TS / Dart.

crates/breez-sdk/core/src/passkey/passkey_client.rs
erdemyerebasmaz: Kept as-is: `preferImmediatelyAvailableCredentials` IS honored on Android 9+ (API 28+, the minimum Credential Manager passkey PRF target). The API 34 number in the RN provider doc is about something else (the doc there can be revisited separately). Happy to revisit if you've seen behavior that suggests otherwise on pre-API-34 devices.

crates/breez-sdk/core/src/passkey/passkey_client.rs
erdemyerebasmaz: Resolved indirectly: the credential-ID fold into `deriveSeeds` (#915 work) put Flutter and RN back in the "populates credentialId" bucket, since their providers now read it from the native bridge's `DeriveSeedsOutput` instead of relying on the dropped `take_last_observed` placeholders (see flutter `passkey_prf_provider.dart:252` / RN `passkey-prf-provider.ts:377`). The current doc listing "CLI / file-backed / hardware providers" as the `None` case is accurate for the platform-bundled providers.

crates/breez-sdk/core/src/passkey/mod.rs
erdemyerebasmaz: Fixed in 5ebed2bc: `Passkey::new` now runs `validate_label` on `config.default_label` once at construction and silently falls back to the constant default when invalid, so per-call `label = None` derivations no longer fail validation every time.

packages/react-native/ios/BreezSdkSparkPasskey.swift
```diff
@@ -63,34 +72,36 @@ class BreezSdkSparkPasskey: NSObject {
...
```
dangeross, line 98: The post-create grace doesn't survive between bridge calls here, so the immediate post-register `deriveSeeds` can still race the credential's PRF-readiness window and cost an extra biometric prompt.

`createPasskey` and `deriveSeeds` are separate bridge methods, and each builds its own `PasskeyAssertionCore` (`:50`, `:91`) with the default `graceTracker`. The tracker armed by `register` in `createPasskey` is discarded when that core goes out of scope, and the `deriveSeeds` core gets a fresh `PostCreateGraceTracker` whose `deadline` is nil, so `consume()` is a no-op.

The Android wrappers handle this by holding one module-scoped `PostCreateGraceTracker` and injecting it into every per-call core (`BreezSdkSparkPasskeyModule.kt:52`, passed at `:111`/`:225`), so the grace spans the createPasskey to deriveSeeds sequence even though the cores are per-call. The iOS core already accepts the same injectable `graceTracker` param (`PasskeyAssertionCore.swift:388`), so the wiring would mirror Android: a persistent `private let graceTracker = PostCreateGraceTracker()` on the module, passed as `graceTracker: graceTracker` into each core. The SPM `PasskeyProvider` isn't affected since it keeps one core for its lifetime.

Was the shared-tracker wiring just missed on the iOS wrappers when it landed on Android?

packages/flutter/ios/Classes/BreezSdkSparkPasskeyPlugin.swift
```diff
@@ -20,265 +22,179 @@ public class BreezSdkSparkPasskeyPlugin: NSObject, FlutterPlugin {
...
```
dangeross, line 68: Same post-create grace gap as the React Native iOS wrapper: the grace armed in `handleCreatePasskey` doesn't survive to `handleDeriveSeeds`, so the immediate post-register derive can race the PRF-readiness window and cost an extra prompt.

Each handler builds its own `PasskeyAssertionCore` with the default `graceTracker` (`:62`, `:112`), so the tracker armed by `register` is discarded and the `deriveSeeds` core's tracker has a nil `deadline`, making `consume()` a no-op.

The Android Flutter plugin avoids this by holding one `PostCreateGraceTracker` field and injecting it into every per-call core (`BreezSdkSparkPasskeyPlugin.kt:54`, passed at `:128`/`:224`). The iOS core accepts the same injectable param (`PasskeyAssertionCore.swift:388`), so the fix mirrors Android: a persistent `private let graceTracker = PostCreateGraceTracker()` passed as `graceTracker: graceTracker` into each core.

crates/breez-sdk/core/src/passkey/mod.rs
```diff
@@ -88,126 +96,197 @@ fn validate_label(label: &str) -> Result<(), PasskeyError> {
...
         validate_label(&label)?;
-        let root_key = self.prf_provider.derive_prf_seed(label.clone()).await?;
-        let mnemonic = prf_to_mnemonic(&root_key)?;
-        Ok(Wallet {
+
+        // [account_master, label]: dual-salt single ceremony where the
+        // platform supports it, fallback to two prompts otherwise.
+        let salts = vec![ACCOUNT_MASTER_SALT.to_string(), label.clone()];
```
dangeross, line 198: `setup_wallet` derives `[ACCOUNT_MASTER, label]` in one `derive_seeds` call and treats `seeds[0]` (the Nostr label-store identity) and `seeds[1]` (the wallet seed) as belonging to the same credential. The built-in providers don't guarantee that once the call splits into more than one assertion.

WebAuthn PRF output is credential-specific, and a single `derive_seeds` can run two separate assertions: the dropped-`prf.eval.second` recovery path (JS `index.js:344`, iOS `PasskeyAssertionCore.swift:483`, Android `CredentialManagerPrfCore.kt:411`), or any N > 2 salts. None pins the later assertion to the credential the first one resolved to: `allowCredentials` is computed once and reused unchanged, and the returned `credential_id` is the last assertion's.

So if an authenticator drops `prf.eval.second` (the Android core comment cites older Credential Manager) and the user has more than one passkey for this RP with no `allow_credentials` pin, prompt 1 can resolve to credential X and prompt 2 to credential Y. Then `seeds[0]` is X's account-master and `seeds[1]` is Y's wallet seed: the wallet returned is Y's, but its label is published to X's Nostr feed, and `credential_id` reports Y. A later discovery sign-in for Y won't find the label, and one for X points at a wallet it can't re-derive. Same cross-identity hazard as the `OnceCell` rebind, but inside a single `setup_wallet` call.

Low probability for the SDK's own 2-salt flow (needs the dropped-second path, multiple credentials, and a divergent pick, mostly older Android), but a host calling the public `PrfProvider` trait with N > 2 salts hits the multi-pair path routinely, and "N outputs from a passkey" reads as a single credential.

Each provider already captures the first assertion's observed credential id. Should the recovery and multi-pair assertions pin to it (add it to `allowCredentials`) so every salt in one `derive_seeds` call resolves to the same credential?

dangeross (commented):
Looking good

packages/react-native/ios/BreezSdkSparkPasskey.swift
erdemyerebasmaz: Fixed in e77fe8c00 (commit ae573e651). The RN module now holds a module-scoped `PostCreateGraceTracker` (the module is a singleton per bridge) and injects it into the `createPasskey` and `deriveSeeds` cores, so the grace spans the createPasskey to deriveSeeds sequence. Mirrors the Android module.

packages/flutter/ios/Classes/BreezSdkSparkPasskeyPlugin.swift
erdemyerebasmaz: Fixed in e77fe8c00 (commit ae573e651). The Flutter plugin now holds a module-scoped `PostCreateGraceTracker` (the plugin instance is retained by the registrar) and injects it into each per-call `PasskeyAssertionCore`, so the grace armed by `createPasskey` survives to `deriveSeeds`. Mirrors the Android plugin.

crates/breez-sdk/core/src/passkey/nostr_client.rs
erdemyerebasmaz: Confirmed still valid. The background NIP-65 sync does not rescue it: it republishes the pre-write events, not the newly written label. Fixed in commit e77fe8c00: `store_label` now fans out the write to every reachable relay batch (signs once, broadcasts the same event, skips batches that already carry the label), restoring the read/write overlap the base relied on.

packages/react-native/ios/BreezSdkSparkPasskey.swift
erdemyerebasmaz: Fixed in commit ae573e651. The RN module now holds one module-scoped `PostCreateGraceTracker` (the module is a singleton per bridge) and injects it into the `createPasskey` and `deriveSeeds` cores, so the grace spans the createPasskey to deriveSeeds sequence. Mirrors the Android module. And yes, the shared-tracker wiring was simply missed on the iOS shims when it landed on Android.

packages/flutter/ios/Classes/BreezSdkSparkPasskeyPlugin.swift
erdemyerebasmaz: Fixed in commit ae573e651. The Flutter plugin now holds one module-scoped `PostCreateGraceTracker` (the plugin instance is retained by the registrar) and injects it into each per-call core, so the grace armed by `createPasskey` survives to `deriveSeeds`. Mirrors the Android plugin.

crates/breez-sdk/core/src/passkey/mod.rs
erdemyerebasmaz: Fixed in commit a164ef1f2. Each core now pins allow/allowCredentials to the credential the first assertion resolves to, so every salt in one `derive_seeds` call derives from the same passkey, including the dropped prf.eval.second recovery path and N>2 pairs. The JS core also bypasses its per-call registry merge when pinned so the pin cannot be re-widened. This makes the previously-assumed "same credential across every chunk" invariant actually hold, so setup_wallet binds seeds[0] (the Nostr identity) and seeds[1] (the wallet seed) to one credential.

crates/breez-sdk/wasm/src/models/passkey_prf_provider.rs
```diff
@@ -1,17 +1,88 @@
...
+                    .unwrap_or_else(|| "Credential not found".to_string());
+                return PrfProviderError::CredentialNotFound(message);
+            }
+            _ => {}
+        }
+    }
 
-pub(crate) fn js_error_to_passkey_prf_error(js_error: JsValue) -> PasskeyPrfError {
     let error_message = js_error
         .as_string()
```
dangeross: This PR adds typed `message` extraction for the four named error classes above, but the catch-all fallback still uses `js_error.as_string()`. `as_string()` returns `Some` only for a JS string primitive, so a thrown `Error` object (the common case for anything outside the four named classes) yields `None` and collapses to the literal `"Passkey PRF error occurred"`, discarding the real `error.message`.

The practical effect: an unrecognized provider failure (a network error, a `SecurityError`, an unexpected throw) reaches the host as an opaque generic with no detail to log or branch on.

Would it be worth reading `message` in the fallback the same way the named branches just above do? `js_sys::Error`'s typed `name()` / `message()` getters could also fold the typed dispatch and this fallback into one pass.

erdemyerebasmaz: Addressed in 782fa3464. The mapping now reads the error through `js_sys::Error`, so `name` and `message` come from the typed getters and the real `error.message()` is surfaced on every path (the four named classes and the catch-all alike); only a genuinely empty message falls back to a generic label. The typed dispatch and the fallback are folded into one pass.

docs/breez-sdk/snippets/rust/src/passkey.rs
```diff
@@ -1,91 +1,306 @@
...
-    let passkey = Passkey::new(prf_provider, None);
+    let passkey = PasskeyClient::new(prf_provider, None, None);
 
-    // Derive the wallet from the passkey (pass None for the default wallet)
-    let wallet = passkey.get_wallet(Some("personal".to_string())).await?;
+    let response = passkey
+        .connect_with_passkey(ConnectWithPasskeyRequest {
+            label: Some("personal".to_string()),
+            allow_credentials: vec![],
+            exclude_credentials: vec![],
```
roeierez: NIT: use ..Default::default() for both credentials?

docs/breez-sdk/snippets/rust/src/passkey.rs
```diff
@@ -1,91 +1,306 @@
...
-        breez_api_key: Some("<breez api key>".to_string()),
-        ..NostrRelayConfig::default()
-    };
-    let passkey = Passkey::new(prf_provider, Some(relay_config));
+    let passkey = PasskeyClient::new(prf_provider, None, None);
+
+    let response = passkey
+        .register(RegisterRequest {
+            label: Some("personal".to_string()),
+            exclude_credentials: vec![],
```
roeierez: Also here?

roeierez (approved):
Only NITS. otherwise approved.

docs/breez-sdk/snippets/rust/src/passkey.rs
erdemyerebasmaz: Done in 09d073856: `connect_with_passkey` and `register` now use `..Default::default()` like `sign_in`.

I swept the rest of the snippets for the same thing so nothing is missed:
- Same fix applied to the credential-metadata anchor's `register`, and to the Python and C# `connect` anchors (they relied on the uniffi default elsewhere but spelled out an empty list here).
- Swift / Kotlin / Go already omit these lists.
- WASM, React Native and Flutter keep the explicit empty lists on purpose: their generators (wasm-bindgen, ubrn, FRB) do not honor `uniffi(default = [])`, so the fields are required there (e.g. Flutter generates `RegisterRequest({this.label, required this.excludeCredentials})`). Worth a separate look if we want those bindings to default them too.
- The demonstration anchors (recover-already-exists; the sign-in pin in credential-metadata) keep their placeholder lists deliberately.

docs/breez-sdk/snippets/rust/src/passkey.rs
erdemyerebasmaz: Done in 09d073856. See the reply on the connect-with-passkey comment above for the full sweep across the other languages.

docs/breez-sdk/src/guide/passkey.md
```diff
@@ -93,111 +82,110 @@ Connects the domain to iOS and macOS applications for passkey sharing:
...
+- **`userName`** (optional): Maps to the WebAuthn `user.name`. Treated as the user's unique identifier for the credential and shown in the OS account picker during sign-in. Pass a stable per-user value if each registration should surface as a distinct entry (Apple's Passwords app, in particular, dedupes credentials by `(rpId, user.name)`). Defaults to `rpName`. Registration-only.
+- **`userDisplayName`** (optional): Maps to the WebAuthn `user.displayName`. The user-friendly label the OS / browser MAY (but is not required to) show in the picker; behavior varies by platform. Defaults to `userName`. Registration-only.
+- **`breezApiKey`**: Your Breez API key. Used to authenticate to the Breez relay (<a target="_blank" href="https://github.com/nostr-protocol/nips/blob/master/42.md">NIP-42</a>) for label storage. Without an API key, label sync falls back to public relays only.
+- **`passkeyConfig`** (optional): Carries {{#name default_label}}, the label used when {{#name PasskeyClient.register}} / {{#name PasskeyClient.sign_in}} receive no label. Falls back to the SDK's internal `"Default"` when unset.
```
dangeross: Lets use {#name parameter}

docs/breez-sdk/src/guide/passkey.md
```diff
@@ -93,111 +82,110 @@ Connects the domain to iOS and macOS applications for passkey sharing:
...
+{{#tabs passkey:connect-with-passkey}}
 
-{{#tabs passkey:check-availability}}
+### Two-button flow (web)
 
-<h2 id="connecting-with-passkey">
-    <a class="header" href="#connecting-with-passkey">Connecting with a passkey</a>
-    <a class="tag" target="_blank" href="https://breez.github.io/spark-sdk/breez_sdk_spark/passkey/struct.Passkey.html#method.get_wallet">API docs</a>
-</h2>
+`connect_with_passkey` is not surfaced on the WASM target. The unified flow needs a silent "no credential here" signal so it can fall through to register, and WebAuthn deliberately collapses that case into the same `NotAllowedError` as a user cancel for privacy reasons. There is no reliable way on web for the SDK to tell the two apart.
```
dangeross: We're talking specifically about Wasm, so lets use `connectWithPasskey`

docs/breez-sdk/src/guide/passkey.md
```diff
@@ -93,111 +82,110 @@ Connects the domain to iOS and macOS applications for passkey sharing:
...
 
-To connect with a passkey, instantiate the built-in `PasskeyProvider`, pass it to `Passkey`, call {{#name Passkey.get_wallet}} to derive a wallet, then pass its seed to {{#name connect}}. The label defaults to `"Default"` when omitted.
+The recommended UX on web is two buttons: a **Sign In** button calling `signIn` and a separate **Create Account** button calling `register`. Let the user pick the right one instead of trying to auto-detect. See [Direct sign-in / register](#direct-sign-in--register) for the call shapes.
 
-{{#tabs passkey:connect-with-passkey}}
+### Direct sign-in / register
+
+For finer control, call {{#name PasskeyClient.sign_in}} and {{#name PasskeyClient.register}} directly. Use this path when separating Sign-In and Create-Account flows, or when adding a new label on a returning user without going through `connect_with_passkey`.
```
dangeross:
```suggestion
For finer control, call {{#name PasskeyClient.sign_in}} and {{#name PasskeyClient.register}} directly. Use this path when separating Sign-In and Create-Account flows, or when adding a new label on a returning user without going through {{#name PasskeyClient.connect_with_passkey}}.
```

docs/breez-sdk/snippets/rust/src/passkey.rs
erdemyerebasmaz: Done. 09d0738568d75bfcd66ab6623274a5e02d6614d8
Made them optional instead of defaulting to an empty list: 9dedab56cafcc07b1977a8dfe6c1ad8a621d30bb

docs/breez-sdk/src/guide/passkey.md
```diff
@@ -93,111 +82,110 @@ Connects the domain to iOS and macOS applications for passkey sharing:
...
 
-Apps that share an RP ID, whether the default `keys.breez.technology` or a custom domain, recognize the same user's passkey without per-app enrollment. Users don't need to manually enroll a passkey before using a built-in provider; it auto-registers a new credential on first use if one doesn't already exist for the RP ID.
+### Unified flow (mobile)
 
-Pass a `PasskeyProvider` instance to `Passkey` as shown in the [Connecting with a passkey](#connecting-with-passkey) snippets below. That is the recommended path for almost every app.
+One "Use Passkey" button backed by {{#name PasskeyClient.connect_with_passkey}}: silent sign-in for a returning user, automatic fall-through to registration on a fresh device. The response's `credential` field carries whichever credential signed in or was registered; the register path also fills the attestation fields (`aaguid`, `backupEligible`). See [Credential metadata](./passkey_credential_metadata.md) for using these.
```
dangeross: Shared (mobile) prose, so use the language macros for these fields:

```suggestion
One "Use Passkey" button backed by {{#name PasskeyClient.connect_with_passkey}}: silent sign-in for a returning user, automatic fall-through to registration on a fresh device. The response's {{#name credential}} field carries whichever credential signed in or was registered; the register path also fills the attestation fields ({{#name aaguid}}, {{#name backup_eligible}}). See [Credential metadata](./passkey_credential_metadata.md) for using these.
```

docs/breez-sdk/src/guide/passkey.md
```diff
@@ -93,111 +82,110 @@ Connects the domain to iOS and macOS applications for passkey sharing:
...
 
-## Checking passkey availability
+Internally the silent attempt pins `preferImmediatelyAvailableCredentials = true` so the OS fast-fails (no UI, sub-300ms on iOS / Android) when no local credential exists; only {{#enum PrfProviderError::CredentialNotFound}} flips to register, all other errors (`Cancel`, `Timeout`, `Configuration`) propagate unchanged.
```
dangeross: `preferImmediatelyAvailableCredentials` is the request field; the macro renders it per language.

```suggestion
Internally the silent attempt pins {{#name prefer_immediately_available_credentials}} to true so the OS fast-fails (no UI, sub-300ms on iOS / Android) when no local credential exists; only {{#enum PrfProviderError::CredentialNotFound}} flips to register, all other errors (`Cancel`, `Timeout`, `Configuration`) propagate unchanged.
```

docs/breez-sdk/src/guide/passkey.md
```diff
@@ -93,111 +82,110 @@ Connects the domain to iOS and macOS applications for passkey sharing:
...
+Pass `wallet.seed` to {{#name connect}} in either case.
 
-<h2 id="storing-a-label">
-    <a class="header" href="#storing-a-label">Storing a label</a>
-    <a class="tag" target="_blank" href="https://breez.github.io/spark-sdk/breez_sdk_spark/passkey/struct.Passkey.html#method.store_label">API docs</a>
-</h2>
+### Error recovery
 
-Publish a label to Nostr so it can be discovered later.
+The SDK collapses every passkey failure into seven actionable [`ErrorKind`](https://breez.github.io/spark-sdk/breez_sdk_spark/passkey/enum.ErrorKind.html) values: branching on `error.kind()` is the canonical recovery pattern:
```
dangeross: `error.kind()` and `ErrorKind` are Rust-only: `kind()` is a bare `pub fn`, not `#[uniffi::export]`, so it isn't callable on Swift/Kotlin/Flutter/Go/Python/C#, and the Wasm path uses the typed `instanceof` classes instead. The per-language snippets confirm it (Swift/Kotlin/Python/Go all match `PrfProviderError` variants directly). Suggest keying this table on `{{#enum PrfProviderError::...}}` variants and presenting `error.kind()` / `ErrorKind` as a Rust-only convenience. While here, `excludeCredentials` (L166) becomes `{{#name exclude_credentials}}` and `NotAssociated` (L169) becomes `{{#enum PasskeyAvailability::NotAssociated}}`; the web exception classes on L172 are correctly left literal.

docs/breez-sdk/src/guide/passkey.md
```diff
@@ -93,111 +82,110 @@ Connects the domain to iOS and macOS applications for passkey sharing:
...
+Replace `TEAMID` with your Apple Developer Team ID and `com.example.yourapp` with your bundle identifier. Your app must also declare the <a target="_blank" href="https://developer.apple.com/documentation/bundleresources/entitlements/com.apple.developer.associated-domains">Associated Domains</a> capability in Xcode (**Signing & Capabilities** → **Associated Domains** → `webcredentials:<your-rp-domain>`).
 
 <div class="warning">
-<h4>Expo Managed Workflow</h4>
-If you're using Expo, the Breez SDK plugin can configure this automatically. See the <a href="install_react_native.html#plugin-options">React Native/Expo installation guide</a> for details on the <code>enablePasskey</code> option.
+<h4>iOS / macOS: Associated Domains entitlement required</h4>
+Without the Associated Domains entitlement declared in Xcode, passkey operations on iOS / macOS will fail with a configuration error even though {{#name check_availability}} returns {{#enum PasskeyAvailability::Available}} (the OS-level check can't verify entitlements at runtime).
```
dangeross: Nit: L122 uses the qualified `{{#name PasskeyClient.check_availability}}`; match it here.

```suggestion
Without the Associated Domains entitlement declared in Xcode, passkey operations on iOS / macOS will fail with a configuration error even though {{#name PasskeyClient.check_availability}} returns {{#enum PasskeyAvailability::Available}} (the OS-level check can't verify entitlements at runtime).
```

docs/breez-sdk/src/guide/passkey_credential_metadata.md
```diff
@@ -0,0 +1,41 @@
+# Credential metadata
+
+Every passkey flow returns the credential it used or created, so your app can manage passkeys itself: pin a specific credential on sign-in, prevent duplicate registrations, and build a settings screen that lists registered passkeys by provider and backup status.
+
+{{#name register}}, {{#name sign_in}}, and {{#name connect_with_passkey}} each return a `credential` field. It is unset only for PRF backends that do not surface one (CLI / file-backed / hardware providers); the built-in platform providers always populate it.
```
dangeross:
```suggestion
{{#name register}}, {{#name sign_in}}, and {{#name connect_with_passkey}} each return a {{#name credential}} field. It is unset only for PRF backends that do not surface one (CLI / file-backed / hardware providers); the built-in platform providers always populate it.
```

docs/breez-sdk/src/guide/passkey_credential_metadata.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Fields
+
+`PasskeyCredential` carries:
```
dangeross:
```suggestion
{{#name PasskeyCredential}} carries:
```

docs/breez-sdk/src/guide/passkey_credential_metadata.md
```diff
@@ -0,0 +1,41 @@
...
+
+| Field | Available | Use it for |
+|---|---|---|
+| {{#name credential_id}} | always | Passing back as `allow_credentials` / `exclude_credentials` (see below). |
```
dangeross: These render snake_case here but camelCase (`excludeCredentials`) in passkey.md:166: same field, two spellings. The macro fixes both.

```suggestion
| {{#name credential_id}} | always | Passing back as {{#name allow_credentials}} / {{#name exclude_credentials}} (see below). |
```

docs/breez-sdk/src/guide/passkey_prf_providers.md
```diff
@@ -0,0 +1,30 @@
+# PRF providers
+
+The built-in {{#name PasskeyProvider}} covers the common case. Reach for this page when:
+
+- You need fine-grained `PasskeyProvider` options (custom `userName`, `userDisplayName`, etc.).
```
dangeross:
```suggestion
- You need fine-grained {{#name PasskeyProvider}} options (custom {{#name user_name}}, {{#name user_display_name}}, etc.).
```

docs/breez-sdk/src/guide/passkey_prf_providers.md
```diff
@@ -0,0 +1,30 @@
...
+- You're integrating Python, Go, or C# (no built-in `PasskeyProvider` ships for those bindings).
+- You need a custom `PrfProvider` (CLI YubiKey, FIDO2, air-gapped backup file, hardware module).
+
+## Built-in PasskeyProvider options
+
+In addition to `rpId` and `rpName`, the `PasskeyProvider` constructor accepts:
```
dangeross:
```suggestion
In addition to {{#name rp_id}} and {{#name rp_name}}, the {{#name PasskeyProvider}} constructor accepts:
```

docs/breez-sdk/src/guide/passkey_prf_providers.md
```diff
@@ -0,0 +1,30 @@
...
+| {{#name user_name}} | {{#name rp_name}} | Maps to the WebAuthn `user.name`. Treated as the user's unique identifier for the credential and shown in the OS account picker during sign-in. Pass a stable per-user value if each registration should surface as a distinct entry (Apple's Passwords app, in particular, dedupes credentials by `(rpId, user.name)`; iOS additionally treats this as the only label exposed by `ASAuthorizationPlatformPublicKeyCredentialProvider`). Registration-only. |
+| {{#name user_display_name}} | {{#name user_name}} | Maps to the WebAuthn `user.displayName`. The user-friendly label the OS / browser MAY (but is not required to) show in the picker; Android Credential Manager and the WebAuthn L3 picker surface it, iOS ignores. Registration-only. |
+
+<div class="warning">
+<h4>C# / Go / Python limitation</h4>
+
+The SDK does not ship a built-in `PasskeyProvider` for C#, Go, or Python (no native passkey API to wrap on those targets). Integrators on those bindings implement their own `PrfProvider` and pass it to `PasskeyClient::new(...)` directly.
```
dangeross: `PasskeyClient::new(...)` is Rust syntax that reads as gibberish on other tabs. (Same for the bare `PasskeyProvider` / `PrfProvider` backticks at L6/L9/L26/L28 if you want those macro'd too.)

```suggestion
The SDK does not ship a built-in {{#name PasskeyProvider}} for C#, Go, or Python (no native passkey API to wrap on those targets). Integrators on those bindings implement their own {{#name PrfProvider}} and pass it to {{#name PasskeyClient}} directly.
```

docs/breez-sdk/src/guide/uxguide_passkey.md
```diff
@@ -5,49 +5,83 @@
...
+| Path | OS prompts on mobile |
+|---|---|
+| Returning user (silent sign-in succeeds) | **1** prompt (single dual-salt assertion derives master + label) |
+| New user (silent sign-in fast-fails → register) | **2** prompts (1 create + 1 dual-salt assertion) |
 
-### Credential ID management
+The SDK's bulk-PRF path runs the create + label-store + assertion as the two prompts above; a naive sequence would cost three.
 
-`createPasskey()` returns the credential ID of the newly registered passkey. Store this ID locally (for example in `localStorage`, `SharedPreferences`, or `UserDefaults`) so you can:
+**Web:** Two CTAs ("Sign In" and "Create Account"), letting the user pick the right one. WebAuthn deliberately collapses "no credential found" and "user cancelled" into the same `NotAllowedError` for privacy reasons, so the SDK cannot reliably auto-fall-through on web the way it can on mobile. The unified `connect_with_passkey` is therefore not surfaced on the WASM target: call {{#name PasskeyClient.sign_in}} and {{#name PasskeyClient.register}} as separate entry points and trust the user to know whether they're returning or new.
```
dangeross: Web-scoped paragraph, so the JS spelling is right here.

```suggestion
**Web:** Two CTAs ("Sign In" and "Create Account"), letting the user pick the right one. WebAuthn deliberately collapses "no credential found" and "user cancelled" into the same `NotAllowedError` for privacy reasons, so the SDK cannot reliably auto-fall-through on web the way it can on mobile. The unified `PasskeyClient.connectWithPasskey` is therefore not surfaced on the WASM target: call `PasskeyClient.signIn` and `PasskeyClient.register` as separate entry points and trust the user to know whether they're returning or new.
```

docs/breez-sdk/src/guide/uxguide_passkey.md
```diff
@@ -5,49 +5,83 @@
...
-1. **Prevent duplicate credentials.** On subsequent calls to `createPasskey()`, pass the stored IDs via the `excludeCredentialIds` parameter. When the authenticator already holds one of these credentials, the platform will reject the registration with a "credential already registered" error instead of silently creating a duplicate.
-2. **Track which credentials belong to this app instance.** The credential ID is an opaque byte array assigned by the authenticator. It is not sensitive, but it is only useful on the device where the credential was created.
+### Adding a new label to an existing identity
+
+For a user who already has a passkey and wants to create another wallet under a new label:
+
+1. {{#name PasskeyClient.sign_in}} with the new label. This derives master + new-label seeds in one assertion AND seeds the SDK's identity cache.
+2. {{#name PasskeyLabels.store}} (via {{#name PasskeyClient.labels}}) to publish the new label to Nostr. This uses the cached identity for free: no additional prompt.
+
+Total: **1** OS prompt. The reverse order (`labels().store` before `sign_in`) costs 2 prompts because each call would derive the master salt independently.
```
dangeross: Should also look at making `labels().store` more language agnostic.
```suggestion
Total: **1** OS prompt. The reverse order (`labels().store` before {{#name PasskeyClient.sign_in}}) costs 2 prompts because each call would derive the master salt independently.
```

docs/breez-sdk/src/guide/uxguide_passkey.md
```diff
@@ -5,49 +5,83 @@
...
+
+### Credential ID and user-handle management
+
+`createPasskey()` returns a `PasskeyCredential` carrying:
```
dangeross: This whole section is written in web/JS terms in an otherwise cross-platform guide: `createPasskey()` (L34, L43) → `{{#name create_passkey}}`, the field names `credentialId` / `userId` / `aaguid` / `backupEligible` (L36-39, L44) → `{{#name ...}}`, and `excludeCredentials` (L43) → `{{#name exclude_credentials}}`.

docs/breez-sdk/src/guide/uxguide_passkey.md
```diff
@@ -5,49 +5,83 @@
...
 const exclude = existingId
   ? [Uint8Array.from(atob(existingId), c => c.charCodeAt(0))]
   : [];
-const newId = await prfProvider.createPasskey(exclude);
+const newCredential = await prfProvider.createPasskey({ excludeCredentials: exclude });
 ```
 
+### Recovery paths
+
+The canonical branch point is `error.kind()`: it collapses the nine `PrfProviderError` variants into seven actionable [`ErrorKind`](https://breez.github.io/spark-sdk/breez_sdk_spark/passkey/enum.ErrorKind.html) values. Match on `ErrorKind` instead of the individual variants:
```
dangeross: Same Rust-only issue as the passkey.md recovery table: `error.kind()` / `ErrorKind` (here and at L77, L86) aren't available off Rust, so reframe around `{{#enum PrfProviderError::...}}` variants. `UserCancelled` (L69) → `{{#enum PrfProviderError::UserCancelled}}`.

docs/breez-sdk/src/guide/passkey_credential_metadata.md
```diff
@@ -0,0 +1,41 @@
...
+{{#tabs passkey:credential-metadata}}
+
+For cross-device continuity, back this store with synced storage: iCloud Keychain (`kSecAttrSynchronizable`) on iOS, Block Store on Android, or your own synced backend. Plain local storage does not survive reinstall or replicate to a second device.
+
+## Reusing stored credential IDs
+
+The IDs you store are passed back to the SDK through two optional request lists. Both are unset by default, so the simple sign-in and registration paths work without them:
+
+- `allow_credentials` on {{#name sign_in}} lists the registered credentials a sign-in may use. The OS offers only those, which streamlines its picker. Passing the credential a returning user last signed in with also keeps a seed-deriving wallet on that credential, so they re-open the same wallet.
+- `exclude_credentials` on {{#name register}} lists the credentials already registered for this account, so the OS refuses to create a second one on the same device and raises {{#enum PrfProviderError::CredentialAlreadyExists}} instead. Route that to sign-in:
```
dangeross:
```suggestion
- {{#name allow_credentials}} on {{#name sign_in}} lists the registered credentials a sign-in may use. The OS offers only those, which streamlines its picker. Passing the credential a returning user last signed in with also keeps a seed-deriving wallet on that credential, so they re-open the same wallet.
- {{#name exclude_credentials}} on {{#name register}} lists the credentials already registered for this account, so the OS refuses to create a second one on the same device and raises {{#enum PrfProviderError::CredentialAlreadyExists}} instead. Route that to sign-in:
```

dangeross (commented):
Looks good. Just docs nits

erdemyerebasmaz:
> Looks good. Just docs nits

Thanks for the review, feedback is addressed in fd01cea80502de1146867bfcc450a09d88067e24

dangeross (approved):
LGTM

PR #911, opened 2026-05-26
Emit SdkEvent::Optimization in both client and server mode

danielgranhao:
## Summary

`SdkEvent::Optimization` was defined and exposed through every binding, but core never emitted it: `handle_wallet_event` in the client runtime logged `WalletEvent::Optimization` and dropped it, and server mode didn't subscribe to wallet events at all. External listeners never saw any optimization progress.

This PR wires the bridge in a runtime-agnostic place and reshapes the underlying plumbing so server mode keeps its "background tasks disabled" contract:

- **Dedicated channel in spark-wallet.** Optimization moves off the omnibus `WalletEvent` broadcast onto its own `broadcast::Sender<OptimizationEvent>` exposed via `SparkWallet::subscribe_optimization_events`. The channel is silent unless an optimization run is in flight, so subscribers pay no clone cost for unrelated wallet events. `WalletEvent::Optimization` is removed (only consumer was the SDK itself).
- **Runtime-agnostic forwarder.** A small `optimization_forwarder` spawn in core subscribes to the new channel and re-emits `SdkEvent::Optimization`. The `WalletEvent::Optimization` arm in the client runtime's `handle_wallet_event` is gone.
- **Lazy spawn in server mode.** Client mode spawns the forwarder eagerly (auto-optimization can fire from `BackgroundProcessor` independently). Server mode defers the spawn to the first `start_leaf_optimization` call via an `OnceCell`, so SDK instances that never opt into optimization carry no forwarder task — matching the `background_tasks_enabled = false` promise.

Plus a regression itest (`tests/optimization.rs`) that triggers `start_leaf_optimization`, asserts `Started → Completed` (rejects `Skipped` and lone `Completed` without `Started`), and runs in both client and server mode via paired fixtures.

### DIFF
--- a/crates/breez-sdk/breez-itest/src/fixtures/mod.rs
+++ b/crates/breez-sdk/breez-itest/src/fixtures/mod.rs
@@ -3,7 +3,10 @@ pub mod docker;
 pub mod lnurl;
 
 use anyhow::Result;
-use breez_sdk_spark::{MaxFee, Network, StableBalanceConfig, StableBalanceToken, default_config};
+use breez_sdk_spark::{
+    LeafOptimizationConfig, MaxFee, Network, StableBalanceConfig, StableBalanceToken,
+    default_config, default_server_config,
+};
 use rand::RngCore;
 use rstest::fixture;
 use tracing::info;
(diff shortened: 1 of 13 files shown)


roeierez:
Thanks Daniel!
I see the Tokio task in server mode. It’s not the end of the world, but it does raise a flag for me and makes me wonder why we even need optimization events in server mode.
It seems to me that optimization events are only useful when the SDK performs optimizations automatically, rather than when they are triggered by a user call.
Happy to get your thoughts on this.

danielgranhao:
> Thanks Daniel! I see the Tokio task in server mode. It’s not the end of the world, but it does raise a flag for me and makes me wonder why we even need optimization events in server mode. It seems to me that optimization events are only useful when the SDK performs optimizations automatically, rather than when they are triggered by a user call. Happy to get your thoughts on this.

Yes, I had that in mind so I tried to minimize impact by only starting the task when optimization is actually taking place (even if cost is very low). The reasons I opted for this:
- gives integrator good visibility into optimization progress. It's true polling is possible, but seems like worse UX?
- consistency across client/server models. I don't like that we would have events that never fire on server mode
- cost seems very low

danielgranhao:
After offline discussions we've agreed to go for a blocking API for manual optimizations. I'll open a new PR soon.

PR #912, opened 2026-05-26
Insert child payments during conversions to avoid sync

dangeross:
Conversion-leg `Payment` records previously only landed in storage on the
next `sync_wallet`, so `conversion_details.from`/`to` on the parent payment
showed `null` until a sync filled them in. Now both legs are inserted as
soon as the conversion completes.

`FlashnetClient::execute_swap` returns directly the `WalletTransfer`/`TokenTransaction` it
already produced for the outbound asset (no extra RPC). The sent leg is
built from that and inserted as-is; the received leg is polled by id and claimed
before insertion. On timeout, the next sync remains the fallback.


roeierez (approved):
Nice! LGTM

danielgranhao (approved):
LGTM

PR #913, opened 2026-05-27
Auto-recover refundable conversions via Flashnet API

dangeross:
When a BTC ↔ token conversion fails between `transfer_asset` (Phase 1) and the local-state write that drives the refunder, the BTC stays locked at the pool and the SDK's existing refunder never finds it (no local row to act on).

Root cause: recovery was gated on a single local-DB write surviving. Any disruption (Android backgrounding, process killed, storage glitch) silently stranded funds.

## Approach

Stop relying on local DB as the recovery source of truth. Treat Flashnet's `list_clawback_transfers` endpoint as authoritative; local state is a display cache only.

- **`convert()` failure paths** emit `RuntimeEvent::ConversionRecoveryNeeded { clawback_id, pool_id }` instead of writing a local "needs refund" flag.
- **Client runtime** spawns a 30s retry loop calling `recover_conversion` until success or shutdown. Startup also runs a single `FullScan` to catch orphans from prior process lifetimes.
- **Server runtime** no-ops the event. Operators drive recovery on demand via the public `BreezSdk::refund_pending_conversions` method (now returns counts).
- **5-min age threshold** on FullScan protects against racing in-flight swaps on a multi-device wallet (same seed, two devices).

### DIFF
--- a/crates/breez-sdk/bindings/examples/cli/langs/csharp/Commands.cs
+++ b/crates/breez-sdk/bindings/examples/cli/langs/csharp/Commands.cs
@@ -423,15 +423,13 @@ public static class Commands
                     .Select(s => Enum.Parse<SparkHtlcStatus>(s.Trim(), ignoreCase: true))
                     .ToArray();
                 paymentDetailsFilterList.Add(new PaymentDetailsFilter.Spark(
-                    htlcStatus: statuses,
-                    conversionRefundNeeded: null
+                    htlcStatus: statuses
                 ));
             }
 
             if (txHash != null)
             {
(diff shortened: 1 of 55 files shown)


JssDWt:
As a general rule I would prefer if the client-side database _is_ the source of truth for determining where money is. There are other places where we treat the server-side state as the source of truth, I want to address those later too, like leaf refresh, etc. 

You think it's possible to do this by persisting the intent before doing the action? I think that means the ssp endpoints have to be idempotent, because you may call them twice (when you persisted, called endpoint, didn't persist result, then restart).

dangeross:
Superseded by #927

roeierez:
was opened by mistake

PR #914, opened 2026-05-27
Collapse leaf optimization API to blocking optimize_leaves

danielgranhao:
## Summary

Replaces the polling-based optimization API (`start_leaf_optimization` /
`cancel_leaf_optimization` / `get_leaf_optimization_progress`) with a single
blocking `optimize_leaves(request)` call.

Also fixes a pre-existing wiring bug: on `main`, the SDK runtime received
`WalletEvent::Optimization` from spark-wallet but only logged it — it never
forwarded the event to `SdkEvent::Optimization` subscribers. This PR wires
that emission, so the (now-renamed) `SdkEvent::AutoOptimization` actually
reaches integrators.

## API

```rust
sdk.optimize_leaves(OptimizeLeavesRequest { mode: Full })             // Full
sdk.optimize_leaves(OptimizeLeavesRequest { mode: SingleRound })  // one round
```

`OptimizeLeavesRequest { mode }` is mandatory (consistent with the rest of
the SDK API) and `mode` defaults to `Full`, so `OptimizeLeavesRequest::default()`
runs a full optimization. The call returns `OptimizeLeavesResponse { outcome }`,
where `outcome` is `OptimizationOutcome { Completed { rounds_executed }, InProgress }`.
The result is wrapped in a response struct so fields can be added later without
a breaking change.

- **Full**: blocks until done; `outcome` is `Completed { rounds_executed }`.
  `rounds_executed == 0` indicates the wallet was already optimal at call
  time (no swap performed).
- **SingleRound**: executes at most one round per call. `outcome` is
  `InProgress` while work remains, `Completed` on the round that converges
  (or `Completed { rounds_executed: 0 }` if the very first call sees an
  already-optimal wallet). Lets callers drive a UI / cancel between rounds.

Concurrent calls reject with `SdkError::OptimizationAlreadyRunning`. Runs
preempted by the SDK to free leaves for a payment return
`SdkError::OptimizationCancelled`.

Terminal detection uses a planner-emitted `fully_converges` hint, so a
single-swap, fully-converging plan returns `Completed` in the same call
(no extra no-op iteration needed to discover the end).

## Events

`SdkEvent::Optimization` → `SdkEvent::AutoOptimization`. Events now fire
**only** for the background auto-optimizer; manual `optimize_leaves` calls
are silent — inspect the return value instead. Internally,
`OptimizationEvent` was renamed `AutoOptimizationEvent` at the spark and
spark-wallet layers to reflect this.

`AutoOptimizationEvent::Skipped` is retained — it's a distinct concept on
the background-event path, separate from the `OptimizationOutcome` return
value.

## Breaking changes

- `BreezSdk::start_leaf_optimization` / `cancel_leaf_optimization` /
  `get_leaf_optimization_progress` removed.
- `SdkEvent::Optimization` → `SdkEvent::AutoOptimization`.

### DIFF
--- a/crates/breez-sdk/core/src/models/mod.rs
+++ b/crates/breez-sdk/core/src/models/mod.rs
@@ -1851,14 +1851,21 @@ pub enum OptimizationMode {
     SingleRound,
 }
 
-/// Optional configuration for [`BreezSdk::optimize_leaves`]. Callers that
-/// want the default `Full` behavior can pass `None` (or omit the argument
-/// where the binding supports it).
-#[derive(Debug, Clone, Deserialize, Serialize)]
+/// Request for [`BreezSdk::optimize_leaves`]. Defaults to
+/// [`OptimizationMode::Full`].
+#[derive(Debug, Clone, Default, Deserialize, Serialize)]
 #[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
-pub struct OptimizeLeavesOptions {
+pub struct OptimizeLeavesRequest {
     /// Controls how much work the call performs before returning.
     pub mode: OptimizationMode,
+}
+
+/// Response from a [`BreezSdk::optimize_leaves`] call.
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct OptimizeLeavesResponse {
+    /// The outcome of the optimization run.
+    pub outcome: OptimizationOutcome,
 }
 
 /// Outcome of a [`BreezSdk::optimize_leaves`] call.
@@ -1877,7 +1884,8 @@ pub struct OptimizeLeavesOptions {
 ///
 /// ```ignore
 /// loop {
-///     match sdk.optimize_leaves(SingleRound).await? {
+///     let request = OptimizeLeavesRequest { mode: OptimizationMode::SingleRound };
+///     match sdk.optimize_leaves(request).await?.outcome {
 ///         OptimizationOutcome::InProgress => continue,
 ///         OptimizationOutcome::Completed { .. } => break,
 ///     }
--- a/crates/breez-sdk/core/src/models/mod.rs
+++ b/crates/breez-sdk/core/src/models/mod.rs
@@ -1867,17 +1867,19 @@ pub struct OptimizeLeavesOptions {
 /// The SDK holds no cross-call state — callers driving a `SingleRound`
 /// loop maintain their own cumulative counter if they need one.
 ///
+/// A `Completed { rounds_executed: 0 }` outcome means the wallet was
+/// already optimal at call time (no swap was needed).
+///
 /// **`SingleRound` loop pattern**: terminate on anything that isn't
-/// `InProgress` — `Completed` (the planner confirmed this swap finished
-/// optimization) or `Skipped` (the wallet was already optimal at call
-/// time).
+/// `InProgress`. `Completed` covers both the final swap of a productive
+/// run and the "already optimal" no-op case (the latter as
+/// `rounds_executed: 0`).
 ///
 /// ```ignore
 /// loop {
 ///     match sdk.optimize_leaves(SingleRound).await? {
 ///         OptimizationOutcome::InProgress => continue,
-///         OptimizationOutcome::Completed { .. }
-///         | OptimizationOutcome::Skipped => break,
+///         OptimizationOutcome::Completed { .. } => break,
 ///     }
 /// }
 /// ```
@@ -1888,14 +1890,13 @@ pub enum OptimizationOutcome {
     /// Returned by `Full` runs on success, and by `SingleRound` runs
     /// whose swap was the final one needed (the planner produced a
     /// single-swap plan with a convergence guarantee).
+    /// `rounds_executed == 0` means the wallet was already optimal —
+    /// no work was performed.
     Completed { rounds_executed: u32 },
     /// `SingleRound` only: a round ran but the planner could not
     /// guarantee it was the last. The caller should invoke
     /// `optimize_leaves` again.
     InProgress,
-    /// No optimization work was performed because the wallet was already
-    /// optimal at call time. Returned by either mode on a no-op call.
-    Skipped,
 }
 
 /// A contact entry containing a name and payment identifier.
--- a/crates/breez-sdk/core/src/models/mod.rs
+++ b/crates/breez-sdk/core/src/models/mod.rs
@@ -1840,11 +1840,62 @@ pub struct LnurlReceiveMetadata {
     pub sender_comment: Option<String>,
 }
 
+/// Mode of a manually-triggered optimization run.
+#[derive(Debug, Clone, Default, Deserialize, Serialize, PartialEq)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
+pub enum OptimizationMode {
+    /// Run until no further optimization is productive.
+    #[default]
+    Full,
+    /// Execute a single round and return so the caller can drive progress.
+    SingleRound,
+}
+
+/// Optional configuration for [`BreezSdk::optimize_leaves`]. Callers that
+/// want the default `Full` behavior can pass `None` (or omit the argument
+/// where the binding supports it).
+#[derive(Debug, Clone, Deserialize, Serialize)]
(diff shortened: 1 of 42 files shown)


crates/breez-sdk/core/src/sdk/api.rs
```diff
@@ -242,31 +242,34 @@ impl BreezSdk {
...
-    /// Returns the current optimization progress snapshot.
-    pub fn get_leaf_optimization_progress(&self) -> OptimizationProgress {
-        self.spark_wallet.get_leaf_optimization_progress().into()
+    /// Manual runs do not emit events; events ([`SdkEvent::AutoOptimization`])
+    /// are reserved for the background auto-optimizer.
+    #[cfg_attr(feature = "uniffi", uniffi::method(default(options = None)))]
+    pub async fn optimize_leaves(
+        &self,
+        options: Option<OptimizeLeavesOptions>,
+    ) -> Result<OptimizationOutcome, SdkError> {
```
dangeross: What are your thoughts about keeping the API Request/Response struct naming for consistency? I guess the options/outcome would be wrapped in a structs in this case

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -1840,11 +1840,62 @@ pub struct LnurlReceiveMetadata {
...
+    /// whose swap was the final one needed (the planner produced a
+    /// single-swap plan with a convergence guarantee).
+    Completed { rounds_executed: u32 },
+    /// `SingleRound` only: a round ran but the planner could not
+    /// guarantee it was the last. The caller should invoke
+    /// `optimize_leaves` again.
+    InProgress,
+    /// No optimization work was performed because the wallet was already
+    /// optimal at call time. Returned by either mode on a no-op call.
+    Skipped,
```
roeierez: Perhaps we can avoid this variant and just use Completed {round_executed: 0} ?

crates/spark/src/tree/leaf_optimizer.rs
```diff
@@ -51,20 +53,24 @@ impl LeafOptimizationOptions {
...
-    /// The total number of rounds to execute.
-    pub total_rounds: u32,
+/// Internal flag tracking whether an optimization run is in flight.
+///
+/// Held in an `Arc<Mutex<_>>` so the public `is_running()` getter, the
+/// atomic check-and-set in `start`/`run`, and the cleanup in
+/// [`RunningGuard`]'s `Drop` all coordinate through the same lock.
+#[derive(Clone, Debug, Default)]
+struct RunState {
+    is_running: bool,
```
roeierez, line 63: What is the risk of not holding this flag and let user run optimization in parallel on his own risk? Just looking to see what is the tradeoff for simplifying here.

roeierez (changes requested):
Looks good.

crates/spark/src/tree/leaf_optimizer.rs
danielgranhao: No critical risk. The risk is wasted work and potentially worse optimization results. But I would still lean toward keeping it because I don't see any reason to have concurrent optimizations. 

Also, we have the cancellation mechanism that is used when sending a payment if it's having trouble reserving leaves. I think any simplification gains from removing the lock (~30 LOC) are offset by the added complexity required to make the cancel mechanism work with multiple concurrent optimization tasks.

crates/breez-sdk/core/src/models/mod.rs

crates/breez-sdk/core/src/sdk/api.rs
```diff
@@ -242,31 +242,34 @@ impl BreezSdk {
...
-
-    /// Returns the current optimization progress snapshot.
-    pub fn get_leaf_optimization_progress(&self) -> OptimizationProgress {
-        self.spark_wallet.get_leaf_optimization_progress().into()
+    /// Manual runs do not emit events; events ([`SdkEvent::AutoOptimization`])
+    /// are reserved for the background auto-optimizer.
+    #[cfg_attr(feature = "uniffi", uniffi::method(default(options = None)))]
+    pub async fn optimize_leaves(
+        &self,
+        options: Option<OptimizeLeavesOptions>,
```
roeierez: Since it is anyway fallbacks to Full perhaps it is better to make this non optional?

roeierez (approved):
Other than one comment LGTM

dangeross (commented):
Looks solid

crates/breez-sdk/core/src/sdk/api.rs
danielgranhao: Done

crates/breez-sdk/core/src/sdk/api.rs
danielgranhao: Done

docs/breez-sdk/src/guide/optimize.md
```diff
@@ -32,46 +32,37 @@ Keep multiplicity as low as possible while meeting your performance requirements
...
-<h3 id="start-optimization">
-    <a class="header" href="#start-optimization">Start optimization</a>
-    <a class="tag" target="_blank" href="https://breez.github.io/spark-sdk/breez_sdk_spark/struct.BreezSdk.html#method.start_leaf_optimization">API docs</a>
+<h3 id="optimize-leaves-full">
+    <a class="header" href="#optimize-leaves-full">Run optimization to completion</a>
+    <a class="tag" target="_blank" href="https://breez.github.io/spark-sdk/breez_sdk_spark/struct.BreezSdk.html#method.optimize_leaves">API docs</a>
 </h3>
 
-You can manually trigger the optimization task to start running in the background. If optimization is already running, no new task will be started.
+Call {{#name optimize_leaves}} with an {{#name OptimizeLeavesRequest}} using the default {{#enum OptimizationMode::Full}} mode to run optimization until no further work is productive. The call blocks for the duration of the run and returns an {{#name OptimizeLeavesResponse}} whose {{#name outcome}} is {{#enum OptimizationOutcome::Completed}} with the number of rounds executed. A `rounds_executed` of `0` means the wallet was already optimal at call time.
```
dangeross:
```suggestion
Call {{#name optimize_leaves}} with an {{#name OptimizeLeavesRequest}} using the default {{#enum OptimizationMode::Full}} mode to run optimization until no further work is productive. The call blocks for the duration of the run and returns an {{#name OptimizeLeavesResponse}} whose {{#name outcome}} is {{#enum OptimizationOutcome::Completed}} with the number of rounds executed. A {{#name rounds_executed}} of `0` means the wallet was already optimal at call time.
```

docs/breez-sdk/src/guide/optimize.md
```diff
@@ -32,46 +32,37 @@ Keep multiplicity as low as possible while meeting your performance requirements
...
-    <a class="tag" target="_blank" href="https://breez.github.io/spark-sdk/breez_sdk_spark/struct.BreezSdk.html#method.cancel_leaf_optimization">API docs</a>
+<h3 id="optimize-leaves-single-round">
+    <a class="header" href="#optimize-leaves-single-round">Run optimization one round at a time</a>
+    <a class="tag" target="_blank" href="https://breez.github.io/spark-sdk/breez_sdk_spark/struct.BreezSdk.html#method.optimize_leaves">API docs</a>
 </h3>
 
-You can cancel an ongoing optimization task and wait for it to stop completely. Optimization is done in rounds, and the current round will complete before stopping.
+To display progress or cancel between rounds, pass an {{#name OptimizeLeavesRequest}} with {{#enum OptimizationMode::SingleRound}}. Each call executes one round and the response {{#name outcome}} is {{#enum OptimizationOutcome::InProgress}} (more work remains) or {{#enum OptimizationOutcome::Completed}} (terminal — either the planner confirmed this swap finished optimization, or `rounds_executed == 0` indicating the wallet was already optimal). Cancel between rounds simply by stopping the loop.
```
dangeross:
```suggestion
To display progress or cancel between rounds, pass an {{#name OptimizeLeavesRequest}} with {{#enum OptimizationMode::SingleRound}}. Each call executes one round and the response {{#name outcome}} is {{#enum OptimizationOutcome::InProgress}} (more work remains) or {{#enum OptimizationOutcome::Completed}} (terminal — either the planner confirmed this swap finished optimization, or a {{#name rounds_executed}} of `0` indicates the wallet was already optimal). Cancel between rounds simply by stopping the loop.
```

dangeross (commented):
Just doc nits

dangeross (approved):
LGTM

PR #916, opened 2026-05-28
tree/token tests: source refresh_start from the store clock

danielgranhao:
Four tree-store tests flaked on macOS but not CI: they took `refresh_start`
from the host clock (`SystemTime::now()`) but compared it against `added_at`
timestamps stamped by the DB clock. Under load, Docker's Linux VM
clock can lag the host past the tests' 10ms buffer, so a just-written row
lands before the boundary and is wrongly culled. CI shares the host/container
clock, so it never reproduced there.

Fix: source `refresh_start` from `store.now()` (the DB clock for DB-backed
stores) so the boundary and row timestamps share a clock.

What actually flaked: the four tight-margin (10ms-sleep) tree tests. The
bigger-margin (+10s / -60s) helpers in both suites are switched too purely
for consistency — they never flaked, since their margins dwarf the skew. The
token tests never flaked at all: their tight cases already used `store.now()`,
so only their buffered helpers changed. In-memory callers keep the host clock
(their `store.now()` is just `SystemTime::now()`).

### DIFF
--- a/crates/spark-mysql/src/token_store.rs
+++ b/crates/spark-mysql/src/token_store.rs
@@ -2020,7 +2020,10 @@ mod tests {
         let token_outputs = shared_tests::create_token_outputs(1, vec![100, 200]);
         fixture
             .store
-            .set_tokens_outputs(&[token_outputs], shared_tests::future_refresh_start())
+            .set_tokens_outputs(
+                &[token_outputs],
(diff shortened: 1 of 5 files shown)


JssDWt (approved):

dangeross (approved):
LGTM

PR #917, opened 2026-05-29
Fix unilateral exit on legacy mainnet leaves

dangeross:
Unilateral exit failed on leaves from trees created before 2026-03-17 with `Parent ID … not found in the tree`. The Spark operator intentionally omits the tree root from `query_nodes(include_parents=true)` ancestor expansion for these legacy mainnet trees, so the single bulk response was missing nodes the chain walk needed.

The walk now re-fetches any missing parent **by node ID** (direct by-ID queries aren't subject to the root-skip) and merges it into the node map. It also stops gracefully when it reaches a node outside the exitable set (`Available`/`Splitted`/`OnChain`) instead of erroring. Mirrors the reference Spark TS SDK's `buildUnilateralExitChain`.

### DIFF
--- a/crates/internal/src/main.rs
+++ b/crates/internal/src/main.rs
@@ -99,7 +99,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     // Setup logging
     let log_filter = std::env::var("SPARK_LOG_FILTER")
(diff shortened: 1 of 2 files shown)


danielgranhao (approved):
LGTM

JssDWt (approved):

PR #918, opened 2026-05-29
parse: LUD-01 fallback

JssDWt:
We did not support the [fallback scheme](https://github.com/lnurl/luds/blob/luds/01.md#fallback-scheme) in LUD-01. 

Also we would GET general URLs, which we probably should only do if it's actually identified as an LNURL link.

### DIFF
--- a/crates/breez-sdk/common/src/input/parser/tests.rs
+++ b/crates/breez-sdk/common/src/input/parser/tests.rs
@@ -1017,6 +1017,71 @@ async fn test_http_url_without_lightning_querystring_does_not_match() {
 }
 
 #[async_test_all]
+async fn test_http_url_with_garbage_lightning_querystring_does_not_match() {
+    // `lightning=` values that are not a bech32 LNURL must not resolve. None
+    // of these should reach the network — if any did, MockRestClient would
+    // panic on the missing response queue.
+    let mock_dns_resolver = MockDnsResolver::new();
+    let mock_rest_client = MockRestClient::new();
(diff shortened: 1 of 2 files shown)


crates/breez-sdk/common/src/input/parser/tests.rs
```diff
@@ -958,6 +958,82 @@ async fn test_lnurlp_onion_with_explicit_port_uses_http_scheme() {
...
+
+    let lnurl = "lnurl1dp68gurn8ghj7mr0vdskc6r0wd6z7mrww4excttsv9un7um9wdekjmmw84jxywf5x43rvv35xgmr2enrxanr2cfcvsmnwe3jxcukvde48qukgdec89snwde3vfjxvepjxpjnjvtpxd3kvdnxx5crxwpjvyunsephsz36jf";
+    let wrapper = format!("https://service.example/giftcard/redeem?id=123&lightning={lnurl}");
+
+    let result = input_parser.parse(&wrapper).await;
+    assert!(
+        matches!(result, Ok(InputType::LnurlPay(_))),
+        "expected LnurlPay from LUD-01 lightning= fallback, got: {result:?}"
+    );
+}
```
dangeross, line 978: Is it worth adding a garbage `lightning=` test (non LNURL string)?

danielgranhao (approved):
LGTM

crates/breez-sdk/common/src/input/parser/tests.rs
JssDWt: Added some garbage tests there

dangeross (approved):
LGTM

PR #919, opened 2026-05-29
Refactor payments module into submodules

dangeross:
Splits the monolithic `sdk/payments.rs` into a `payments/` folder module to make the send/receive flow readable and to give the upcoming cross-chain payment type (#799) a clear home. 

Behavior-preserving for all supported payment types. One minor edge change: an *unsupported* input combined with an invalid amount now surfaces "Unsupported payment method" first (previously the amount error).

- **Module split:** `payments.rs` → `payments/` submodules — `prepare/`, `send/`, `receive`, `polling`, `conversion`, `validation`, and a thin `mod.rs` (public API + internal shims). LNURL-pay moves out of `lnurl.rs` into the prepare/send modules.
- **Per-type validation:** each `prepare/<type>` owns its `validate_request` (shared amount / fee-policy checks live in a `validation` module, per-type rules alongside the handler), called from that type's `prepare`.
- **Uniform conversion dispatch:** every `send::<type>::convert_token` owns its purpose + fee shaping, so `execute_pre_send_conversion` collapses to a flat match. Adding a payment type is now "one file in `prepare/`, one in `send/`, two match arms."
- **Pure helpers extracted + unit-tested:** amount/fee-policy validation, dust limits, fee-overpayment, conversion amount-override.
- **Bug fix:** receiving to a Spark invoice with a malformed `sender_public_key` now returns `InvalidInput` instead of panicking.


crates/breez-sdk/core/src/sdk/payments/prepare/bolt11.rs
```diff
@@ -0,0 +1,306 @@
...
+    request: &PrepareSendPaymentRequest,
+    invoice: &Bolt11InvoiceDetails,
+    spark_transfer_fee_sats: Option<u64>,
+    token_identifier: Option<&String>,
+    fee_policy: FeePolicy,
+) -> Result<Option<PrepareSendPaymentResponse>, SdkError> {
+    let Some(token_amount) = request.amount else {
+        return Ok(None);
+    };
+    let (estimated_sats, conversion_estimate) = conversion::estimate_sats_from_token_conversion(
```
roeierez, line 201: I was struggling the use case where we may reach this point having both invoice with amount and user token amount. Intuitively I would suggest returning validation error for such case and only continue here if the invoice is amount-less. am I missing something?

crates/breez-sdk/core/src/sdk/payments/prepare/bolt11.rs
```diff
@@ -0,0 +1,306 @@
...
+        .spark_wallet
+        .extract_spark_address(&request.payment_request)?;
+
+    let spark_transfer_fee_sats = if spark_address.is_some() {
+        Some(0)
+    } else {
+        None
+    };
+
+    if let Some(response) = maybe_prepare_from_token_conversion(
```
roeierez: Here there is actually a new "branch" in the code for the cases:
1. Token provided (FromBitcoin): token_denominated
2. Token not provided (The fall through path):  sats_denominated

I think it would be easier if we explicitly check these conditions here before calling the functions, e.g something like:

```
if is_token_denominated(request, &token_identifier) {
        prepare_token_denominated(sdk, request, detailed_bolt11_invoice,
            spark_transfer_fee_sats, token_identifier, fee_policy).await
    } else {
        prepare_sat_denominated(sdk, request, detailed_bolt11_invoice,
            spark_transfer_fee_sats, token_identifier, fee_policy).await
 }
```

The "maybe" patterns makes me expand that function without really knowing if it is the case I am looking for. 
Same I think for bitcoin_address prepare.

crates/breez-sdk/core/src/sdk/payments/conversion.rs
```diff
@@ -0,0 +1,711 @@
...
+    }
+    let estimate = estimate_conversion(
+        sdk,
+        conversion_options,
+        token_identifier,
+        ConversionAmount::MinAmountOut(amount),
+    )
+    .await?;
+    Ok((amount, estimate))
+}
```
roeierez: Also here I think perhaps better to branch explicitly between token_denominated or sats_denominated at the high level (instead of asking is_token_conversion at a deeper level).

roeierez (changes requested):
Looks really good! a lot clearer.

crates/breez-sdk/core/src/sdk/payments/prepare/bolt11.rs
dangeross: Makes sense

crates/breez-sdk/core/src/sdk/payments/prepare/bolt11.rs
dangeross: Done [93678f2](https://github.com/breez/spark-sdk/pull/919/commits/93678f248b5ecc0d41a439fe027ad4d243820436)

crates/breez-sdk/core/src/sdk/payments/conversion.rs
dangeross: Done [93678f2](https://github.com/breez/spark-sdk/pull/919/commits/93678f248b5ecc0d41a439fe027ad4d243820436)

crates/breez-sdk/core/src/sdk/payments/prepare/mod.rs
```diff
@@ -0,0 +1,123 @@
+mod bitcoin_address;
+mod bolt11;
+pub(super) mod lnurl_pay;
```
JssDWt: Perhaps lnurl_pay doesn't belong in here after all? Should lnurl_pay go inside `lnurl`, and use the functions of `payments` to do its internal things?

crates/breez-sdk/core/src/sdk/payments/conversion.rs
```diff
@@ -0,0 +1,857 @@
...
+//   2. no conversion_estimate → send_internal directly, return
+//   3. acquire stable-balance payment guard (suppresses auto-convert)
+//   4. execute_pre_send_conversion: decide AmountIn vs MinAmountOut, dispatch to
+//      send::<type>::convert_token (per-type fee shape)
+//   5. pre_link_conversion_children (self-transfer only — parent known up front)
+//   6. complete_conversion_and_send: wait for the conversion receive payment,
+//      compute amount_override, send_internal, link children, persist Completed,
+//      fetch payment with conversion details
+//   7. emit payment event (unless suppressed)
+pub(super) async fn orchestrate_send(
```
JssDWt: I feel this function should live in the payments module rather than the conversion module. It orchestrates to either convert or not convert.

JssDWt (commented):
Looks a lot better.

crates/breez-sdk/core/src/sdk/payments/prepare/bolt11.rs
```diff
@@ -0,0 +1,410 @@
...
+
+    prepare_sats_denominated(
+        sdk,
+        request,
+        detailed_bolt11_invoice,
+        spark_transfer_fee_sats,
+        token_identifier,
+        fee_policy,
+    )
+    .await
```
danielgranhao, line 120: Shouldn't we have an if-else here? I saw the rustdocs of `prepare_token_denominated` indicate it returns Ok(None) when conversion can't proceed with conversion, but in that case, we can't treat the amount as sats so we should return error instead of calling `prepare_sats_denominated`

Also, looks like same pattern is used in bitcoin address prepare and on `resolve_send_amount_with_conversion_estimate`. 

Maybe the token path should return explicit errors on converter refusal?

crates/breez-sdk/core/src/sdk/payments/conversion.rs
```diff
@@ -0,0 +1,793 @@
...
+/// Returns whether the prepare used `AmountIn` (user specified the token amount)
+/// rather than `MinAmountOut` (user specified sats).
+///
+/// At prepare time the `AmountIn` path derives `amount` from `estimate.amount_out`
+/// (plus any existing sat balance), so `amount >= amount_out`. The `MinAmountOut`
+/// path guarantees `amount_out >= amount` (strictly greater when there's slack).
+/// With no estimate the send isn't a conversion.
+fn uses_amount_in(amount: u128, conversion_estimate: Option<&ConversionEstimate>) -> bool {
+    conversion_estimate.is_some_and(|e| amount >= e.amount_out)
+}
```
danielgranhao, line 588: Definitely for a potential follow-up (because it requires API changes), but I think this kind of decision reconstruction (which we do as a pattern) is something we should try to eliminate as much as possible. It really hurts readability because it is implicit. Also, it could simplify send significantly. Could be addressed by recording decisions on the prepare response. 

Examples of other reconstructions that would be nice to avoid:
- from_token_identifier
- conversion_amount
- sats_change in compute_amount_override
- is_spark_route
- Fee re-derivation in convert_token

crates/breez-sdk/core/src/sdk/payments/prepare/spark_address.rs
```diff
@@ -0,0 +1,221 @@
...
+    let is_to_bitcoin = matches!(
+        conversion_estimate,
+        Some(ConversionEstimate {
+            options: ConversionOptions {
+                conversion_type: ConversionType::ToBitcoin { .. },
+                ..
+            },
+            ..
+        })
+    );
```
danielgranhao: There's an identical block in the spark invoice prepare. Could be extracted out

crates/breez-sdk/core/src/sdk/payments/send/bitcoin_address.rs
```diff
@@ -0,0 +1,130 @@
...
+    };
+
+    let exit_speed: ExitSpeed = confirmation_speed.clone().into();
+
+    // Calculate fee based on selected speed
+    let fee_sats = match confirmation_speed {
+        OnchainConfirmationSpeed::Fast => fee_quote.speed_fast.total_fee_sat(),
+        OnchainConfirmationSpeed::Medium => fee_quote.speed_medium.total_fee_sat(),
+        OnchainConfirmationSpeed::Slow => fee_quote.speed_slow.total_fee_sat(),
+    };
```
danielgranhao: Could be extracted into a fee_for_speed() to be reused in `convert_token` below

crates/breez-sdk/core/src/sdk/payments/prepare/bitcoin_address.rs
```diff
@@ -0,0 +1,451 @@
...
+        .await?
+        .into();
+
+    let min_fee_sats = fee_quote.speed_slow.total_fee_sat();
+    let output_amount_sats = total_u64.saturating_sub(min_fee_sats);
+    if output_amount_sats < dust_limit_sats {
+        return Err(SdkError::InvalidInput(format!(
+            "Amount is below the minimum of {dust_limit_sats} sats required for this address after lowest fees of {min_fee_sats} sats"
+        )));
+    }
```
danielgranhao: Can we reuse the `validate_dust` helper?

crates/breez-sdk/core/src/sdk/payments/send/spark_address.rs
```diff
@@ -0,0 +1,164 @@
...
+        .parse::<SparkAddress>()
+        .map_err(|_| SdkError::InvalidInput("Invalid spark address".to_string()))?;
+    let purpose = if spark_address.identity_public_key == sdk.spark_wallet.get_identity_public_key()
+    {
+        ConversionPurpose::SelfTransfer
+    } else {
+        ConversionPurpose::OngoingPayment {
+            payment_request: address.to_string(),
+        }
+    };
```
danielgranhao: Can be shared with similar block in spark invoice convert_token

crates/breez-sdk/core/src/sdk/payments/prepare/spark_address.rs
```diff
@@ -0,0 +1,221 @@
...
+
+/// Validates a spark address request.
+fn validate_request(request: &PrepareSendPaymentRequest) -> Result<(), SdkError> {
+    validation::validate_amount(request.amount)?;
+    validation::validate_fee_policy_for_conversion(request)?;
+
+    // Amount is required for spark addresses
+    if request.amount.is_none() {
+        return Err(SdkError::InvalidInput("Amount is required".to_string()));
+    }
```
danielgranhao: Here we validate amount then immediately validate it again. Later in `prepare` we validate once again right after calling `validate_request`. Maybe `validate_request` can return the validated amount, so we can consolidate these checks?

Same pattern on bitcoin address prepare

crates/breez-sdk/core/src/sdk/lnurl/pay.rs
```diff
@@ -0,0 +1,404 @@
...
+        sdk.config.network.into(),
+        request.validate_success_action_url,
+    )
+    .await?
+    {
+        ValidatedCallbackResponse::EndpointError { data } => {
+            return Err(LnurlError::EndpointError(data.reason).into());
+        }
+        ValidatedCallbackResponse::EndpointSuccess { data } => data,
+    };
```
danielgranhao: Repeated in `prepare_fees_included`. Can be extracted.

danielgranhao (commented):
Awesome! Mainly small mechanical things like extracting duplicates. Also, a comment suggesting a follow-up that I think would be valuable.

danielgranhao (approved):
LGTM

crates/breez-sdk/core/src/sdk/payments/conversion.rs
```diff
@@ -0,0 +1,909 @@
...
+/// that need a defensive lower bound (e.g. LNURL invoice sizing on the `AmountIn`
+/// path) should apply their own slippage buffer.
+pub(in crate::sdk) async fn estimate_sats_from_token_conversion(
+    sdk: &BreezSdk,
+    conversion_options: Option<&ConversionOptions>,
+    token_identifier: Option<&String>,
+    amount: u128,
+    fee_policy: FeePolicy,
+) -> Result<(u128, Option<ConversionEstimate>), SdkError> {
+    let (is_token_conversion, is_send_all) = is_token_conversion(
```
roeierez: I think also here we should move this line to the callers and then only enter this flow if we are in token conversion flow. Will be easier to reason about the flow.

roeierez (changes requested):
Other than one comment LGTM

roeierez (approved):
LGTM

PR #920, opened 2026-05-29
Forward debug/trace logs to external loggers

danielgranhao:
The external logger layer hard-capped events at INFO, so debug/trace were silently dropped regardless of the configured `log_filter` (only the file log and WASM honored it).

Removes the cap and applies the filter per-layer, matching the file layer and WASM. External loggers now honor `log_filter` (default `debug`). Adds a regression test.

### DIFF
--- a/crates/breez-sdk/core/src/logger.rs
+++ b/crates/breez-sdk/core/src/logger.rs
@@ -1,5 +1,5 @@
 use std::fs::OpenOptions;
-use tracing::{Event, Level, Subscriber};
+use tracing::{Event, Subscriber};
 use tracing_subscriber::{
     EnvFilter, Layer,
     fmt::{FormatFields, format::Writer},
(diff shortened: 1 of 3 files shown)


crates/breez-sdk/core/src/logger.rs
```diff
@@ -48,11 +46,12 @@ pub(super) fn init_logging(
 ) -> Result<(), SdkError> {
     let filter = log_filter.unwrap_or(DEFAULT_FILTER);
```
JssDWt, line 71: Can you change DEFAULT_FILTER to

```
const DEFAULT_FILTER: &str = concat!(
    "info",
    // First-party crates: keep debug logging.
    ",breez_sdk_spark=debug",
    ",breez_sdk_common=debug",
    ",breez_sdk_spark_wasm=debug",
    ",breez_sdk_spark_bindings=debug",
    ",spark=debug",
    ",spark_wallet=debug",
    ",spark_postgres=debug",
    ",spark_mysql=debug",
    ",flashnet=debug",
    ",platform_utils=debug",
    // Noisy third-party crates: silence below warn.
    ",h2=warn",
    ",rustls=warn",
    ",rustyline=warn",
    ",hyper=warn",
    ",hyper_util=warn",
    ",tower=warn",
    ",Connection=warn",
    ",tonic=warn",
);
```

danielgranhao: Done

dangeross (commented):
Looks good. Do we need to adjust the Wasm defaults also?

dangeross (approved):
LGTM. Just a CI fmt issue

JssDWt (approved):
LGTM apart from the format issue

PR #922, opened 2026-06-01
Adds an env-gated mainnet itest suite

dangeross:
Adds `breez-itest` mainnet tests for the conversion paths that have no reliable regtest pool liquidity, and fixes a flashnet reverse-quote off-by-one surfaced while building it.

### Flashnet fix

`calculate_amount_in` flooring three "scale up then divide" sites (slippage, output-fee, input-fee) returned an `amount_in` that came up 1–3 sats short of the user's target on its own forward quote (e.g. LNURL pay 750 sats failing with `Validation returned 752 but expected at least 753`). All three now use the existing `div_ceil` helper. Added a round-trip regression test pinning `forward(calculate_amount_in(target)) >= target` across small targets and slippages.

### Mainnet itests

Gated tests in `crates/breez-sdk/breez-itest/tests/`, gated on `MAINNET_TEST_MNEMONIC` + `BREEZ_API_KEY` (skip with a warning otherwise, so normal CI is unaffected):

- **`mainnet_token_conversion`**
  - Success: Bitcoin→Token (Spark address) + Token→Bitcoin (Bolt11).
  - Failure: non-existent token rejected at prepare; insufficient funds rejected at send.
  - Spark-invoice conversion: Bob creates a Spark **invoice** asking for USDB, Alice pays with explicit \`FromBitcoin\`. Covers the \`prepare/spark_invoice.rs\` + \`send/spark_invoice.rs\` paths from #919.
- **`mainnet_stable_balance`**
  - Batch auto-convert: two sub-threshold payments accumulate past the threshold, triggering \`auto_convert\`.
  - Auto-fill spend: Bob pays Alice's Bolt11 invoice from his USDB via the auto-filled Token→BTC path (no explicit conversion options).
  - Per-receive auto-convert: a single payment ≥ \`pool_min\` triggers \`per_receive_convert\` on the incoming payment directly (separate code path from batch).
  - Deactivation: \`update_user_settings(StableBalanceActiveLabel::Unset)\` drives \`deactivation_convert\`, draining Bob's USDB back to sats.
- **`mainnet_server_mode_conversion`** — Bitcoin→Token (Spark address) + Token→Bitcoin (Bolt11).
- **`mainnet_teardown`** — send-all-with-conversion drains Bob → Alice as sats. Doubles as a real test of the send-all-with-conversion path.

Wallets: env mnemonic = "Alice" (funder, keeps her funds); "Bob" is derived deterministically (fixed BIP-39 passphrase) so funds stranded by a crash are recovered by the next teardown.

### CI

New `.github/workflows/breez-itest-mainnet.yaml` triggered by `workflow_dispatch` (and `workflow_call`) with `tests` and `token_id` inputs. Runs the conversion tests, then the teardown as a final `if: always()` step so funds are returned even when tests fail. Single-threaded via a dedicated `make breez-itest-mainnet` target.

### Test plan

- [x] `cargo test -p flashnet --lib calculate_amount_in` — 14 tests pass (13 existing + new round-trip).
- [x] `make fmt-check` / `cargo clippy -D warnings` clean across `flashnet`, `breez-sdk-spark`, `breez-sdk-itest`.
- [x] `cargo test -p breez-sdk-itest --test mainnet_*` skips cleanly when env unset.
- [x] Live mainnet run: all 9 tests pass (`mainnet_token_conversion` ×3, `mainnet_stable_balance` ×3, `mainnet_server_mod_conversion` ×2, `mainnet_teardown`).


danielgranhao (commented):
Looks good! Did you consider adding a Lightning address ToBitcoin conversion test? I think it may be worth it because it relies on separate conversion logic and it should be fine for these to use the mainnet lnurl server

dangeross:
It's currently costing around ~26 sats per full run

danielgranhao (approved):
LGTM

roeierez (approved):

PR #924, opened 2026-06-01
higher-level spark signer to prepare for Turnkey integration

JssDWt:
This PR reworks how the SDK handles signing so it can plug into external signing backends like **Turnkey**.

**A higher-level "Spark signer" shaped like Turnkey.** Previously the SDK signed Spark operations through a low-level, primitive-by-primitive signing interface. We introduced a higher-level signer whose operations map directly onto the things a Spark wallet actually does — preparing transfers, claims, deposits, Lightning receives, FROST signing, and so on — mirroring the surface Turnkey exposes. The entire Spark wallet now goes through this higher-level signer; the old low-level signer survives only as the default in-process implementation, wrapped behind an adapter. A Turnkey-style backend can now implement the wallet's signing flows directly instead of reimplementing low-level cryptographic primitives.

**Two distinct external signers instead of one.** The SDK previously exposed a single external signer that mixed everything together. We split it into two clearly-scoped external interfaces:

- a **Breez signer** for the general SDK-level signing the app needs (LNURL auth, sync, message signing, encryption), and
- a **Spark signer** for the Spark wallet flows.

Integrators now provide both when supplying their own signing backend, and the seed-based default builds both automatically — so existing seed-based usage is unchanged. These two interfaces are available across all the SDK's language bindings (native, Go/Kotlin/Swift/Python, and WASM/TypeScript).


crates/breez-sdk/wasm/src/sdk_builder.rs
```diff
@@ -299,12 +303,8 @@ impl SdkBuilder {
 
     #[wasm_bindgen(js_name = "withKeySet")]
     pub fn with_key_set(mut self, config: crate::models::KeySetConfig) -> Self {
-        self.key_set_type = config.key_set_type.clone().into();
-        self.use_address_index = config.use_address_index;
         self.account_number = config.account_number;
         let core_config = breez_sdk_spark::KeySetConfig {
-            key_set_type: config.key_set_type.into(),
-            use_address_index: config.use_address_index,
```
dangeross, line 307: One minor issue here is a non-typed language might accept `keySetType`/`useAddressIndex` being passed and expecting the same behavior/wallet

dangeross (commented):
Looks good on first pass, need to run over it again. 
The docs `customizing.md` also need updating removing the key set type descriptions

crates/breez-sdk/breez-itest/src/helpers.rs
```diff
@@ -801,15 +801,23 @@ pub async fn build_sdk_from_mnemonic(
...
+/// Build SDK instance from a mnemonic.
+///
+/// The external signer interface is now split into two foreign traits
+/// (`ExternalBreezSigner` + `ExternalSparkSigner`). `default_external_signer`
+/// only produces the `ExternalBreezSigner` half; there is no default
+/// `ExternalSparkSigner` factory (a stateless wrapper cannot satisfy the
+/// transfer/claim flows, which the integrator must back with their own tree
+/// state). This helper therefore builds the SDK from a seed, which constructs
+/// both signers internally.
 ///
```
danielgranhao: nit: seems unnecessary here in itest helpers


```suggestion
```

crates/spark-itest/src/helpers.rs
```diff
@@ -33,11 +34,15 @@ pub async fn build_test_wallet(
...
+    let spark_signer = Arc::new(SparkSignerAdapter::new(signer.clone()));
+    let mut builder = WalletBuilder::new(config, spark_signer);
     match backend {
         Backend::InMemory => {}
         Backend::Postgres(conn_str) => {
-            let identity = signer.get_identity_public_key().await?.serialize();
+            let identity = signer
+                .derive_public_key(&"m/0'".parse::<bitcoin::bip32::DerivationPath>()?)
+                .await?
+                .serialize();
```
danielgranhao: can't we just use `spark_signer.get_identity_public_key().await?` ?

crates/spark-itest/src/helpers.rs
```diff
@@ -48,7 +53,10 @@ pub async fn build_test_wallet(
                 .with_token_output_store(Arc::new(token));
         }
         Backend::Mysql(conn_str) => {
-            let identity = signer.get_identity_public_key().await?.serialize();
+            let identity = signer
+                .derive_public_key(&"m/0'".parse::<bitcoin::bip32::DerivationPath>()?)
+                .await?
+                .serialize();
```
danielgranhao: same here

crates/spark/src/services/coop_exit.rs
```diff
@@ -283,26 +292,6 @@ impl CoopExitService {
...
-            .await?;
-
-        // 2. ECIES-encrypt the key tweaks per operator.
-        let encrypted_key_tweaks = self
-            .transfer_service
-            .encrypt_key_tweaks(&key_tweak_input_map)?;
-        let key_tweak_package: HashMap<String, Vec<u8>> = encrypted_key_tweaks
-            .into_iter()
-            .map(|(k, v)| (hex::encode(k.serialize()), v))
-            .collect();
```
danielgranhao, line 304: nit: numbered comments that follow should be updated since 1 and 2 were removed

crates/spark/src/signer/default_signer.rs
```diff
@@ -38,240 +31,58 @@ fn account_number(network: Network) -> u32 {
...
 
 #[derive(Clone)]
 pub struct DefaultSigner {
-    key_set: KeySet,
+    /// The master node every key is derived from.
+    master: Xpriv,
+    /// Path of the key used for ECIES (encrypt-to-self, decrypt). This is the
+    /// key counterparties encrypt secrets to; the Spark layer derives the wallet
+    /// identity at the same path.
+    encryption_path: DerivationPath,
```
danielgranhao: nit: Seems like this field isn't needed? It's always set to `identity_path()`. Could either be a const or `identity_path()` can be inlined at the two read sites.

crates/spark/src/signer/spark_signer.rs
```diff
@@ -0,0 +1,460 @@
...
+//! layer talks to. Its methods mirror the Turnkey Spark activity surface, so a
+//! Turnkey backend is a thin adapter on top, while the in-process default
+//! ([`SparkSignerAdapter`](super::SparkSignerAdapter)) does the same work
+//! locally over the low-level [`Signer`](super::Signer) trait.
+//!
+//! The trait is **derivation-path-agnostic**: methods speak in Spark concepts
+//! (leaf id, static-deposit index, transfer/claim) and never in BIP-32 paths.
+//! Each implementation resolves those concepts to its own key material —
+//! Turnkey inside its enclave, the default adapter by reproducing the exact
+//! key derivation we use today.
```
danielgranhao: nit: reference to "today" will not age well :)

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -89,7 +89,11 @@ pub struct ConnectRequest {
 #[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
 pub struct ConnectWithSignerRequest {
     pub config: Config,
-    pub signer: std::sync::Arc<dyn crate::signer::ExternalSigner>,
+    /// External signer for non-Spark SDK signing (LNURL-auth, sync, message
+    /// signing, ECIES).
+    pub signer: std::sync::Arc<dyn crate::signer::ExternalBreezSigner>,
```
danielgranhao: Maybe `breez_signer` for symmetry with `SdkBuilder::new_with_signer`?

crates/breez-sdk/breez-itest/src/helpers.rs
danielgranhao: Ah, and the method is called `build_sdk_with_external_signer`, which is used by some itests that were covering the external signer path. I think either we implement proper external signing here or we remove `build_sdk_with_external_signer` and the itests that rely on it. Wdyt?

docs/breez-sdk/src/guide/external_signer.md
danielgranhao: This needs to be updated right?

docs/breez-sdk/src/guide/external_signer.md
```diff
@@ -8,11 +8,13 @@ The External Signer feature allows you to provide custom signing logic for the S
 
 ## Using the Default External Signer
```
danielgranhao: This section is reading a bit weird because IIUC it should be about using defaults but in fact there's no default for `ExternalSparkSigner` and `Implementing a Custom Signer` ends up being mandatory instead of optional

danielgranhao (commented):
Looks very good! Just minor things. Also, there's a rustdoc in `crates/breez-sdk/core/src/bindings.rs` lines 68-70 that still mentions the removed key set type and index flag.

crates/spark-itest/src/helpers.rs
JssDWt: Yes

danielgranhao (commented):
Other than the failing docs/cli CI jobs, LGTM

danielgranhao (approved):

crates/spark/src/signer/spark_signer.rs
```diff
@@ -0,0 +1,436 @@
...
+/// Discriminator for the kind of token-transaction signature requested. Lets
+/// policy-enforcing signers gate freeze vs. spend separately.
+#[derive(Debug, Clone, Copy)]
+pub enum TokenTransactionKind {
+    /// Issuer-side signature on a freeze message.
+    Freeze,
+    /// Owner-side signature on a partial token transaction (`compute_hash`).
+    Partial,
+    /// Owner-side signature on a finalized token transaction
+    /// (`SHA256(SHA256(tx_hash || op_pubkey_hash))`).
```
dangeross: I think this doc formula doesn't match what `create_per_operator_signatures` computes, where it hashes each input separately and concatenates the digests: `SHA256(SHA256(tx_hash) || SHA256(op_pubkey))`. As written here it reads as a single inner hash over the raw concatenation, `SHA256(SHA256(tx_hash || op_pubkey_hash))`, which is a different digest.

crates/breez-sdk/wasm/src/sdk_builder.rs
dangeross: Wdyt about this? An option might be renaming the `with_key_set` fn to force migration

crates/breez-sdk/wasm/src/signer.rs
```diff
@@ -686,404 +369,855 @@ impl breez_sdk_spark::signer::ExternalSigner for WasmExternalSigner {
...
+    pub async fn sign_message(&self, message: Vec<u8>) -> Result<EcdsaSignatureBytes, JsValue> {
+        self.inner
+            .sign_message(message)
+            .await
+            .map(Into::into)
+            .map_err(default_spark_err)
+    }
+
+    #[wasm_bindgen(js_name = "signFrost")]
+    pub async fn sign_frost(&self, jobs: JsValue) -> Result<JsValue, JsValue> {
```
dangeross: Is there a reason to have `JsValue` input/result here and not `Vec<Type>`? I think I've seen it used elsewhere in wasm

dangeross (commented):
Overall LGTM. 

There is a stale rustdoc on `with_key_set` in `crates/breez-sdk/core/src/bindings.rs` that references key set type and address index flag

crates/breez-sdk/wasm/src/signer.rs
JssDWt: Fixed

crates/breez-sdk/wasm/src/sdk_builder.rs
JssDWt: Went with with_account_number instead. Removed all references to keyset

crates/breez-sdk/wasm/src/signer.rs
```diff
@@ -686,404 +369,854 @@ impl breez_sdk_spark::signer::ExternalSigner for WasmExternalSigner {
...
-    ) -> Result<core_types::PublicKeyBytes, SignerError> {
-        let promise = self
+    #[wasm_bindgen(structural, method, js_name = "signEcdsaRecoverable", catch)]
+    pub fn sign_ecdsa_recoverable(
+        this: &JsExternalBreezSigner,
+        message: MessageBytes,
+        path: String,
+    ) -> Result<Promise, JsValue>;
+
+    #[wasm_bindgen(structural, method, js_name = "eciesEncrypt", catch)]
```
roeierez: Shouldn't this be encryptEcies ? Also same for decryptEcies?

crates/breez-sdk/core/src/signer/external_spark_types.rs
```diff
@@ -0,0 +1,766 @@
...
+    ) -> Result<Self, SdkError> {
+        Ok(Self {
+            index: r.index,
+            user_statement: r.user_statement.clone(),
+        })
+    }
+}
+
+/// FFI-safe representation of `spark_wallet::PreparedStaticDepositClaim`.
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
roeierez, line 749: Perhaps omit the Debug attribute as we don't want to log the secret?

crates/spark/src/signer/spark_signer_adapter.rs
```diff
@@ -0,0 +1,625 @@
...
+            transfer_id: _,
+            sender_identity_public_key: _,
+            leaves,
+            operator_recipients,
+            threshold,
+        } = request;
+
+        let mut per_operator: BTreeMap<Identifier, Vec<proto::ClaimLeafKeyTweak>> = BTreeMap::new();
+
+        for leaf in &leaves {
```
roeierez, line 378: The trait doc states that this function should also verify the sender signature (per leaf). Currently we don't to that here (we do that on the claim side). I think nothing prevents us from verifying this here as well.

crates/spark/src/signer/spark_signer_adapter.rs
```diff
@@ -0,0 +1,625 @@
...
+            threshold,
+        } = request;
+
+        // Per-operator accumulator of this transfer's leaf key tweaks.
+        let mut per_operator: BTreeMap<Identifier, Vec<proto::SendLeafKeyTweak>> = BTreeMap::new();
+        let mut new_leaf_keys = Vec::with_capacity(leaves.len());
+
+        for leaf in &leaves {
+            let signing_key = SecretSource::Derived(signing_path(&leaf.node.id)?);
+            let new_signing_key = SecretSource::Derived(signing_path(&leaf.new_leaf_id)?);
```
roeierez, line 266: Previously the new key was random. Just curious what is the advantage of this new key generation?

roeierez (commented):
Looks good!

crates/spark/src/signer/spark_signer_adapter.rs
JssDWt: This came from integrating turnkey: https://github.com/breez/spark-sdk/pull/943/changes#diff-9668147cdd17fb00e0e90bcaa4bf80b78498038fc4c0fb199227f91f0517512eR574

The turnkey call takes a `new_leaf_derivation` as a parameter. This change here brings the interface in line with that signer, by giving control of what key to generate to the caller.

crates/spark/src/signer/spark_signer_adapter.rs
JssDWt: The sender signature is verified by a higher level, that's not necessarily a responsibility of the signer to verify imo. It is verified in `verify_pending_transfer`, called by `try_claim_transfer` before calling `claim_transfer_with_leaves` which leads to this function eventually.

The turnkey signer does verify this signature in the signer. 

Updated the trait doc.

crates/breez-sdk/core/src/signer/external_spark_types.rs
JssDWt: Made SecretBytes's Debug redact the secret bytes.

crates/breez-sdk/wasm/src/signer.rs
JssDWt: Done

roeierez (approved):
LGTM

dangeross (approved):
LGTM

PR #925, opened 2026-06-02
Avoid per-build migration contention

danielgranhao:
When there is latency between the server and the DB, a multi-tenant deployment can easily run into contention issues because every instance build competes for the same global DB locks.


roeierez (approved):
LGTM

dangeross (approved):
LGTM

PR #926, opened 2026-06-02
fix: unify WASM byte types on Uint8Array

erdemyerebasmaz:
## Problem

The WASM passkey surface had two byte representations and a structural gap:

1. **Two byte shapes.** tsify-generated response/request types used `number[]`; the hand-written `PrfProvider` boundary used `Uint8Array`. This split required the duplicate boundary types `PasskeyCredentialJSON`, `DeriveSeedsResultJSON`, and `DeriveSeedsOptionsJSON`, and pushed `number[]`-to-`Uint8Array` conversions onto integrators.
2. **No-cast conformance gap.** The built-in `PasskeyProvider` did not structurally satisfy `PrfProvider`, so injecting it forced a cast (`new PasskeyClient(new PasskeyProvider(...) as any, ...)`).

## Changes

**Unify the WASM byte types on `Uint8Array`.** `PasskeyCredential`'s byte fields serialize via `serde_bytes` (serde-wasm-bindgen emits `Uint8Array`) with a tsify type override so the generated `.d.ts` matches. `RegisterRequest.excludeCredentials` / `SignInRequest.allowCredentials` take only the tsify override, since the deserializer already reads a JS `Uint8Array` into a `Vec<u8>`.

**Result**: one shape both internal and external, no `*JSON` duplicates, no app-side `number[]`/`Uint8Array` conversions, and `PasskeyProvider` satisfies `PrfProvider` with **no cast**. The WASM snippet drops its `as any`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -780,6 +780,7 @@ dependencies = [
  "platform-utils",
  "serde",
  "serde-wasm-bindgen",
+ "serde_bytes",
  "spark-wallet",
  "tracing",
  "tracing-subscriber",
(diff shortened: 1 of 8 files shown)


roeierez (approved):
LGTM

PR #927, opened 2026-06-03
Token conversion: Harden recovery against silent failures

dangeross:
Recovery from a failed token conversions was gated on DB writes between `transfer_asset` and `clawback`. If that write failed (storage glitch, app backgrounded, process killed) the refunder never saw the row and the funds stayed locked at the pool indefinitely.

The error path also threw away the rich `AssetTransfer` returned from `execute_swap`, leaving `convert()`'s error branch to re-derive the local `payment_id` via an operator round-trip — the same network outage that caused the swap to fail would also fail that derivation, skipping the local `RefundNeeded` mark entirely.

### Fix

Two layers of recovery:

#### Recovery via local payment metadata

`FlashnetError::Execution` now carries the `AssetTransfer`. `convert()`'s error branch derives the storage `payment_id` locally — no operator round-trip. The existing `RefundNeeded` marker path is unchanged from there on.

#### Flashnet listing as a catch-all

New `list_clawback_transfers` binding + `reconcile_with_flashnet` pass picks up transfers Flashnet still considers clawback-eligible but that don't have a local marker (storage write failed, process killed). A 5-minute age threshold guards against racing a concurrent same-seed instance's in-flight healthy swap.

- **Client mode**: runs once per SDK init, after the local refunder pass.
- **Server mode**: runs per call inside `refund_pending_conversions`.

### DIFF
--- a/crates/breez-sdk/core/src/token_conversion/flashnet.rs
+++ b/crates/breez-sdk/core/src/token_conversion/flashnet.rs
@@ -2,21 +2,26 @@ use std::{collections::HashMap, str::FromStr, sync::Arc, time::Duration};
 
 use bitcoin::secp256k1::PublicKey;
 use flashnet::{
-    AssetTransfer, BTC_ASSET_ADDRESS, CacheStore, ClawbackRequest, ClawbackResponse,
+    AssetTransfer, BTC_ASSET_ADDRESS, CacheStore, ClawbackRequest, ClawbackTransfer,
     ExecuteSwapRequest, ExecuteSwapResponse, FlashnetClient, FlashnetConfig, FlashnetError,
-    GetMinAmountsRequest, ListPoolsRequest, PoolSortOrder, SimulateSwapRequest,
+    GetMinAmountsRequest, ListClawbackTransfersRequest, ListPoolsRequest, PoolSortOrder,
+    SimulateSwapRequest,
 };
+use platform_utils::time::{SystemTime, UNIX_EPOCH};
 use spark_wallet::{SparkWallet, TransferId};
 use tokio::sync::broadcast;
 use tracing::{debug, error, info, warn};
 
 use crate::{
     AmountAdjustmentReason, EventEmitter, Network, Payment, PaymentDetails, PaymentMetadata,
-    Storage,
+    RefundPendingConversionsResponse, Storage,
     persist::{StorageListPaymentsRequest, StoragePaymentDetailsFilter},
     token_conversion::{ConversionAmount, DEFAULT_CONVERSION_MAX_SLIPPAGE_BPS},
     utils::{
-        payments::{fetch_and_process_payment, insert_payment_with_metadata},
+        payments::{
+            fetch_and_process_payment, insert_payment_with_metadata,
+            resolve_and_insert_payment_metadata,
+        },
         polling::{PollSchedule, poll_until},
     },
 };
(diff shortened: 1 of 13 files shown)


crates/breez-sdk/core/src/token_conversion/flashnet.rs
```diff
@@ -36,6 +38,27 @@ const RECEIVED_LEG_POLL_INITIAL_DELAY_MS: u64 = 500;
...
+/// Reconcile listing page size. Not paginated — subsequent inits pick up any
+/// overflow.
+const RECONCILE_LISTING_LIMIT: u32 = 100;
+
+/// Missing/unparseable timestamps fall back to old-enough/recent respectively.
+fn transfer_is_older_than(transfer: &ClawbackTransfer, cutoff_secs: u64) -> bool {
+    let Some(created_at) = transfer.created_at.as_deref() else {
+        return true;
+    };
+    let Ok(parsed) = chrono::DateTime::parse_from_rfc3339(created_at) else {
```
roeierez: Just wanted to make sure this is the correct date format because it seems that in the test we are using something different: https://github.com/breez/spark-sdk/blob/main/crates/flashnet/src/models.rs#L841

crates/breez-sdk/core/src/token_conversion/flashnet.rs
```diff
@@ -120,111 +143,206 @@ impl FlashnetTokenConverter {
...
+                warn!("Clawback for {clawback_id} not accepted: {:?}", r.error);
+                false
+            }
+            Err(e) => {
+                error!("Clawback for {clawback_id} failed: {e}");
+                return Err(e.into());
+            }
+        };
+
+        // Clawback already settled; resolution failure here only loses
```
roeierez: We fall through here also when outcome=false above and seems like we mark the payment is as refunded bellow.  Is that what we want?

crates/breez-sdk/core/src/token_conversion/flashnet.rs
```diff
@@ -950,8 +1106,43 @@ impl TokenConverter for FlashnetTokenConverter {
...
+            Err(e) => {
+                warn!("Local refund pass failed: {e}");
+                RefundPendingConversionsResponse::default()
+            }
+        };
+        let remote = match self.reconcile_with_flashnet().await {
+            Ok(r) => r,
+            Err(e) => {
+                warn!("Reconcile with Flashnet failed: {e}");
+                RefundPendingConversionsResponse::default()
```
roeierez, line 1099: Seems that if we have an error in both passes it is not propagated upstream? Thinking about server mode that invoked refund_pending and gets (0,0) as response. No indication of any error here.
Should we add a failures to the response?

roeierez (changes requested):

crates/breez-sdk/core/src/token_conversion/flashnet.rs
dangeross: Test data was incorrect. Orchestra uses RFC3339 formatted dates

crates/breez-sdk/core/src/token_conversion/flashnet.rs
dangeross: Added a `failed` field to the response struct

roeierez (approved):
LGTM

danielgranhao (approved):
LGTM

PR #928, opened 2026-06-03
Set sane DB TCP liveness params

danielgranhao:
## Problem

A managed Postgres/MySQL instance behind a NAT/load balancer may silently drop
idle TCP connections (no FIN/RST). The connection lingers half-open in the
pool: the recycle check sees it as live and hands it out, and the next query
blocks on the dead socket until the kernel gives up (~15 min), surfacing as a
long hang then `os error 110`. 

## Fix

Apply TCP keepalive defaults so the NAT mapping stays warm and a dead peer is
detected promptly. Each value is applied only when the connection string left
it unset, so explicit URL params always win.

**Postgres** (`tokio-postgres`): `keepalives_idle` 60s, `keepalives_interval`
10s, `keepalives_retries` 3, `tcp_user_timeout` 30s.

**MySQL** (`mysql_async`): `tcp_keepalive` 60s. The driver's default is no
keepalive at all, and the pool doesn't ping on checkout. `mysql_async` 0.36.2
exposes no probe interval/retries or `tcp_user_timeout`, so idle time is the
only liveness knob available.

### DIFF
--- a/crates/spark-mysql/src/pool.rs
+++ b/crates/spark-mysql/src/pool.rs
@@ -39,7 +39,7 @@ pub fn create_pool(config: &MysqlStorageConfig) -> Result<Pool, MysqlError> {
     let opts: Opts = Opts::from_url(&connection_string)
         .map_err(|e| MysqlError::Initialization(format!("Invalid connection string: {e}")))?;
 
-    let mut builder = OptsBuilder::from_opts(opts);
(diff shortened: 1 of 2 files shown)


dangeross (approved):
LGTM

roeierez (approved):
LGTM

PR #930, opened 2026-06-03
Upgrade boltz-client (USDC destinations)

danielgranhao:
Boltz now supports USDC alongside USDT/USDT0

### DIFF
--- a/crates/breez-sdk/core/src/cross_chain/boltz.rs
+++ b/crates/breez-sdk/core/src/cross_chain/boltz.rs
@@ -12,7 +12,7 @@ use std::time::Duration;
 use boltz_client::{
     BoltzError, BoltzService as BoltzClient,
     config::{BoltzConfig as BoltzClientConfig, MAX_SLIPPAGE_BPS},
-    models::{ChainId, PreparedSwap},
+    models::{Asset, PreparedSwap},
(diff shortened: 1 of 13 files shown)


crates/breez-sdk/core/src/cross_chain/boltz.rs
```diff
@@ -565,6 +583,19 @@ fn boltz_err_to_sdk(err: &BoltzError) -> SdkError {
     SdkError::Generic(format!("Boltz: {err}"))
 }
 
+/// Map a route's asset ticker back to the boltz-client [`Asset`]. Inverse of
+/// `Asset::as_str` (used when building the route): `CrossChainRoutePair.asset`
+/// is a plain string, so the prepare path parses it to select the destination
+/// by `(chain, asset)`. Returns `None` for tickers Boltz does not deliver.
+fn asset_from_ticker(ticker: &str) -> Option<Asset> {
```
JssDWt: Should this be a `TryFrom` in the boltz client crate? So the logic to add more assets stays in one place? Or use `supported_destinations` on `BoltzService`?

danielgranhao: Done

PR #931, opened 2026-06-04
refactor(LNURL): dedicated authorize/accept methods for atomic lightning address transfer

erdemyerebasmaz:
**TL;DR:** replaces #829's `accept_lightning_address_transfer` and overloaded `register_lightning_address` with a dedicated `authorize` / `accept` pair, before #829 lands.

### API
```rust
// current owner: grant the transfer
authorize_lightning_address_transfer(
    AuthorizeTransferRequest { transferee_pubkey },
) -> TransferAuthorization

// new owner: complete the takeover
accept_lightning_address_transfer(
    AcceptTransferRequest { authorization, description },
) -> LightningAddressInfo
```
Renamed `accept_lightning_address_transfer` to `authorize_lightning_address_transfer` and the optional `transfer` field on `register_lightning_address` is replaced with `accept_lightning_address_transfer`. `username` now lives inside `TransferAuthorization`, so the shared blob is self-contained.

### Why

**1. "accept" was misleading: the behavior is "authorize."** The current owner runs it to *grant* a transfer, not accept one, and both the code and #829 already say so:
> Produce a transfer authorization [...] granting the right to take it over to `transferee_pubkey`  (code doc comment)
>
> [...] the authorization that grants B the right to take over the username  (#829 description)

**2. Keep each method to one job.** Folding the transfer into `register_lightning_address` via an optional `transfer` field gives one method two jobs: with the field set, a registration call quietly becomes a takeover, changing which endpoint is hit, what gets signed, and what the server does. A dedicated pair gives each method a single purpose and one reason to change:
- `register_lightning_address` does one job again: first-time registration.
- The transfer flow is discoverable as a unit (`authorize` + `accept` sit together).
- One honest return type each, no overloaded request/response.

Wire protocol unchanged, so no server changes.

#### Misc.

Applied the Comment Style guidelines to the lightning address transfer comments, and dropped "single atomic operation" from the high-level docs (the guide keeps it). The Comment Style section in `CLAUDE.md` landed on `main` after most of the atomic LN address transfer feature (#829) was written.


crates/breez-sdk/cli/src/command/mod.rs
erdemyerebasmaz, line 1: I will split CLI command into two dedicated commands.

erdemyerebasmaz, line 1: I do not like some of the wording related to SDK & wallet: 
```
/// Authorize transferring the lightning address username currently
/// registered on this SDK to another wallet.

/// Run on the *current owner's* SDK.

/// Errors with [SdkError::Generic] if no lightning address is registered
/// on this SDK.

/// Run on the *new owner's* SDK with the
```

They indicate whether something is running on the current owner's SDK instance or the new owner's SDK instance, but I think the current/new owner abstraction works better at this level.

crates/breez-sdk/core/src/lnurl.rs
```diff
@@ -41,7 +41,17 @@ impl std::fmt::Display for LnurlServerError {
 pub struct RegisterLightningAddressRequest {
     pub username: String,
     pub description: String,
-    pub transfer: Option<lnurl_models::LightningAddressTransfer>,
```
dangeross, line 44: Claude is saying `lnurl_models::LightningAddressTransfer` is now unused and can be removed

dangeross (commented):
Looks good

crates/breez-sdk/core/src/lnurl.rs
erdemyerebasmaz: Great catch. Addressing it on the parent PR.

PR #932, opened 2026-06-04
experiment: generic wasm postgres implementation 

JssDWt:
No need to maintain a separate postgres implementation in wasm if we can port types from tokio-postgres to wasm implementations.


JssDWt:
Closing due to inactivity. Branch is alive at https://github.com/JssDWt/spark-sdk/tree/jssdwt-postgres-wasm

PR #933, opened 2026-06-04
test(passkey): make client tests hermetic, fixing flaky Wasm CI

erdemyerebasmaz:
## Problem

The **Wasm test** CI job flaked intermittently on `main` and every branch off it. The passkey client unit tests drove the real Nostr label store, so `register` / `sign_in` opened live websockets to public relays (`relay.primal.net`, `relay.damus.io`, ...) during `cargo test --features browser-tests`. In the headless-Firefox `wasm-bindgen-test` runner those calls intermittently stalled past the 20s per-test timeout, the browser was SIGKILL'd, and the job failed.

Evidence it is the network, not the code:
- Same SHA `3dab84f61` passed in [run 26972088526](https://github.com/breez/spark-sdk/actions/runs/26972088526) and failed in others.
- In a passing run the only slow tests were the three network-touching ones (register 3.6s, connect-fallback 2.8s, default_label 0.5s); the other 179 ran in ~1s.
- In a failing run `register_returns_credential_and_publishes_label` was the last test invoked and never completed (20s gap, then SIGKILL).

## Fix

Put the label store behind a `LabelStore` trait (mirroring the existing `PrfProvider` mock pattern) and build it through an injectable `store_builder` on `Passkey`. **Production behavior is unchanged**: the default builder still constructs `NostrSaltClient`. Tests inject an in-memory `MockLabelStore`, so the whole passkey suite now runs with zero network (32 tests in 0.02s, was 3.6s+ for register alone).

## Coverage

The mock records store interactions, closing gaps the test names promised but never verified:
- **register**: asserts the label is actually published, and the kept `aaguid` / `backup_eligible` attestation round-trips.
- **fast-path sign-in**: asserts the store is never queried and nothing is re-published (discovery sign-in is the positive control).
- **sign-in**: asserts a bare `credential_id` with no attestation.

## Verification
- 32/32 passkey tests pass in 0.02s (native, `--features passkey`).
- Tests compile for `wasm32-unknown-unknown --features browser-tests`.
- fmt + clippy clean (native + wasm); default/native builds unaffected (`mod passkey` is feature-gated).

## Note (out of scope)
The same dependency means `register()` blocks in production on the best-effort label publish for up to ~90s worst case when relays stall. This PR fixes the tests only; whether that publish should be backgrounded or bounded is worth a separate look.

### DIFF
--- a/crates/breez-sdk/core/src/passkey/mod.rs
+++ b/crates/breez-sdk/core/src/passkey/mod.rs
@@ -72,7 +72,21 @@ use tracing::warn;
 
 use crate::Seed;
 use derivation::derive_nostr_keypair;
-use nostr_client::NostrSaltClient;
+use nostr_client::{LabelStore, NostrSaltClient};
+
+/// Builds the per-identity [`LabelStore`] from the Nostr keys derived in a
+/// PRF ceremony plus the optional Breez API key. The default builds a
+/// network-backed [`NostrSaltClient`]; tests inject an in-memory double so
+/// unit tests never reach the relays.
+type LabelStoreBuilder =
+    Arc<dyn Fn(nostr::Keys, Option<String>) -> Arc<dyn LabelStore> + Send + Sync>;
+
+/// Default store builder: a network-backed [`NostrSaltClient`].
+fn default_label_store_builder() -> LabelStoreBuilder {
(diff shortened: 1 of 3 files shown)


dangeross (approved):
LGTM

PR #934, opened 2026-06-05
wire execute_before through token transfers and fulfill_spark_invoice

peterrojs:
threads execute_before_unix_micros through TokenService::transfer_tokens and its inner builder into the spark_token_primitives::TransferBuildRequest, replacing the hardcoded None. fulfill_spark_invoice reads the token invoice's expiry_time and sets execute_before on the resulting token transaction, so the SO rejects the transfer once the deadline lapses.

this lets a receiver issue a spark token invoice with an expiry_time and have the payer's wallet fulfill it with a signed deadline, allowing longer quote windows while keeping output locking bounded per fulfillment attempt.

the public SparkWallet::transfer_tokens signature is unchanged. non invoice callers (burn, consolidation, the send pipeline, flashnet, the internal cli, itests) pass None, so the deadline is opt in and scoped to invoice fulfillment. the expiry to micros conversion is factored into execute_before_from_expiry with unit tests for the absent, in range, and pre epoch cases.

the deadline itself is enforced operator side once execute_before_unix_micros is set on the transaction, so there is no rust side rejection path to unit test beyond the conversion.

### DIFF
--- a/crates/spark-wallet/src/wallet.rs
+++ b/crates/spark-wallet/src/wallet.rs
@@ -1366,7 +1366,7 @@ impl SparkWallet {
 
         let tx = self
             .token_service
-            .transfer_tokens(outputs, selected_outputs, selection_strategy)
+            .transfer_tokens(outputs, selected_outputs, selection_strategy, None)
             .await?;
         Ok(tx)
     }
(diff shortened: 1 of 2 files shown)


dangeross (approved):
LGTM

PR #939, opened 2026-06-08
docs(snippets): wrap long snippet lines + line-length guidance

erdemyerebasmaz:
## What

Wraps anchored documentation-snippet lines that ran past ~100 columns so the rendered mdbook docs don't clip on narrow viewports, and adds a CLAUDE.md rule to keep them short going forward.

## Approach

No custom linter, no new xtask command. Line length is handled the simplest way:

- **Native formatters** cap width where they support it: Rust (`cargo fmt`), Dart (`dart format`), Swift (`swift format`), Python (pylint `C0301`). Rely on those.
- **Go, C#, TypeScript, Kotlin** have no native line-wrapping, so this PR does a one-time hand-wrap of their over-long anchored lines, and the new CLAUDE.md rule asks future edits to keep them short on a best-effort basis.

## Changes

- One-time wrap of long anchored lines across the bindings. Line-break-only, semantics preserved: object/argument lists split one element per line, trailing comments moved above their statement, long log/print strings split via each language's concatenation. LNURL/invoice literals are single unbreakable tokens and are left as-is.
- `CLAUDE.md`: a "Snippet Line Length" rule.


dangeross (approved):
Just the C# CI job failing, otherwise LGTM

erdemyerebasmaz:
**Reshaped.** Dropped the custom `check-snippet-line-length` lint and the `format-doc-snippets` command discussed above. A detector that only fails CI (leaving the wrapping to a human) isn't worth it, and native formatters can't enforce width for Go/C#/TS/Kotlin anyway. This PR is now just the one-time wraps + a CLAUDE.md rule; earlier comments about the lint/stack are historical.

PR #941, opened 2026-06-09
Refresh leaves ony query for available

roeierez:
Turns out sent leaves that are unclaimed by the receiver still return as part of the user's leaves which on some cases (large scale send payments) increases the leaves storage a lot!
Since we anyway filter only available leaves before inserting into the store we can instead query for only available statuses from the first place.


dangeross (approved):
LGTM

JssDWt (approved):

PR #942, opened 2026-06-09
Cross-chain: docs

dangeross:
Document the USD stablecoin payment flow (Orchestra + Boltz): the opt-in `cross_chain_config`, parsing recipient addresses into `InputType::CrossChainAddress`, discovering routes, preparing and sending. Tight integration into the existing `config.md`, `parse.md`, and `send_payment.md` pages, plus a new `cross_chain.md` deep-dive (providers, supported address formats, slippage, quote expiry, status lifecycle, retry safety, limitations). New snippet anchors across all 9 languages.

### snippets-processor

Fix the `{{#name}}` macro for dotted identifiers with a lowercase prefix. `{{#name conversion_info.delivered_amount}}` was rendering as `conversion_info.deliveredAmount` for camelCase / PascalCase languages; only the trailing segment was case-converted while the prefix stayed snake_case. The processor now detects the prefix shape — uppercase = type name (emitted verbatim), lowercase = snake_case field (case-converted per language) — so it renders as `conversionInfo.deliveredAmount` for Swift/Kotlin/JS, `ConversionInfo.DeliveredAmount` for Go/C#, and unchanged `conversion_info.delivered_amount` for Rust/Python. Also fixes pre-existing references like `leaf_optimization_config.auto_enabled` in `config.md` / `server_mode.md`.

### DIFF
--- a/docs/breez-sdk/snippets/rust/src/config.rs
+++ b/docs/breez-sdk/snippets/rust/src/config.rs
@@ -134,3 +134,19 @@ pub(crate) fn configure_background_tasks() -> Result<()> {
     info!("Config: {:?}", config);
     Ok(())
 }
+
+pub(crate) fn configure_cross_chain() -> Result<()> {
+    // ANCHOR: cross-chain-config
+    let mut config = default_config(Network::Mainnet);
+    config.api_key = Some("<breez api key>".to_string());
+
+    // Override the default slippage tolerance (basis points; 10 to 500).
+    // Set cross_chain_config to None to disable the feature.
+    config.cross_chain_config = Some(CrossChainConfig {
+        default_slippage_bps: Some(50),
+        default_target_overpay_bps: None,
+    });
+    // ANCHOR_END: cross-chain-config
+    info!("Config: {config:?}");
+    Ok(())
+}
--- a/docs/breez-sdk/snippets/rust/src/config.rs
+++ b/docs/breez-sdk/snippets/rust/src/config.rs
@@ -140,8 +140,7 @@ pub(crate) fn configure_cross_chain() -> Result<()> {
     let mut config = default_config(Network::Mainnet);
     config.api_key = Some("<breez api key>".to_string());
 
-    // Override the default slippage tolerance (basis points; 10 to 500).
-    // Set cross_chain_config to None to disable the feature.
+    // Set to enable cross-chain payments. Slippage override is optional (10 to 500 bps).
     config.cross_chain_config = Some(CrossChainConfig {
         default_slippage_bps: Some(50),
         default_target_overpay_bps: None,
--- a/docs/breez-sdk/snippets/rust/src/cross_chain.rs
+++ b/docs/breez-sdk/snippets/rust/src/cross_chain.rs
@@ -71,6 +71,7 @@ async fn send_payment_cross_chain(
     prepare_response: PrepareSendPaymentResponse,
 ) -> Result<()> {
     // ANCHOR: cross-chain-send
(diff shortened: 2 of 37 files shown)


docs/breez-sdk/src/guide/cross_chain.md
```diff
@@ -0,0 +1,121 @@
...
+1. **Pass a caller-supplied {{#name idempotency_key}}** on {{#name SendPaymentRequest}}. The top-level dispatcher first looks for an existing payment with that id and short-circuits the retry if found; otherwise the key is used as the Spark transfer id.
+2. **Omit {{#name idempotency_key}}** — the SDK derives a deterministic UUIDv5 from the provider's quote/swap id. Re-sending the same prepared shape produces the same id and dedupes at the Spark protocol layer even if the first attempt's persistence step never completed.
+
+### USDB-source sends
+
+For sends where the source is USDB (Orchestra), the underlying token transfer does not accept an idempotency token at the Spark protocol layer. There is no upstream dedup. Callers that retry on transient errors can fire a second deposit and overpay.
+
+This matches the existing contract for direct token sends.
+
+If you need at-most-once semantics for a USDB-source cross-chain send, debounce retries at the application layer until the SDK either returns a payment or a terminal error.
```
roeierez: isn't it more accurate to say that the idempotency depends on whether the first leg is token conversion? Even if using Boltz there may still be a token conversion (sending from usdb balance) which seems to not support idempotency.

roeierez (approved):
One comment otherwise LGTM

docs/breez-sdk/snippets/rust/src/config.rs
```diff
@@ -134,3 +134,19 @@ pub(crate) fn configure_background_tasks() -> Result<()> {
...
     Ok(())
 }
+
+pub(crate) fn configure_cross_chain() -> Result<()> {
+    // ANCHOR: cross-chain-config
+    let mut config = default_config(Network::Mainnet);
+    config.api_key = Some("<breez api key>".to_string());
+
+    // Override the default slippage tolerance (basis points; 10 to 500).
+    // Set cross_chain_config to None to disable the feature.
```
dangeross: `default_config` leaves `cross_chain_config` unset, so cross-chain is opt-in. This frames it as opt-out. Can we lead with the enable path instead?

```suggestion
    // Cross-chain is opt-in: default_config leaves cross_chain_config unset.
    // Set it to enable the feature, optionally overriding the slippage default (10 to 500 bps).
```

Same comment in all 9 language config snippets (`wasm/config.ts`, `go/config.go`, `swift/.../Config.swift`, etc.).

docs/breez-sdk/snippets/rust/src/cross_chain.rs
```diff
@@ -0,0 +1,86 @@
...
+    sdk: &BreezSdk,
+    prepare_response: PrepareSendPaymentResponse,
+) -> Result<()> {
+    // ANCHOR: cross-chain-send
+    let optional_idempotency_key = Some("<idempotency key uuid>".to_string());
+    let send_response = sdk
+        .send_payment(SendPaymentRequest {
+            prepare_response,
+            options: None,
+            idempotency_key: optional_idempotency_key,
```
dangeross, line 80: Retry safety says a token-leg cross-chain send rejects a caller-supplied `idempotency_key` with `InvalidInput`, so this errors for a USDB-source or auto-converted route. Worth a note that the key only applies to no-token-leg (BTC source) sends, or dropping it from the cross-chain example to avoid the trap?

Same in the other 8 language `cross_chain` snippets.

PR #943, opened 2026-06-09
Turnkey integration

JssDWt:
## Add Turnkey as a signing backend

Lets an app run the Spark SDK with its keys held in Turnkey (a managed key-custody service) instead of on the device. Signing happens remotely inside Turnkey's secure enclave, so the wallet's private keys never live in the app.
It is fully opt-in: seed-based wallets are unchanged.

A new `create_turnkey_signer(...)` builds a Turnkey-backed signer from your Turnkey credentials. Pass it to the SDK like any other signer, and every wallet operation (receive, send, Lightning, on-chain deposits, tokens) then signs through Turnkey.

### Design choices

- **Keys stay in the enclave.** The SDK never holds the wallet's signing keys;
  every signature is requested from Turnkey. Note: two narrow operations (see next point) are the only places a key is taken out, and neither key is one used to move funds.
- **The two unavoidable exceptions use a dedicated key.** A couple of operations (refunding an on-chain deposit, and the SDK's own data encryption) need a key the app can use locally. Rather than expose a wallet key, we export a separate key reserved for this, so no key used to move funds ever leaves the enclave.
- **Drop-in, not a rewrite.** Turnkey plugs into the SDK's existing signer interface, so the rest of the SDK is untouched.
- **One implementation for every platform.** A single minimal network client works on both native and web (WASM) builds.
- **Flexible credentials.** Both of Turnkey's API-key types are supported, including their default, so users need not generate a special key.
- **Resilient by default.** Transient failures and rate limits are retried with backoff and honor Turnkey's retry hints. Long-running operations are tracked by id rather than re-sent, so a slow operation is never run twice.
- **Off unless requested.** The whole backend sits behind a feature flag, so default builds pull in none of it.

### Testing

The integration tests run the same end-to-end flows against both the seed backend and Turnkey, verifying the two behave identically: wallet info/address, on-chain deposit, send/receive, Lightning receive, deposit refund, and tokenmint.

For isolation, each Turnkey test creates its own throwaway wallet and deletes it afterward. Cleanup only removes wallets it created, and the leftover sweep is age-gated, so multiple people or CI jobs can run the tests at once without deleting each other's wallets. Tests skip cleanly when Turnkey credentials are
not configured.

### Test environment variables

The Turnkey integration cases (`cargo test -p breez-sdk-itest --test signer_backends turnkey`) read the variables below. If any required Turnkey variable is missing, those cases skip cleanly (logged, not failed), so the suite still runs locally without credentials.

| Variable | Required | Default | Purpose |
|---|---|---|---|
| `TURNKEY_ORG_ID` | Yes | (none) | Organization (or sub-organization) id that owns the wallets the tests create. |
| `TURNKEY_API_PUBLIC_KEY` | Yes | (none) | Public half of the Turnkey API key (hex). secp256k1 or P-256 (Turnkey's default); the curve is detected automatically. |
| `TURNKEY_API_PRIVATE_KEY` | Yes | (none) | Private half of the API key (hex); signs every request. |
| `TURNKEY_BASE_URL` | No | `https://api.turnkey.com` | Turnkey API endpoint. |


crates/breez-sdk/core/src/turnkey/management.rs
```diff
@@ -0,0 +1,213 @@
+//! Turnkey wallet lifecycle helpers (create / delete / list).
+//!
+//! Not part of the integrator-facing API: integrators bring their own wallet
+//! (see [`super::create_turnkey_signer`]). These exist so a test harness can
+//! provision a throwaway wallet per test and reap abandoned ones, keeping each
+//! test isolated. The configured API key authenticates at the organization
+//! level, so one key can create and delete many wallets in that org.
```
danielgranhao: It is exposed in rust api right? Maybe gate it behind #[cfg(test)] or a feature?

crates/breez-sdk/core/src/turnkey/accounts.rs
```diff
@@ -0,0 +1,207 @@
...
+    let start = 32usize.saturating_sub(bytes.len());
+    out[start..].copy_from_slice(&bytes);
+    Ok(out)
+}
+
+pub(crate) fn ecdsa_from_rs(r_hex: &str, s_hex: &str) -> Result<ecdsa::Signature, TurnkeyError> {
+    let mut compact = [0u8; 64];
+    compact[..32].copy_from_slice(&decode_scalar_32(r_hex)?);
+    compact[32..].copy_from_slice(&decode_scalar_32(s_hex)?);
+    ecdsa::Signature::from_compact(&compact).map_err(|e| TurnkeyError::Deserialize(e.to_string()))
```
danielgranhao: should we normalize to low-S here so that we match the seed backend behavior?

crates/breez-sdk/core/src/turnkey/breez_signer.rs
```diff
@@ -0,0 +1,187 @@
...
+            .sign_raw(sign_with, hex::encode(&message.bytes), HASH_FUNCTION_NO_OP)
+            .await
+            .map_err(to_signer_err)?;
+        // Output layout (per the trait): [31 + recovery_id] || r(32) || s(32).
+        // Turnkey reports the recovery id in `v`; the exact encoding is verified
+        // at integration time (Ethereum-style 27/28 is normalized here).
+        let recovery_id = hex::decode(&result.v)
+            .ok()
+            .and_then(|b| b.last().copied())
+            .map_or(0, |b| if b >= 27 { b.saturating_sub(27) } else { b });
```
danielgranhao: Shouldn't we error instead of defaulting to 0?

danielgranhao (commented):
Looking good

crates/breez-sdk/core/src/turnkey/accounts.rs
JssDWt: I think that's up to turnkey to normalize to low-S on the backend if needed?

crates/breez-sdk/core/src/turnkey/management.rs
JssDWt: gated behind test-utils (and turnkey)

crates/breez-sdk/core/src/turnkey/breez_signer.rs
```diff
@@ -0,0 +1,195 @@
...
+            .client
+            .sign_raw(sign_with, hex::encode(&message.bytes), HASH_FUNCTION_NO_OP)
+            .await
+            .map_err(to_signer_err)?;
+        // Output layout (per the trait): [31 + recovery_id] || r(32) || s(32).
+        // Turnkey reports the recovery id in `v` (hex), possibly Ethereum-style
+        // (27/28), normalized here. The signature recovers to the wrong key
+        // under a wrong recovery id, so a missing or out-of-range `v` is an
+        // error rather than a guessed value.
+        let v = u32::from_str_radix(result.v.trim(), 16).map_err(|e| {
```
dangeross, line 127: Worth Turnkey-specific tests for `sign_ecdsa_recoverable`/`sign_ecdsa`/`hmac_sha256`? Today they only have seed-specific tests (lnurl_auth, rtsync), and signer_backends runs with RTSync and LNURL off.

danielgranhao (approved):
LGTM

crates/breez-sdk/core/src/turnkey/config.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// Network the wallet operates on; selects the Spark address format
+    /// (mainnet or regtest) used for Spark-protocol and Schnorr signing.
+    pub network: crate::Network,
+    /// Spark account number: the `{account}` in every derivation path
+    /// (`m/8797555'/{account}'/...`). Unset uses the network default, matching
+    /// the seed-based signer, so the same wallet seed derives the same keys on
+    /// either backend.
+    pub account_number: Option<u32>,
+    /// Retry policy for Turnkey requests.
+    pub retry: TurnkeyRetryConfig,
```
dangeross: `retry` is the one field with a sensible default, unlike the credentials and ids. Worth making it `Option` (default when unset)? Saves callers, especially on wasm, from spelling out the whole block just to take the defaults.

crates/breez-sdk/core/src/turnkey/factory.rs
```diff
@@ -0,0 +1,88 @@
...
+fn to_signer_err<E: std::fmt::Display>(e: E) -> SignerError {
+    SignerError::Generic(e.to_string())
+}
+
+/// The two Turnkey-backed signers, ready to pass to the SDK's signer-based
+/// connect: `breez` for non-Spark SDK signing, `spark` for the Spark wallet.
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct TurnkeySigners {
+    pub breez: Arc<dyn ExternalBreezSigner>,
+    pub spark: Arc<dyn ExternalSparkSigner>,
```
dangeross: `ExternalSigners` names these as `breez_signer` and `spark_signer`. Should we align them (either way)?

crates/breez-sdk/core/src/turnkey/breez_signer.rs
```diff
@@ -0,0 +1,195 @@
...
+        let recovery_id = u8::try_from(normalized)
+            .ok()
+            .filter(|id| *id <= 3)
+            .ok_or_else(|| {
+                SignerError::Generic(format!("Turnkey recovery id '{}' out of range", result.v))
+            })?;
+        let mut bytes = Vec::with_capacity(65);
+        bytes.push(31u8.saturating_add(recovery_id));
+        bytes.extend_from_slice(&decode_scalar_32(&result.r).map_err(to_signer_err)?);
+        bytes.extend_from_slice(&decode_scalar_32(&result.s).map_err(to_signer_err)?);
```
dangeross: Is it possible to do low-s normalized recoverable sigs here? `ecdsa_from_rs` now normalizes, but this path passes Turnkey's `s` through raw.

dangeross (commented):
Looking good. 

There is a stale comment in `crates/breez-sdk/core/src/session_store/encrypting.rs` that "receiver keypair is a child of the wallet's identity_master_key"

crates/breez-sdk/core/src/turnkey/factory.rs
JssDWt: Removed this struct, as it was the same as ExternalSigners.

dangeross (approved):
LGTM

crates/breez-sdk/core/src/turnkey/stamp.rs
```diff
@@ -0,0 +1,162 @@
...
+    ) -> Result<Self, TurnkeyError> {
+        let priv_bytes = hex::decode(private_key_hex)
+            .map_err(|e| TurnkeyError::InvalidApiKey(format!("private key hex: {e}")))?;
+
+        // secp256k1 (always available).
+        if let Ok(secret_key) = SecretKey::from_slice(&priv_bytes) {
+            let secp = Secp256k1::new();
+            let public_key = secret_key.public_key(&secp);
+            let derived = hex::encode(public_key.serialize());
+            if expected_public_key_hex.is_empty()
```
roeierez: public key can be empty?

crates/breez-sdk/core/src/turnkey/transport.rs
```diff
@@ -0,0 +1,380 @@
...
+                    return Err(TurnkeyError::Http {
+                        status: resp.status,
+                        body: resp.body,
+                    });
+                }
+                Err(e) if attempt < self.retry.max_retries => {
+                    attempt = attempt.saturating_add(1);
+                    let delay = self.retry.delay_for_attempt(attempt);
+                    if !retry_within_budget(started.elapsed(), delay, timeout) {
+                        return Err(TurnkeyError::Transport(e.to_string()));
```
roeierez, line 224: for 409 status (already exists) we will return error here. This I guess may happen in "lost requests" as result of client network issues, etc... Do we have a way to poll for this request and return its idempotent response rather than return error in this case?

roeierez (changes requested):
Looks great!

crates/breez-sdk/core/src/turnkey/transport.rs
JssDWt: We do this for `create_account` and `create_wallet` already. I'll have a look at other cases.

crates/breez-sdk/core/src/turnkey/stamp.rs
JssDWt: No, fixed

crates/breez-sdk/core/src/turnkey/transport.rs
JssDWt: Added retry for the other cases as well. The ones that can generally be recovered by retrying.

roeierez (approved):
LGTM

PR #946, opened 2026-06-11
Defer wallet background start until runtime subscribes

dangeross:
Wallet restores on the [cross-chain PR](https://github.com/breez/spark-sdk/pull/799) can show balance 0 for up to 60s after `connect()` returns, even when callers use `getInfo({ ensureSynced: true })`. Identified in the cross-chain PR / Boltz initialization, a race window has been identified in the SDK startup ordering enough to make the race fire reliably in WASM sessions.

### Root cause

`SparkWallet`'s `BackgroundProcessor` is spawned inside `WalletBuilder::build()`, but the SDK runtime loop that subscribes to wallet events runs much later, at the end of `BreezSdk::init_and_start`. If the operator stream connects in that window, `EventManager::notify_listeners(WalletEvent::Synced)` fires with no subscribers and `tokio::sync::broadcast` drops the send (it does not retain messages for late subscribers). Without that first `Synced`, nothing triggers the initial sync until the 60s periodic timer, and `initial_synced_watcher` (which `ensure_synced` blocks on) stays false the whole time.

A real session log shows it firing:

```
[44.249Z] Received connected event                             ← operator stream connects
[44.488Z] Received event: Connected
[44.488Z] Failed to send wallet event, no listeners attached   ← StreamConnected lost
[44.584Z] Failed to send wallet event, no listeners attached   ← Synced lost
[44.900Z] Boltz service initialized
[44.905Z] Initialized and started breez sdk.                   ← subscriber attaches, 321ms too late
```

Both `WalletEvent::StreamConnected` and `WalletEvent::Synced` are dropped before the SDK's runtime loop reaches `spark_wallet.subscribe_events()`.

The race has existed since the SDK adopted the deferred-subscribe pattern. The gap between `build_spark_wallet` and `spawn_client_runtime_loop` was usually short enough that the runtime subscribed before the operator stream connected. The cross-chain init adds awaited HTTP work (`api.boltz.exchange/v2/chain/contracts` fetch plus WebSocket handshake) into that gap, pushing the subscription past the ~330ms operator-connect time on consistent network conditions. WASM single-threaded scheduling makes the new timing reproducible session-to-session.

### Fix

Defer the background processor start until after the runtime subscribes.

- `spark-wallet`: extract the spawn into `SparkWallet::start_background_processing(&self, cancellation_token)`. `WalletBuilder` callers using the default `with_background_processing(true)` (spark-itest) are unchanged.
- `breez-sdk-spark`: `build_spark_wallet` passes `.with_background_processing(false)`. `ClientRuntime::start_sdk_services` calls `start_background_processing` immediately after `spawn_client_runtime_loop`, by which point `wallet_events` and `sync_requests` receivers are established. `ServerRuntime` unchanged.

### DIFF
--- a/crates/spark-wallet/src/wallet.rs
+++ b/crates/spark-wallet/src/wallet.rs
@@ -165,10 +165,23 @@ macro_rules! with_leafs_spent_retry {
 }
 
 pub struct SparkWallet {
-    /// Cancellation token sender for internal use. When dropped, background tasks will stop.
-    /// This is `Some` only when no external cancellation token was provided.
+    /// Sender held for lifetime management when no external cancellation token
+    /// was provided to the builder. Dropping it cancels the receiver clones in
+    /// background tasks, so they exit when the wallet is dropped.
     #[allow(dead_code)]
     cancel: Option<watch::Sender<()>>,
+    /// Receiver moved into the background tasks on the first call to
+    /// [`Self::start_background_processing`]. Wrapped in `Mutex<Option<_>>` so
+    /// that first call can `take()` it: keeping a permanent receiver on `self`
+    /// would prevent [`watch::Sender::closed`] (used by the SDK's `disconnect`)
+    /// from ever firing while the wallet is alive. Tied to either the external
+    /// token passed via [`WalletBuilder::with_cancellation_token`] or the
+    /// internal sender in [`Self::cancel`].
+    cancellation_token: tokio::sync::Mutex<Option<watch::Receiver<()>>>,
+    /// Single-flight guard for [`Self::start_background_processing`] so callers
+    /// can invoke it repeatedly without spawning duplicate operator
+    /// subscriptions, refresh tasks, or auto-optimizer loops.
+    background_started: tokio::sync::OnceCell<()>,
     config: SparkWalletConfig,
     deposit_service: Arc<DepositService>,
     event_manager: Arc<EventManager>,
@@ -212,7 +225,6 @@ impl SparkWallet {
             None,
             None,
             None,
-            true,
             None,
         )
         .await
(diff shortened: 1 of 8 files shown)


danielgranhao (approved):
LGTM

crates/spark-wallet/src/wallet.rs
```diff
@@ -1316,6 +1299,33 @@ impl SparkWallet {
         self.event_manager.listen()
     }
 
+    /// Spawns the operator event stream, leaf/token refresh, and auto-optimizer.
+    /// Callers that disable `with_background_processing` on the builder are
+    /// responsible for invoking this once their event subscribers are attached;
+    /// the first `WalletEvent::Synced` is otherwise dropped if no listener is
+    /// connected yet.
+    pub async fn start_background_processing(&self, cancellation_token: watch::Receiver<()>) {
```
roeierez: Since now this is public method perhaps we don't need the `with_background_processing` flag anymore?

roeierez (changes requested):

crates/spark-wallet/src/wallet.rs
dangeross: I kept for backwards compatibility if others use the spark-wallet. But can remove

roeierez: I thought so but TBH I think it is better to be strict here and document a small breaking change.

roeierez: Now that this method is public users can call it more than one time while before it was internal mechanism.  Do we want to address that?

JssDWt (approved):

crates/spark-wallet/src/wallet.rs
dangeross: The `with_background_processing` flag and SdkBuilder fn is removed

dangeross: Is now behind a OnceCell

roeierez (approved):
LGTM

PR #949, opened 2026-06-11
Store deposit tx vout

### LINKED ISSUE #944: Expose output_index (vout) on PaymentDetails::Deposit
Deposit payments returned by `listPayments` / `getPayment` only expose the transaction id of the claimed static deposit: https://github.com/breez/spark-sdk/blob/main/crates/breez-sdk/core/src/models/mod.rs#L408

A single transaction can fund the same static deposit address at multiple output indexes so there's no way to distinguish payments.

Conversely, the `SPARK_STATIC_DEPOSIT_FINISHED` webhooks payloads do contain the output index https://sdk-doc-spark.breez.technology/guide/webhooks.html#static-deposit-finished

I think the output_index is dropped by an adaptor here https://github.com/breez/spark-sdk/blob/2c80c32db4489447af4c5600c39c272d6e8845d2/crates/breez-sdk/core/src/models/adaptors.rs#L201 

Would it be possible to include and store the output index going forward?

dangeross:
Closes #944

### DIFF
--- a/crates/breez-sdk/core/src/models/adaptors.rs
+++ b/crates/breez-sdk/core/src/models/adaptors.rs
@@ -199,6 +199,7 @@ impl PaymentDetails {
             },
             SspUserRequest::ClaimStaticDeposit(request) => PaymentDetails::Deposit {
                 tx_id: request.transaction_id.clone(),
+                vout: u32::try_from(request.output_index).ok(),
             },
         };
 
--- a/crates/breez-sdk/core/src/models/adaptors.rs
+++ b/crates/breez-sdk/core/src/models/adaptors.rs
@@ -198,7 +198,19 @@ impl PaymentDetails {
                 conversion_info: None,
             },
             SspUserRequest::ClaimStaticDeposit(request) => {
-                let vout = u32::try_from(request.output_index).unwrap_or_default();
+                let Ok(vout) = u32::try_from(request.output_index) else {
+                    // vout=0 is a valid output index, so we can't safely default.
+                    // Returning Ok(None) mirrors the early return for a missing
+                    // user_request above: the downstream branch records the deposit
+                    // as Pending without details and the sync loop keeps advancing
+                    // the offset. Erroring here would deadlock the whole sync until
+                    // the offending transfer leaves the SSP's response.
+                    warn!(
+                        "Invalid output_index {} in SspUserRequest::ClaimStaticDeposit; skipping details",
+                        request.output_index
+                    );
+                    return Ok(None);
+                };
                 PaymentDetails::Deposit {
                     tx_id: request.transaction_id.clone(),
                     vout,
--- a/crates/breez-sdk/core/src/models/adaptors.rs
+++ b/crates/breez-sdk/core/src/models/adaptors.rs
@@ -197,10 +197,13 @@ impl PaymentDetails {
                 htlc_details: None,
                 conversion_info: None,
             },
-            SspUserRequest::ClaimStaticDeposit(request) => PaymentDetails::Deposit {
-                tx_id: request.transaction_id.clone(),
-                vout: u32::try_from(request.output_index).ok(),
-            },
+            SspUserRequest::ClaimStaticDeposit(request) => {
+                let vout = u32::try_from(request.output_index).unwrap_or_default();
+                PaymentDetails::Deposit {
+                    tx_id: request.transaction_id.clone(),
+                    vout,
+                }
+            }
         };
 
         Ok(Some(details))
--- a/crates/breez-sdk/core/src/models/mod.rs
+++ b/crates/breez-sdk/core/src/models/mod.rs
@@ -406,6 +406,7 @@ pub enum PaymentDetails {
     },
     Deposit {
         tx_id: String,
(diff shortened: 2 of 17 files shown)


crates/breez-sdk/core/src/persist/postgres/storage.rs
```diff
@@ -1750,7 +1767,8 @@ const SELECT_PAYMENT_SQL: &str = "
            lrm.sender_comment AS lnurl_sender_comment,
            lrm.payment_hash AS lnurl_payment_hash,
            pm.conversion_status,
-           pm.parent_payment_id
+           pm.parent_payment_id,
+           p.deposit_vout
```
JssDWt: nit: let's put it next to p.deposit_tx_id?

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -406,6 +406,7 @@ pub enum PaymentDetails {
     },
     Deposit {
         tx_id: String,
+        vout: Option<u32>,
```
JssDWt: Could we make this non-optional if we'd drop existing data and resync?

crates/breez-sdk/core/src/persist/mysql/storage.rs
```diff
@@ -469,6 +469,24 @@ impl MysqlStorage {
...
+            // repopulates them with the real vout from the SSP user_request payload:
+            // vout=0 is a valid output index, so we can't safely default-backfill it.
+            // Reset the bitcoin sync offset to trigger that resync.
+            vec![
+                Migration::AddColumn {
+                    table: "brz_payments",
+                    column: "deposit_vout",
+                    definition: "INT UNSIGNED NULL",
+                },
+                Migration::sql("DELETE FROM brz_payments WHERE deposit_tx_id IS NOT NULL"),
```
JssDWt: If these migrations are in reverse order, we can make deposit_vout non-nullable.

dangeross: Would you have it default to 0 for non deposit payments?

crates/breez-sdk/core/src/models/adaptors.rs
```diff
@@ -197,9 +197,13 @@ impl PaymentDetails {
                 htlc_details: None,
                 conversion_info: None,
             },
-            SspUserRequest::ClaimStaticDeposit(request) => PaymentDetails::Deposit {
-                tx_id: request.transaction_id.clone(),
-            },
+            SspUserRequest::ClaimStaticDeposit(request) => {
+                let vout = u32::try_from(request.output_index).unwrap_or_default();
```
danielgranhao: Unlikely to bite, but anyway I think it's worth returning an error here instead of defaulting to 0

danielgranhao (commented):
Just a nit, otherwise LGTM

JssDWt (commented):
Agree with Daniel's comment. Then LGTM

danielgranhao (approved):

JssDWt (approved):
LGTM

crates/breez-sdk/core/src/models/adaptors.rs
dangeross: Seems throwing an error here would brick the sync, so I'm returning None instead

crates/breez-sdk/core/src/persist/mysql/storage.rs
```diff
@@ -469,6 +469,33 @@ impl MysqlStorage {
...
+                Migration::sql(
+                    "CREATE TABLE IF NOT EXISTS brz_payment_details_deposit (
+                        user_id VARBINARY(33) NOT NULL,
+                        payment_id VARCHAR(255) NOT NULL,
+                        tx_id VARCHAR(255) NOT NULL,
+                        vout INT UNSIGNED NOT NULL,
+                        PRIMARY KEY (user_id, payment_id)
+                    )",
+                ),
+                Migration::sql("DELETE FROM brz_payments WHERE deposit_tx_id IS NOT NULL"),
```
roeierez: This will create temporarily missing payments when user call list_payments. To solve that we can copy the deposit payments into the new table instead of deleting. Do you think we should do that?

roeierez (changes requested):
Looks good. One potential migration issue.

crates/breez-sdk/core/src/persist/mysql/storage.rs
dangeross: We don't know the vout's of these deposit txs, this is why we delete them. Alternatively we temporarily fill the table with incorrect vout values

roeierez: Maybe with -1 ? Not ideal but seems better than dropping payments for all wallets that upgrades.

dangeross: This would currently cause an error listing payments because of the u32 type cast. So we're either [back to making it optional](https://github.com/breez/spark-sdk/pull/949#discussion_r3398883326) or an i32

roeierez (approved):
LGTM

PR #950, opened 2026-06-11
fix: break EventEmitter reference cycles that leaked the SDK object graph per lifecycle

### LINKED ISSUE #947: Server mode: native memory leaks on every SDK lifecycle (build → syncWallet → disconnect → close), eventually OOM
## Context

We've been using the Kotlin bindings Breez SDK. We recently switched from one long-lived client-mode SDK instance to the documented server-mode pattern (shared connection manager + one short-lived SDK instance per operation). The memory leak appeared with that switch. 

We have a reconciliation job that loops over all wallets and calls `syncWallet` on each one. When this is enabled the memory grows steadily until the container hits its memory limit and is OOM-killed. 

We noticed that RSS grows linearly with wallet lifecycle count while JVM heap, threads, and FDs stay flat. This hints that the source is in the Breez SDK and not our kotlin service.

## The leaking lifecycle

```kotlin
// Once at startup, shared across all instances:
val ctx = newSharedSdkContext(SdkContextConfig(network, apiKey, mysqlConfig = mysqlConfig))

// Per operation — leaks every time:
val builder = SdkBuilder(defaultServerConfig(network), Seed.Mnemonic(mnemonic, null))
builder.withSharedContext(ctx)
val sdk = builder.build()
builder.close()
try {
  sdk.syncWallet(SyncWalletRequest)
} finally {
  sdk.disconnect()   // succeeds
  sdk.close()        // succeeds — native memory never returned
}
```

## Possible root cause (AI-assisted code analysis — unverified, please validate)

The following came out of an LLM-assisted read of the 0.15.x source; sharing as a pointer, not
a confirmed diagnosis:

- Every server-mode `build()` runs `ServerRuntime::start_sdk_services`
  (`crates/breez-sdk/core/src/sdk/runtime/server.rs`), which registers
  `ServerRuntimeEventHandler { sdk: sdk.clone() }` into `sdk.event_emitter`.
- `BreezSdk` is `#[derive(Clone)]` over `Arc` fields including `event_emitter: Arc<EventEmitter>`,
  and `EventEmitter` stores handlers in `runtime_event_handlers: RwLock<Vec<Box<dyn RuntimeEventHandler>>>`
  (`events.rs`) with an `add_` method but **no removal API**.
- Result: the `EventEmitter` allocation holds a handler that holds a full `BreezSdk` clone that
  holds an `Arc` back to the same `EventEmitter` — a self-referential Arc cycle. Refcount never
  reaches zero, so the entire per-instance object graph (SparkWallet, storage wrapper, channels,
  …) is retained on every `build()`.
- `disconnect()` doesn't break the cycle: in server mode it is an intentional no-op (no shutdown
  receivers), and it never clears event handlers. The uniffi `close()` just drops one outer Arc.
- The retained `SparkWallet` also pins an `InMemoryTreeStore` whose constructor
  (`crates/spark/src/tree/store.rs`) spawns a processor task that only exits "when `command_tx`
  is dropped" — which never happens once the store is leaked, parking one tokio task plus a
  1024-slot channel per lifecycle on the global runtime.
- Client mode registers the analogous `ClientRuntimeEventHandler { sdk: sdk.clone() }`
  (`runtime/client.rs`), so the cycle likely exists there too — but a long-lived client-mode
  instance leaks once per process, which would explain why this only became visible with the
  per-operation server-mode pattern.

danielgranhao:
Fixes #947

## Problem

Every `build()` registers a `RuntimeEventHandler` holding a full `BreezSdk` clone into the per-instance `EventEmitter`. Since `BreezSdk` owns the emitter via `Arc<EventEmitter>`, the emitter transitively owned itself: a reference cycle that kept the entire per-instance object graph (`SparkWallet`, storage wrapper, channels, the `InMemoryTreeStore` processor task) alive after `disconnect()` and drop. In server mode this leaks once per build/sync/disconnect/close cycle, growing RSS linearly until OOM, exactly as reported.

Three sibling cycles existed in client mode through the same registries:

- `StableBalance` is registered as middleware and held `Arc<EventEmitter>` directly.
- That middleware clone also holds `FlashnetTokenConverter`, which held another `Arc<EventEmitter>`.
- With real-time sync enabled, `SyncedStorage` held `Arc<EventEmitter>` while being reachable from the emitter via the handlers.

## Fix

Break the cycles structurally so plain drop releases the graph (no reliance on `disconnect()` being called):

- `RuntimeEventHandler::handle` now receives `&EventEmitter` from the emitter itself, and the server/client handlers store only the components they use (`storage` / `sync_coordinator` + `spark_wallet` + `storage`) instead of a `BreezSdk` clone.
- `SyncedStorage`, `StableBalance` and `FlashnetTokenConverter` now hold `Weak<EventEmitter>` with upgrade-or-skip at their emit sites (an upgrade only fails during teardown, when there is no listener left to notify).
- Invariant documented on the trait and on `add_middleware`: nothing reachable from the emitter's registries may strongly hold the emitter.

Once the graph drops, the `InMemoryTreeStore` command sender drops with it, so its processor task exits: the per-lifecycle task leak from the issue resolves automatically.

## Evidence

Regression tests (`sdk_builder.rs`) assert via `Weak` pointers that the `EventEmitter` and `SparkWallet` are released after disconnect + drop. They fail on `main` with `SDK object graph leaked after drop (EventEmitter alive: true, SparkWallet alive: true)` and pass with this change, in server mode, client mode, and (ignored, network-dependent) server mode after a real `sync_wallet`.

A manual reproduction harness ([attached as a PR comment](https://github.com/breez/spark-sdk/pull/950#issuecomment-4689423247)) runs the reported lifecycle (shared `SdkContext` + short-lived server-mode SDK per operation) in a loop, sampling RSS and live tokio task count:

| Run | RSS | Tokio tasks |
|---|---|---|
| `main`, 300 lifecycles | linear growth, ~115 KB per lifecycle (27.8 MB at iter 50 to 56.8 MB at iter 300), no plateau | +1 leaked task per lifecycle (304 at iter 300) |
| This branch, 300 lifecycles | flat after warmup (~25.4 MB) | constant (4) |
| This branch, 2000-lifecycle soak | byte-identical 24,432 KB across the final 230+ samples | constant (4) |
| This branch, MySQL shared backend + rotating tenant identities, 300 lifecycles | flat (28.7 MB constant) | constant (5) |
| This branch, 40 lifecycles each running a real `sync_wallet` | oscillates 78 to 91 MB with dips (memory returned), no monotone growth | constant (13) |

The pre-fix numbers match the issue report: RSS grows linearly with wallet lifecycle count while everything else stays flat.

`make check` is green (fmt, clippy cargo + WASM, full workspace tests). All changed types are crate-internal, so no binding interface updates are needed.

### DIFF
--- a/crates/breez-sdk/core/src/events.rs
+++ b/crates/breez-sdk/core/src/events.rs
@@ -201,9 +201,14 @@ pub struct EventEmitter {
     synced_event_buffer: Mutex<Option<InternalSyncedEvent>>,
 }
 
+/// Handlers registered here are owned by the `EventEmitter` for its whole
+/// lifetime, and the emitter is itself owned by the SDK object graph. To keep
+/// that graph droppable, a handler must not hold a strong reference back to
+/// the `EventEmitter` (directly or via a `BreezSdk` clone); the emitter is
+/// passed into `handle` instead.
 #[macros::async_trait]
 pub(crate) trait RuntimeEventHandler: Send + Sync {
-    async fn handle(&self, event: RuntimeEvent);
+    async fn handle(&self, emitter: &EventEmitter, event: RuntimeEvent);
 }
 
 impl EventEmitter {
@@ -273,6 +278,11 @@ impl EventEmitter {
     /// Add middleware to the event processing chain.
     ///
     /// Middleware can transform or suppress events before they reach external listeners.
+    ///
+    /// Middleware is owned by the emitter for its whole lifetime, so it must
+    /// not hold a strong reference back to the emitter (directly or
+    /// transitively); use `Weak<EventEmitter>` instead, or the SDK object
+    /// graph can never be dropped.
     pub async fn add_middleware(&self, middleware: Box<dyn EventMiddleware>) {
         let mut mw = self.middleware.write().await;
         mw.push(middleware);
@@ -286,7 +296,7 @@ impl EventEmitter {
     pub(crate) async fn emit_runtime_event(&self, event: RuntimeEvent) {
         let handlers = self.runtime_event_handlers.read().await;
         for handler in handlers.iter() {
-            handler.handle(event.clone()).await;
+            handler.handle(self, event.clone()).await;
         }
     }
 
--- a/crates/breez-sdk/core/src/realtime_sync/init.rs
+++ b/crates/breez-sdk/core/src/realtime_sync/init.rs
@@ -30,7 +30,7 @@ pub async fn init_and_start_real_time_sync(
     let synced_storage = Arc::new(SyncedStorage::new(
         Arc::clone(&params.storage),
         Arc::clone(&sync_service),
-        params.event_emitter,
+        &params.event_emitter,
         params.lnurl_server_client,
     ));
 
--- a/crates/breez-sdk/core/src/realtime_sync/storage.rs
+++ b/crates/breez-sdk/core/src/realtime_sync/storage.rs
@@ -2,7 +2,7 @@ use std::{
     collections::HashMap,
     fmt::{Display, Formatter},
     str::FromStr,
-    sync::Arc,
+    sync::{Arc, Weak},
 };
 
 use breez_sdk_common::sync::{
@@ -89,7 +89,10 @@ struct ContactSyncData {
 pub struct SyncedStorage {
     inner: Arc<dyn Storage>,
     sync_service: Arc<SyncService>,
-    event_emitter: Arc<EventEmitter>,
+    /// Weak: the emitter's registered handlers and listeners can reach this
+    /// storage, so a strong back-reference would form a reference cycle that
+    /// leaks the whole SDK object graph.
+    event_emitter: Weak<EventEmitter>,
     lnurl_server_client: Option<Arc<dyn LnurlServerClient>>,
 }
 
@@ -121,17 +124,21 @@ impl NewRecordHandler for SyncedStorage {
             return Ok(());
         }
 
-        self.event_emitter
-            .emit_synced(&InternalSyncedEvent {
-                storage_incoming: incoming_count,
-                ..Default::default()
-            })
-            .await;
+        if let Some(event_emitter) = self.event_emitter.upgrade() {
+            event_emitter
+                .emit_synced(&InternalSyncedEvent {
+                    storage_incoming: incoming_count,
+                    ..Default::default()
+                })
+                .await;
+        }
         Ok(())
     }
 
     async fn on_sync_failed(&self) {
-        self.event_emitter.notify_rtsync_failed().await;
+        if let Some(event_emitter) = self.event_emitter.upgrade() {
+            event_emitter.notify_rtsync_failed().await;
+        }
     }
 }
 
@@ -139,13 +146,13 @@ impl SyncedStorage {
     pub fn new(
         inner: Arc<dyn Storage>,
         sync_service: Arc<SyncService>,
-        event_emitter: Arc<EventEmitter>,
+        event_emitter: &Arc<EventEmitter>,
         lnurl_server_client: Option<Arc<dyn LnurlServerClient>>,
     ) -> Self {
         SyncedStorage {
             inner,
             sync_service,
-            event_emitter,
+            event_emitter: Arc::downgrade(event_emitter),
             lnurl_server_client,
         }
     }
@@ -382,7 +389,7 @@ impl SyncedStorage {
 
         let client = Arc::clone(client);
         let inner = Arc::clone(&self.inner);
-        let event_emitter = Arc::clone(&self.event_emitter);
+        let event_emitter = Weak::clone(&self.event_emitter);
         let span = tracing::Span::current();
 
         tokio::spawn(
@@ -436,7 +443,9 @@ impl SyncedStorage {
                     None
                 };
 
-                if old != new {
+                if old != new
+                    && let Some(event_emitter) = event_emitter.upgrade()
+                {
                     event_emitter
                         .emit(&SdkEvent::LightningAddressChanged {
                             lightning_address: new,
@@ -681,7 +690,7 @@ mod tests {
         );
         let sync_service = Arc::new(SyncService::new(sync_storage));
         let event_emitter = Arc::new(EventEmitter::new(true));
-        SyncedStorage::new(storage, sync_service, event_emitter, None)
+        SyncedStorage::new(storage, sync_service, &event_emitter, None)
     }
 
     fn make_incoming_change(
--- a/crates/breez-sdk/core/src/sdk/runtime/client.rs
+++ b/crates/breez-sdk/core/src/sdk/runtime/client.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use platform_utils::time::{Duration, SystemTime};
 use platform_utils::tokio;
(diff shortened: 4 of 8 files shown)


danielgranhao:
The RSS / task-count numbers in the PR description were produced with the manual harness below (originally `crates/breez-sdk/core/tests/leak_repro.rs`, removed from the branch since the checked-in weak-pointer tests are the durable regression guard). To reproduce, save it as `crates/breez-sdk/core/tests/leak_repro.rs` and run:

```sh
cargo test -p breez-sdk-spark --test leak_repro -- --ignored --nocapture
```

`LEAK_REPRO_ITERS` overrides the iteration count, `LEAK_REPRO_SYNC=1` adds a real `sync_wallet` per lifecycle (requires network), and the `_mysql` variant (requires Docker) mirrors a shared MySQL backend with rotating tenant identities.

<details>
<summary><code>leak_repro.rs</code></summary>

```rust
//! Manual reproduction harness for issue #947 (server-mode SDK lifecycle
//! memory leak). Runs the reported lifecycle (shared context + short-lived
//! server-mode SDK per operation) in a loop and samples process RSS and live
//! tokio task count.
//!
//! Run with:
//! ```sh
//! cargo test -p breez-sdk-spark --test leak_repro -- --ignored --nocapture
//! ```
//! `LEAK_REPRO_ITERS` overrides the iteration count. `LEAK_REPRO_SYNC=1`
//! additionally calls `sync_wallet` per lifecycle (requires network).
#![cfg(all(feature = "sqlite", not(target_family = "wasm")))]

use breez_sdk_spark::{
    Network, SdkBuilder, SdkContextConfig, Seed, SyncWalletRequest, default_server_config,
    new_shared_sdk_context,
};

/// Resident set size of this process in KB (macOS/Linux `ps`).
fn rss_kb() -> i64 {
    let out = std::process::Command::new("ps")
        .args(["-o", "rss=", "-p", &std::process::id().to_string()])
        .output()
        .expect("ps should run");
    String::from_utf8_lossy(&out.stdout)
        .trim()
        .parse()
        .expect("rss should parse")
}

fn test_seed() -> Seed {
    Seed::Mnemonic {
        mnemonic: "abandon abandon abandon abandon abandon abandon abandon abandon abandon \
                   abandon abandon about"
            .to_string(),
        passphrase: None,
    }
}

#[tokio::test(flavor = "multi_thread")]
#[ignore = "manual memory-leak reproduction harness; run with --ignored --nocapture"]
async fn server_mode_lifecycle_memory() {
    let iterations: usize = std::env::var("LEAK_REPRO_ITERS")
        .ok()
        .and_then(|v| v.parse().ok())
        .unwrap_or(300);
    let with_sync = std::env::var("LEAK_REPRO_SYNC").is_ok_and(|v| v == "1");

    let context = new_shared_sdk_context(SdkContextConfig::new(Network::Regtest))
        .await
        .expect("shared context");

    let nanos = std::time::SystemTime::now()
        .duration_since(std::time::UNIX_EPOCH)
        .expect("clock")
        .as_nanos();
    let base_dir =
        std::env::temp_dir().join(format!("breez-leak-repro-{}-{nanos}", std::process::id()));

    let metrics = tokio::runtime::Handle::current().metrics();
    let start_rss = rss_kb();
    println!("iteration,rss_kb,alive_tasks");
    println!("0,{start_rss},{}", metrics.num_alive_tasks());

    for i in 1..=iterations {
        let sdk = SdkBuilder::new(default_server_config(Network::Regtest), test_seed())
            .with_shared_context(context.clone())
            .with_default_storage(
                base_dir
                    .join(format!("wallet-{i}"))
                    .to_string_lossy()
                    .into_owned(),
            )
            .build()
            .await
            .expect("server-mode build");

        if with_sync {
            let _ = sdk.sync_wallet(SyncWalletRequest {}).await;
        }

        sdk.disconnect().await.expect("disconnect");
        drop(sdk);

        if i % 10 == 0 {
            println!("{i},{},{}", rss_kb(), metrics.num_alive_tasks());
        }
    }

    // Let detached per-instance tasks (if any) observe channel closure.
    tokio::time::sleep(std::time::Duration::from_secs(2)).await;
    let end_rss = rss_kb();
    let growth = end_rss - start_rss;
    println!("final,{end_rss},{}", metrics.num_alive_tasks());
    println!(
        "RSS growth over {iterations} lifecycles: {growth} KB \
         (~{} KB per lifecycle)",
        growth / i64::try_from(iterations).expect("iteration count fits i64")
    );

    std::fs::remove_dir_all(&base_dir).ok();
}

/// Same lifecycle as `server_mode_lifecycle_memory`, but with the storage
/// backend on the shared context (`MySQL` via testcontainers) and the tenant
/// identity cycling across builds, mirroring a multi-wallet reconciliation
/// job against one shared database.
#[cfg(feature = "mysql")]
#[tokio::test(flavor = "multi_thread")]
#[ignore = "manual memory-leak reproduction harness; requires Docker"]
async fn server_mode_lifecycle_memory_mysql() {
    use breez_sdk_spark::{KeySetConfig, KeySetType, MysqlStorageConfig, mysql_storage};
    use testcontainers::runners::AsyncRunner;
    use testcontainers_modules::mysql::Mysql;

    let iterations: usize = std::env::var("LEAK_REPRO_ITERS")
        .ok()
        .and_then(|v| v.parse().ok())
        .unwrap_or(300);

    let container = Mysql::default().start().await.expect("mysql container");
    let host_port = container
        .get_host_port_ipv4(3306)
        .await
        .expect("mysql host port");
    let connection_string = format!("mysql://root@127.0.0.1:{host_port}/test");
    let backend =
        mysql_storage(MysqlStorageConfig::with_defaults(connection_string)).expect("mysql backend");

    let context = new_shared_sdk_context(SdkContextConfig {
        storage: Some(backend),
        ..SdkContextConfig::new(Network::Regtest)
    })
    .await
    .expect("shared context");

    let metrics = tokio::runtime::Handle::current().metrics();
    let start_rss = rss_kb();
    println!("iteration,rss_kb,alive_tasks");
    println!("0,{start_rss},{}", metrics.num_alive_tasks());

    for i in 1..=iterations {
        let account = u32::try_from(i % 10).expect("small account number");
        let sdk = SdkBuilder::new(default_server_config(Network::Regtest), test_seed())
            .with_key_set(KeySetConfig {
                key_set_type: KeySetType::Default,
                use_address_index: false,
                account_number: Some(account),
            })
            .with_shared_context(context.clone())
            .build()
            .await
            .expect("server-mode build");

        sdk.disconnect().await.expect("disconnect");
        drop(sdk);

        if i % 10 == 0 {
            println!("{i},{},{}", rss_kb(), metrics.num_alive_tasks());
        }
    }

    tokio::time::sleep(std::time::Duration::from_secs(2)).await;
    let end_rss = rss_kb();
    let growth = end_rss - start_rss;
    println!("final,{end_rss},{}", metrics.num_alive_tasks());
    println!(
        "RSS growth over {iterations} lifecycles: {growth} KB \
         (~{} KB per lifecycle)",
        growth / i64::try_from(iterations).expect("iteration count fits i64")
    );
}
```

</details>

dangeross (approved):
LGTM

Not for the PR, I think an external EventListener holding the SDK will still prevent a complete teardown, keeping it alive until either the listener is removed or the SDK is explicitly released. Relying on GC alone won't reclaim it (Flutter exempt). Maybe worth tuning the docs

JssDWt (approved):

PR #951, opened 2026-06-12
fix: unregister event listeners on disconnect

danielgranhao:
Follow-up to the listener-lifecycle discussion on #950 (not a blocker for it; this is independent and based on main).

## Problem

A registered `EventListener` is held by the SDK until removed. A listener that references the SDK instance therefore pins the whole instance: the SDK holds the listener, and the listener holds the SDK back. In the bindings, garbage collection alone can never reclaim this, because the foreign listener object is rooted on the Rust side for callback dispatch. Teardown required either `remove_event_listener` for every listener or an explicit `close()`/`destroy()`.

## Fix

`disconnect()` now unregisters all external event listeners (before the server-mode early return, which previously skipped any cleanup). After `disconnect`, the SDK holds no references to application objects, so listeners that capture the SDK no longer keep it alive and the instance is always releasable.

Docs updated on `add_event_listener` (listener is held until removed or until `disconnect`) and `disconnect` (also unregisters listeners).

## Tests

- `events::tests::test_clear_external_listeners`: cleared listeners receive no further events and are no longer individually removable.
- `sdk_builder::tests::disconnect_unregisters_event_listeners`: end-to-end via the public API; after `disconnect`, `remove_event_listener` reports the listener already gone.

### DIFF
--- a/crates/breez-sdk/core/src/sdk/api.rs
+++ b/crates/breez-sdk/core/src/sdk/api.rs
@@ -27,6 +27,11 @@ use super::{BreezSdk, helpers::get_deposit_address, parse_input};
 impl BreezSdk {
     /// Registers a listener to receive SDK events
     ///
+    /// The SDK holds the listener until it is removed with
+    /// `remove_event_listener` or until `disconnect` unregisters all
+    /// listeners. A held listener that references the SDK instance keeps
+    /// that instance alive.
+    ///
     /// # Arguments
     ///
     /// * `listener` - An implementation of the `EventListener` trait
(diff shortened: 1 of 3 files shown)


crates/breez-sdk/core/src/sdk/api.rs
```diff
@@ -56,11 +61,16 @@ impl BreezSdk {
     /// This method stops the background tasks started by the `start()` method.
     /// It should be called before your application terminates to ensure proper cleanup.
     ///
+    /// It also unregisters all event listeners, so after this call the SDK
+    /// holds no references to application objects and listeners that
```
dangeross: nit: It might hold other application implemented traits

danielgranhao: Right, fixed

dangeross (approved):
LGTM

JssDWt (approved):

PR #952, opened 2026-06-12
refactor: make EventEmitter references top-down, remove Weak back-references

danielgranhao:
Follow-up to #950 (fixes the structural side of #947). That fix broke the client-mode reference cycle by switching emitter-reachable components to `Weak<EventEmitter>`. `Weak` works but is a convention: any future emitter-reachable struct holding a strong `Arc<EventEmitter>` recreates the leak, caught only by the regression tests. This PR removes the stored back-references entirely.

## Changes

- **`SyncedStorage`** is now a pure `Storage` wrapper (`inner` + `sync_service`, no emitter). It is the emitter-reachable piece (registered handlers hold `sdk.storage`). The record-application and emitting half moved to a new `SyncedRecordHandler`, owned by the `SyncProcessor`. The processor is never emitter-reachable and its tasks stop on shutdown, so the handler holds the emitter strongly and releases it on `disconnect()`.
- **`StableBalance`** split into:
  - `StableBalanceCore`: shared state and queue/limits logic, no emitter reference,
  - `StableBalanceMiddleware`: emitter-owned, holds only the core, never emits,
  - `StableBalance`: worker and API surface, holds the emitter strongly. Owned by `BreezSdk` and the conversion worker task (stops on shutdown), never by the emitter.

  `set_active_token` cache invalidation crosses the split through the shared core.
- **`FlashnetTokenConverter`** stores no emitter at all anymore: `TokenConverter::convert()` takes the emitter as a parameter, the same pattern as `RuntimeEventHandler::handle`. `convert()` was the only emitting method (payment events for the swap legs), and all callers (stable balance worker, send-with-conversion paths) already own the emitter. This makes the middleware -> core -> converter chain cycle-free by construction instead of relocating the stored reference.

No `Weak<EventEmitter>` remains in the crate. The invariant is now: emitter-reachable code either holds no emitter reference, or receives one per call from owners outside the emitter graph.

## Tests

- New regression test: client-mode build with real-time sync and stable balance enabled (offline, lazy gRPC connect) must release the whole object graph after `disconnect()`. This covers the path the existing client-mode test skipped (`real_time_sync_server_url = None`).
- All existing #947/#950/#951 regression tests kept and passing.
- `fmt`, `clippy` (cargo + WASM), and the full core test suite pass.

No public API changes; everything touched is crate-internal.

### DIFF
--- a/crates/breez-sdk/core/src/events.rs
+++ b/crates/breez-sdk/core/src/events.rs
@@ -290,9 +290,10 @@ impl EventEmitter {
     /// Middleware can transform or suppress events before they reach external listeners.
     ///
     /// Middleware is owned by the emitter for its whole lifetime, so it must
-    /// not hold a strong reference back to the emitter (directly or
-    /// transitively); use `Weak<EventEmitter>` instead, or the SDK object
-    /// graph can never be dropped.
+    /// not hold a reference back to the emitter (directly or transitively),
+    /// or the SDK object graph can never be dropped. Code that needs to emit
+    /// must receive the emitter from its caller instead (see
+    /// `RuntimeEventHandler` and `TokenConverter::convert`).
     pub async fn add_middleware(&self, middleware: Box<dyn EventMiddleware>) {
         let mut mw = self.middleware.write().await;
         mw.push(middleware);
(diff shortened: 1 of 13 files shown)


JssDWt (approved):
LGTM

dangeross (approved):
LGTM

PR #954, opened 2026-06-12
fix(spark-wallet): retry transfers when operators report a leaf not yet available

JssDWt:
## Problem

A transfer attempted right after a deposit claim can be rejected by the operators:

```
status: FailedPrecondition, message: "consensus send transfer failed: prepare failed:
... unable to validate leaf ...: leaf ... is not available to transfer, status: CREATING"
```

The local store already counts the leaf as available (its deposit event flipped the status, so `getInfo`/balance includes it), but the operators' transfer-prepare validation still sees the leaf as `CREATING`. There is a short eventual-consistency window after a deposit confirmation where the SDK's view is ahead of the operators'.

This error class was not retried, so the transfer failed outright. It surfaced as a flaky failure of the `test_non_static_deposit_with_faucet` integration test, but it affects any user transferring straight after a deposit confirms.

## Fix

- Detect the condition via the **structured `google.rpc.ErrorInfo` reason** the operators attach to the gRPC status (`LEAF_UNAVAILABLE`), not the human-readable message. The operators consistently set machine-readable reasons (`LEAF_UNAVAILABLE`, `INVALID_STATE`, `DUPLICATE_OPERATION`), so this is a stable contract rather than a brittle string match.
- Back off and retry in `with_leafs_spent_retry!` (200ms → 3s exponential, up to 5 attempts) until the operators converge. This covers every operation routed through the macro (transfers, lightning sends, etc.).
- The retry deliberately **skips `refresh_leaves`**: that call keeps only `Available` leaves, and if the operators' query path also lagged it would drop the leaf and turn this into a spurious insufficient-funds error. The failed attempt releases its reservation via `with_reserved_leaves`, so the next attempt re-selects the same leaf cleanly once the operators catch up.

`operator_error_reason()` is a reusable helper — the structured path the existing `is_transfer_locked_error` TODO ("we want to move to an error code") was waiting for.

## Notes

- Adds the `tonic-types` dependency (0.12.3, matching tonic 0.12.3). It is pure prost type-decoding with `tonic` default-features off (no transport); confirmed WASM-safe via `make clippy-check`.
- The integration test needs the deployed regtest operators, so its timing is not reproducible locally, but the retry now keys on the exact structured reason that failed it. The integration test itself is left untouched.

## Testing

- `make fmt-check`, `make clippy-check` (cargo + WASM) ✓
- `cargo test -p spark-wallet` — 12 passed, including new unit tests pinning the classifier to round-tripped `ErrorInfo` details (reason matches / different reason does not / detail-less status does not).

### DIFF
--- a/crates/spark-wallet/src/wallet.rs
+++ b/crates/spark-wallet/src/wallet.rs
@@ -63,14 +63,12 @@ use crate::{
 
 const SELECT_LEAVES_MAX_RETRIES: usize = 3;
 const MAX_LEAF_SPENT_RETRIES: usize = 3;
-const MAX_RATE_LIMIT_RETRIES: u32 = 5;
-const RATE_LIMIT_BASE_DELAY_MS: u64 = 100;
-const RATE_LIMIT_MAX_DELAY_MS: u64 = 3000;
-/// Retries for a leaf the operators have not finished finalizing yet. The window
-/// after a deposit claim is short, so a few hundred ms of backoff clears it.
-const MAX_LEAF_UNAVAILABLE_RETRIES: u32 = 5;
-const LEAF_UNAVAILABLE_BASE_DELAY_MS: u64 = 100;
-const LEAF_UNAVAILABLE_MAX_DELAY_MS: u64 = 3000;
+/// Backoff for transient operator errors that clear on their own: rate limiting
+/// (ResourceExhausted) and leaves the operators have not finished finalizing yet
+/// (LEAF_UNAVAILABLE).
+const MAX_BACKOFF_RETRIES: u32 = 5;
+const BACKOFF_BASE_DELAY_MS: u64 = 100;
+const BACKOFF_MAX_DELAY_MS: u64 = 3000;
 /// Grace period before claiming orphaned counter-swap transfers.
 /// Generous to absorb clock skew between the server (`created_time`) and the local clock.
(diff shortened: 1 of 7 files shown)


crates/spark-wallet/src/wallet.rs
```diff
@@ -150,6 +182,24 @@ macro_rules! with_leafs_spent_retry {
                     tokio::time::sleep(std::time::Duration::from_millis(delay_ms)).await;
                     continue;
                 }
+                Err(ServiceError::ServiceConnectionError(e))
```
roeierez: I wonder if we really need a separate branch (and constants) for each kind of error or we can use on a single retry counter, delay, etc...

roeierez (changes requested):

crates/spark-wallet/src/wallet.rs
JssDWt: Agreed. Both arms did identical work, so I collapsed them into a single backoff branch in 508502a3c: one `backoff_attempt` counter, one set of constants (`MAX_BACKOFF_RETRIES` / `BACKOFF_BASE_DELAY_MS` / `BACKOFF_MAX_DELAY_MS`), and a combined `is_backoff_retryable_error` predicate composing the rate-limit and leaf-unavailable classifiers.

roeierez (approved):
LGTM

danielgranhao (approved):
LGTM

dangeross (approved):
LGTM

PR #955, opened 2026-06-14
Cross-chain: Change to fees excluded amount mode

dangeross:
Builds on #799. Cross-chain sends now deliver the amount the user specified; fees come on top instead of being deducted. "Send all" is unchanged.

- **Boltz** uses its native exact-out API to commit to a delivery amount.
- **Orchestra** (no exact-out support yet) inflates the source amount client-side to hit the same target.
- **Conversion sends** (e.g. USDB → USDC via a sats-source route) chain the AMM and the provider so the recipient still lands the full amount.
- **"Send all" with conversion** now also surfaces the wallet-side token debit on the prepare response, matching the "fees on top" behaviour. The underlying send is unchanged.
- **Idempotency gate widened**: `idempotency_key` is now rejected on any send with a non-idempotent leg (direct token send or AMM conversion). Previously, conversion sends silently accepted the key but only protected half the flow.


crates/breez-sdk/core/src/cross_chain/boltz.rs
```diff
@@ -184,49 +235,62 @@ impl BoltzService {
...
         ))
     }
 
-    async fn create_swap(
+    /// Boltz exact-out reverse swap: returns a `PreparedSwap` whose
+    /// `invoice_amount_sats` is the inflated source required to deliver
+    /// `output_amount` after every fee layer (spread + miner + CCTP/OFT
+    /// bridge). `create_reverse_swap` then commits an HD key index and
+    /// persists swap state; the only clean exit after that is a timeout.
+    async fn create_swap_exact_out(
```
danielgranhao: The delivered amount can still be below target output amount due to slippage right? Asking because "exact out" might be a bit misleading

crates/breez-sdk/core/src/cross_chain/mod.rs
```diff
@@ -199,15 +244,22 @@ pub enum CrossChainProviderContext {
 #[derive(Debug, Clone)]
 pub(crate) struct CrossChainPrepared {
     pub amount_in: u128,
-    /// Amount the recipient will receive, already net of any destination-chain
-    /// costs (gas, bridge messaging). Destination-side costs are implicit in
-    /// `amount_in - estimated_out` and are not re-counted in `fee_amount`.
+    /// Amount in expressed in the cross-chain asset's base units, using
```
danielgranhao:
```suggestion
    /// Amount expressed in the cross-chain asset's base units, using
```

crates/breez-sdk/core/src/cross_chain/mod.rs
```diff
@@ -199,15 +244,22 @@ pub enum CrossChainProviderContext {
...
+    /// the fiat rate or decimal rescale the SDK used at prepare time. For a
+    /// send, this is the destination side; for a receive, the source side.
+    pub asset_amount_in: u128,
```
danielgranhao: If the rustdoc is correct and this is for the destination side of sends, the name is a bit surprising. By "amount in" I would think this is always sender side?

danielgranhao (commented):
Looks good!

danielgranhao (approved):

crates/breez-sdk/core/src/sdk/payments/prepare/cross_chain.rs
```diff
@@ -232,65 +275,453 @@ pub(crate) async fn prepare(
...
+    // Non-USD pairs downgrade to FeesIncluded: the fiat-inversion and
+    // par-rescale below assume USD pegs on both legs. Caller observes the
+    // downgrade via `response.fee_policy`.
+    if !is_usd_stable_pair(&src_meta.ticker, &route.asset) {
+        tracing::warn!(
+            src_ticker = %src_meta.ticker,
+            dest_asset = %route.asset,
+            "Non-USD-stable pair; downgrading FeesExcluded to FeesIncluded (recipient receives amount − fees)"
+        );
+        return prepare_token_denominated_fees_included(
```
roeierez: It seems that this function expects the amount in the "source"  denominated while we are in the fee excluded flow that consider the amount as "destination" denominated?

roeierez (changes requested):

crates/breez-sdk/core/src/sdk/payments/prepare/cross_chain.rs
dangeross: I clarified the fn docs. The amount is passed through directly from the request in both fns. Regarding "downgrading" to FeesIncluded, the response gives this information, but another option might be to reject FeesExcluded for non-USD-pegged destinations and force them to explicitly set FeesIncluded. Wdyt?

roeierez (approved):
LGTM

PR #956, opened 2026-06-18
Encrypt static deposit key share in claim_static_deposit

danielgranhao:
## What

`claim_static_deposit` previously sent the user's 32-byte static deposit secret key to the SSP in cleartext (hex) over GraphQL. This switches to the SSP's new backward-compatible encrypted flow: the secret is ECIES-encrypted to the SSP identity public key for the selected network (the same scheme used for transfer leaf secret ciphers) and sent in the new `encrypted_deposit_secret_key` field, with the raw `deposit_secret_key` sent as null.

## Changes

- `crates/spark/schema/spark.graphql`: relax `deposit_secret_key` to nullable and add `encrypted_deposit_secret_key` on `ClaimStaticDepositInput`. The `graphql_client` macro regenerates both as `Option<String>`.
- `crates/spark/src/services/deposit.rs`: encrypt the exported deposit secret to `ssp_client.identity_public_key()` via `utils::ecies::encrypt`, hex-encode, and send it in the encrypted field; send `deposit_secret_key: None`.

Only `claim_static_deposit` is used by the SDK, so the other static-deposit mutations are untouched.

**Verified to be working on mainnet**

### DIFF
--- a/crates/spark/src/services/deposit.rs
+++ b/crates/spark/src/services/deposit.rs
@@ -300,6 +300,20 @@ impl DepositService {
             })
             .await?;
 
+        // The SSP co-signs the claim and so needs the static-deposit secret.
+        // Send it ECIES-encrypted to the SSP identity public key (same scheme as
+        // transfer leaf secret ciphers) instead of in cleartext over GraphQL.
+        let encrypted_deposit_secret_key = utils::ecies::encrypt(
+            &self
+                .ssp_client
+                .identity_public_key()
+                .serialize_uncompressed(),
+            &deposit_secret_key.secret_bytes(),
+        )
+        .map_err(|e| {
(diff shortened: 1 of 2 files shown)


crates/spark/src/services/deposit.rs
dangeross: Can be updated "needs it in the clear"

danielgranhao: Good catch. Updated in 981580d08: dropped "needs it in the clear" (no longer accurate now that we encrypt before sending), and moved the SSP co-sign rationale down to the encryption comment to avoid duplicating it.

dangeross (approved):
LGTM

roeierez (approved):

PR #958, opened 2026-06-18
Accept BIP21 URIs in prepare_send_payment

erdemyerebasmaz:
## Background

`parse()` already handles BIP21 thoroughly — it returns `InputType::Bip21` with a structured `Bip21Details` (`amount_sat`, `label`, `message`, and a `payment_methods` list covering on-chain, BOLT11, BOLT12, spark, silent payments). The gap is downstream: `prepare_send_payment` parsed the input but only dispatched the **concrete** methods (`SparkAddress`, `SparkInvoice`, `Bolt11Invoice`, `BitcoinAddress`). A `bitcoin:<addr>?amount=…` URI parsed to `InputType::Bip21` and fell straight through to `Err("Unsupported payment method")`.

That forced every consumer to unwrap the URI themselves before calling prepare — e.g. glow-web just shipped exactly that workaround ([glow-web: "Handle BIP21 bitcoin: URIs in the send flow"](https://github.com/breez/glow-web)).

## Change

Add an `InputType::Bip21` arm to `prepare()` that:
1. Selects the preferred payable method from `payment_methods`, and
2. For a bare address (no embedded amount), falls back to the URI's `amount_sat` when the caller didn't pass an `amount`. Invoices keep their own embedded amount.

Every other input is unchanged.

## Design decision to confirm — payment-method preference

When a unified URI carries more than one method (e.g. on-chain `+ lightning=` `+ spark=`), `select_bip21_payment_method` picks in **cost/speed order: Spark > Lightning > on-chain** (`SparkInvoice` > `SparkAddress` > `Bolt11Invoice` > `BitcoinAddress`). Methods the send flow can't action (BOLT12 offers, silent payments) are skipped. This matches the rails' fee/latency profile for a Spark wallet, but it's a routing decision — happy to flip the order or restrict to a single method if the team prefers callers choose explicitly.

## Verification

- `cargo test -p breez-sdk-spark --lib selection_tests` — 5/5 (preference order + fallback + empty) ✅
- `cargo clippy -p breez-sdk-spark --lib --tests -- -D warnings` ✅
- `cargo fmt -- --check` ✅

End-to-end prepare behavior (amount injection per method) is covered by the unit-tested selection logic; a wallet-level integration test would need a live SDK instance, so it's not included here.

### DIFF
--- a/crates/breez-sdk/core/src/sdk/payments/prepare/mod.rs
+++ b/crates/breez-sdk/core/src/sdk/payments/prepare/mod.rs
@@ -16,10 +16,35 @@ pub(super) async fn prepare(
 ) -> Result<PrepareSendPaymentResponse, SdkError> {
     let parsed_input = sdk.parse(&request.payment_request).await?;
 
+    // A BIP21 URI wraps one or more concrete payment methods. Resolve it to the
+    // preferred method so callers can pay a `bitcoin:` URI directly. A bare
+    // address carries no amount, so fall back to the URI's `amount` when the
+    // caller didn't pass one (an invoice carries its own amount).
+    let (payment_method, request) = match &parsed_input {
+        InputType::Bip21(details) => {
+            let method =
+                select_bip21_payment_method(&details.payment_methods).ok_or_else(|| {
+                    SdkError::InvalidInput("BIP21 URI has no supported payment method".to_string())
+                })?;
+            let mut request = request;
+            if request.amount.is_none()
+                && matches!(
+                    method,
+                    InputType::BitcoinAddress(_) | InputType::SparkAddress(_)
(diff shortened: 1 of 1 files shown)


dangeross (commented):
I feel this is more of an explicit user/application decision to pick one (if multiple) of the child input types in the BIP21 parse response. The choice also influences what parameters are passed with it

erdemyerebasmaz:
> I feel this is more of an explicit user/application decision to pick one (if multiple) of the child input types in the BIP21 parse response. The choice also influences what parameters are passed with it

Sorry, this was not meant to be a PR but an investigation. I agree that the decision should be made by the user/application. Closing.

PR #959, opened 2026-06-18
feat: add ca_cert_pem to SparkSigningOperator

peterrojs:
adds an optional ca_cert_pem field to SparkSigningOperator so callers can verify a signing operator's tls certificate against a custom pem-encoded CA instead of the default webpki roots. this lets the sdk connect to local signing operators that present self-signed development certificates.

the field threads through sdk_builder into create_operator_config, and default_spark_config surfaces any ca_cert already present on the operator config so the round trip is lossless. mirrored in the wasm and flutter models and added to the config doc snippets for rust, python, go, csharp, kotlin and dart. swift, typescript, react-native and nodejs snippets have no SparkSigningOperator construction so they are unaffected.

### DIFF
--- a/crates/breez-sdk/core/src/models/mod.rs
+++ b/crates/breez-sdk/core/src/models/mod.rs
@@ -822,6 +822,11 @@ pub struct SparkSigningOperator {
     pub address: String,
     /// Hex-encoded compressed public key of the operator.
     pub identity_public_key: String,
+    /// Optional PEM-encoded CA certificate for TLS verification.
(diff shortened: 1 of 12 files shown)


dangeross (commented):
Looks good, I think just the config docs snippets need updating for react-native

peterrojs:
> Looks good, I think just the config docs snippets need updating for react-native

done!

dangeross (approved):
LGTM

PR #960, opened 2026-06-20
Boltz cross-chain: sync swap rows for cross-instance recovery

danielgranhao:
## What

A Boltz cross-chain swap is known only to its creator instance; if the creator goes offline, other instances show it stuck `Pending` forever. This bumps `boltz-client` to `809ac77` (seedless mode) and persists + syncs full swap rows so any instance of the same wallet can drive a swap to terminal. boltz-client owns correctness and multi-instance convergence; the SDK just persists and syncs rows with plain last-writer-wins.

## How

- **Seedless boltz-client**: construct via `new_seedless`; drop the random per-device seed, the encrypted seed handle, the instance id, and the key-index counter.
- **Encrypted-at-rest, signer-free storage**: new `StoredCrossChainSwap { provider, id, is_terminal, updated_at, data, secrets }`, shared across cross-chain providers and discriminated by `provider`. The Boltz adapter tags rows `provider = "boltz"`, lifts `key_source` out of the swap JSON, ECIES-encrypts it under a fixed identity-derived path (deterministic per mnemonic+network, so a second instance can decrypt), and persists only the ciphertext in `secrets`. The storage layer needs no signer and treats `data` and `secrets` as opaque. The encrypt/decrypt boundary lives in the adapter.
- **Storage + migrations**: `Storage::{set_cross_chain_swap, get_cross_chain_swap(provider, id), list_active_cross_chain_swaps(provider)}` and a `cross_chain_swaps` table (composite PK `(provider, id)`) across all 7 backends (Rust SQLite/Postgres/MySQL, JS Web/Node-SQLite/Node-Postgres/Node-MySQL); `brz_` prefix on Postgres/MySQL.
- **Realtime sync**: `RecordType::CrossChainSwap` with `data_id = "{provider}:{id}"` — `SyncedStorage::set_cross_chain_swap` emits then delegates; `SyncedRecordHandler::handle_cross_chain_swap_change` applies via the inner store (structural loop guard). Plain last-writer-wins.
- **Convergence**: the always-on delivery tick is the pickup path for synced-in swaps (resume is only a startup accelerator); `conversion_info` is derived locally from the synced swap row.

## Decisions

- **Encrypt at rest** (in the adapter, not the storage layer) rather than plaintext — money-critical secrets get at-rest protection while storage stays signer-free.
- **Provider-keyed table over per-provider tables** — Orchestra has the same row shape (quote/order id, opaque JSON, optional secrets) and the same pre-payment lifetime need on the receive path. Landing the rename now (table is unreleased) means future providers add only an adapter, no storage / trait / sync changes.
- **v1 upgrade = clean break**: seedless `RandomSecretProvider` rejects derived swaps, so in-flight v1 swaps can't run under the new service (funds are safe via the LN hold-invoice timeout refund). Legacy KV keys are left orphaned.

## Tests

- Shared `cross_chain_swaps` CRUD wired into all 7 storage harnesses, exercising composite-PK isolation across providers and provider-scoped `list_active`.
- Adapter encrypt/lift round-trip (secrets encrypted, absent from plaintext `data`, terminal swaps excluded from the active list).
- Plumbing sync round-trip for the `CrossChainSwap` record (serialize → outgoing queue → incoming-apply → second-instance store), replaying the actually-emitted fields under both `"boltz"` and `"orchestra"` provider tags so any hard-coded provider in emit/apply is caught.
- Out of scope (no cross-chain test infra): end-to-end recovery scenarios — that convergence logic lives in and is unit-tested by boltz-client.

### DIFF
--- a/crates/breez-sdk/core/src/cross_chain/boltz_storage_adapter.rs
+++ b/crates/breez-sdk/core/src/cross_chain/boltz_storage_adapter.rs
@@ -25,6 +25,7 @@ use std::sync::Arc;
 use base64::{Engine, engine::general_purpose::STANDARD as BASE64};
 use bitcoin::bip32::DerivationPath;
 use boltz_client::{BoltzError, BoltzStorage, models::BoltzSwap};
+use tracing::warn;
 
 use crate::{Storage, persist::StoredBoltzSwap, signer::BreezSigner};
 
@@ -144,7 +145,12 @@ impl BoltzStorage for BoltzStorageAdapter {
             .map_err(|e| BoltzError::Store(format!("Failed to list active swaps: {e}")))?;
         let mut swaps = Vec::with_capacity(rows.len());
         for stored in rows {
-            swaps.push(self.swap_from_stored(stored).await?);
+            let id = stored.id.clone();
+            // Skip row on error (bad base64/json, decryption).
+            match self.swap_from_stored(stored).await {
+                Ok(swap) => swaps.push(swap),
+                Err(e) => warn!("Skipping Boltz swap '{id}': failed to load from storage: {e}"),
+            }
         }
         Ok(swaps)
     }
--- a/crates/breez-sdk/core/src/cross_chain/boltz_storage_adapter.rs
+++ b/crates/breez-sdk/core/src/cross_chain/boltz_storage_adapter.rs
@@ -13,16 +13,16 @@
 //! decrypted and `key_source` spliced back, reconstructing the full
 //! [`BoltzSwap`].
 //!
-//! The encryption key is derived from the wallet identity at a fixed path, so it
-//! is deterministic across every instance restored from the same mnemonic. That
-//! is what lets a second instance decrypt a swap synced to it and drive it to
-//! terminal (see the realtime-sync `CrossChainSwap` record). Correctness and
-//! cross-instance convergence are boltz-client's job: the SDK just persists and
-//! syncs rows with plain last-writer-wins.
+//! The encryption key is derived from the wallet identity at a fixed path (see
+//! [`BOLTZ_SECRETS_ENCRYPTION_PATH`]), so it is deterministic across every
+//! instance restored from the same mnemonic. That is what lets a second instance
+//! decrypt a swap synced to it and drive it to terminal (see the realtime-sync
+//! `CrossChainSwap` record). Correctness and cross-instance convergence are
+//! boltz-client's job: the SDK just persists and syncs rows with plain
+//! last-writer-wins.
 //!
 //! Terminal swap rows are retained (only excluded from `list_active_swaps`),
-//! keeping a completed swap queryable by id for the wallet's lifetime. Growth is
-//! bounded by swap volume; no TTL in v1.
+//! keeping a completed swap queryable by id for the wallet's lifetime.
 
 use std::sync::Arc;
 
--- a/crates/breez-sdk/core/src/cross_chain/boltz_storage_adapter.rs
+++ b/crates/breez-sdk/core/src/cross_chain/boltz_storage_adapter.rs
@@ -1,173 +1,152 @@
 //! Adapter exposing the SDK [`Storage`] as a [`boltz_client::BoltzStorage`].
 //!
-//! Canonical Boltz swap state lives in a dedicated cache KV namespace
-//! rather than on payment metadata rows. The adapter is payment-row
-//! agnostic: it can run at prepare time, before any [`crate::Payment`]
-//! has been written.
+//! boltz-client runs in seedless mode: each swap carries its own random preimage
+//! and gas/claim key under `key_source` ([`boltz_client::models::SwapKeySource::Stored`]).
+//! Those secrets are money-critical, so before a swap row reaches local storage
+//! this adapter lifts `key_source` out of the swap JSON, ECIES-encrypts it via
+//! the SDK signer, and persists only the ciphertext in
+//! [`StoredBoltzSwap::secrets`]. The rest of the swap is stored as plaintext
+//! JSON in [`StoredBoltzSwap::data`]. On read the ciphertext is decrypted and
+//! `key_source` spliced back, reconstructing the full [`BoltzSwap`].
 //!
-//! # Retention
+//! The encryption key is derived from the wallet identity at a fixed path, so it
+//! is deterministic across every instance restored from the same mnemonic. That
+//! is what lets a second instance decrypt a swap synced to it and drive it to
+//! terminal (see the realtime-sync `BoltzSwap` record). Correctness and
+//! cross-instance convergence are boltz-client's job: the SDK just persists and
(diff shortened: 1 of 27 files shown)


crates/breez-sdk/core/src/cross_chain/boltz_storage_adapter.rs
```diff
@@ -1,173 +1,152 @@
...
-        self.storage
-            .set_cached_item(key, next.to_string())
+            .list_active_boltz_swaps()
             .await
-            .map_err(|e| BoltzError::Store(format!("Failed to persist key index: {e}")))?;
-        Ok(current)
+            .map_err(|e| BoltzError::Store(format!("Failed to list active swaps: {e}")))?;
+        let mut swaps = Vec::with_capacity(rows.len());
+        for stored in rows {
+            swaps.push(self.swap_from_stored(stored).await?);
```
dangeross: One row that fails to decrypt here aborts the whole list, so a single corrupt `secrets` blob stalls convergence for every active swap. `warn!` and skip the bad row instead?

crates/breez-sdk/core/src/persist/mod.rs
```diff
@@ -340,6 +340,39 @@ where
...
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[serde(rename_all = "camelCase")]
+pub struct StoredBoltzSwap {
+    /// Boltz swap id (primary key).
+    pub id: String,
+    /// `swap.status.is_terminal()`, lifted into an indexed column so
+    /// `list_active_boltz_swaps` filters without parsing `data`.
+    pub is_terminal: bool,
+    /// `swap.updated_at`, used for ordering and last-writer-wins.
+    pub updated_at: u64,
```
dangeross: `updated_at` is documented as "used for ordering and last-writer-wins", but nothing reads it: conflict resolution is the sync framework's revision-based merge, `list_active_boltz_swaps` has no `ORDER BY`, and the apply path never compares it. It also duplicates the `updated_at` already inside `data`. Reword the doc, or drop the column?

crates/breez-sdk/wasm/js/web-storage/index.js
```diff
@@ -575,6 +575,19 @@ class MigrationManager {
           }
         },
       },
+      {
+        // Boltz cross-chain swap rows, synced for cross-instance recovery.
+        // `data` is the BoltzSwap JSON with `key_source` lifted out; the lifted
+        // secrets are ECIES-encrypted (base64) into `secrets` by the adapter.
+        // No index on isTerminal: IndexedDB can't key on a boolean, so
+        // listActiveBoltzSwaps scans and filters (swap volume is low).
```
dangeross: Can be removed
```suggestion
        // listActiveBoltzSwaps scans and filters.
```

crates/breez-sdk/wasm/src/persist/mod.rs
```diff
@@ -485,6 +535,15 @@ impl breez_sdk_spark::Storage for WasmStorage {
...
 }
 
+#[wasm_bindgen(typescript_custom_section)]
+const STORED_BOLTZ_SWAP_INTERFACE: &'static str = r#"export interface StoredBoltzSwap {
+    id: string;
+    isTerminal: boolean;
+    updatedAt: number;
+    data: string;
+    secrets: string;
+}"#;
```
dangeross: Any reason not to mirror this with `macros::extern_wasm_bindgen` like the other storage types?

crates/breez-sdk/core/src/realtime_sync/storage.rs
```diff
@@ -404,6 +416,24 @@ impl SyncedRecordHandler {
         Ok(())
     }
 
+    /// Applies a synced-in cross-chain swap row via the inner store (bypassing
+    /// the outgoing-emit wrapper). Plain last-writer-wins: each provider's
+    /// convergence logic, not the SDK, owns correctness. `data_id` is
```
JssDWt: > Plain last-writer-wins: each provider's convergence logic, not the SDK, owns correctness.

What does this mean?

crates/breez-sdk/core/src/cross_chain/boltz_storage_adapter.rs
```diff
@@ -1,185 +1,184 @@
...
+            .list_active_cross_chain_swaps(PROVIDER_TAG_BOLTZ.to_string())
+            .await
+            .map_err(|e| BoltzError::Store(format!("Failed to list active swaps: {e}")))?;
+        let mut swaps = Vec::with_capacity(rows.len());
+        for stored in rows {
+            let id = stored.id.clone();
+            // Skip row on error (bad base64/json, decryption).
+            match self.swap_from_stored(stored).await {
+                Ok(swap) => swaps.push(swap),
+                Err(e) => warn!("Skipping Boltz swap '{id}': failed to load from storage: {e}"),
```
erdemyerebasmaz, line 163: What happens to the skipped swaps? Are they recovered on next retry? Should we fire off an SDK event here about the swap state?

erdemyerebasmaz (commented):
One question.

crates/breez-sdk/core/src/cross_chain/boltz_storage_adapter.rs
dangeross: They are retried. In practice decryption errors shouldn't happen. The encryption is deterministic from the wallet identity. I don't think we should fire off an event as there is no user action to take

crates/breez-sdk/core/src/persist/mod.rs
```diff
@@ -340,6 +340,48 @@ where
...
+    /// so `list_active_cross_chain_swaps` filters without parsing `data`.
+    pub is_terminal: bool,
+    /// Lifted from the underlying swap's `updated_at` into a column so the
+    /// row's freshness is inspectable without parsing `data`.
+    pub updated_at: u64,
+    /// Provider-opaque JSON. Stored verbatim so unknown fields survive across
+    /// versions. A plain string (not a structured value) so the row crosses
+    /// every backend, the WASM boundary, and the foreign-language FFI
+    /// uniformly as TEXT.
+    pub data: String,
```
JssDWt, line 366: I want to note that the contents of this field will always be replaced. Unlike the fields on StoredCrossChainSwap itself. When that is synced over real-time sync, any field added by a new version will also survive the round trip. That would not be the case in the `data` field. I think that's probably fine (is it?), but worth keeping in mind that older versions might delete fields that a newer version added.

JssDWt (commented):
I think LGTM, just a question and a note.

roeierez (approved):

crates/breez-sdk/core/src/realtime_sync/storage.rs
dangeross: I've tightened the docs and removed this comment

erdemyerebasmaz (approved):

JssDWt (approved):

PR #964, opened 2026-06-23
Cross-chain: Orchestra receive

dangeross:
Adds Orchestra-backed cross-chain receive: external chain → Spark (BTC sats or USDB). Companion to the existing cross-chain send.

### What's new

- New `ReceivePaymentMethod::CrossChain { route, amount, destination, fee_mode, max_slippage_bps, target_overpay_bps }` variant.
- `amount` is denominated in the source asset's base units (`route.decimals`), matching `prepare_payment_link`. USD-stable sources sit at USD parity, so `1_000_000` is 1 USDC (6dp) ≈ \$1. `FeesExcluded` (default) treats it as the receiver's target on Spark (translated at parity for USDB or via the live BTC/USD rate for Bitcoin) and the SDK sizes the sender's deposit. `FeesIncluded` treats it as the deposit and the receiver lands `amount − fees`.
- `destination` auto-picks the wallet's active stable-balance token if the route supports it, otherwise Bitcoin. Explicit values must appear in `route.accepted_assets`.
- `ReceivePaymentResponse.cross_chain_info` carries the deposit amount, expected received amount, destination denomination, provider fee (`service_fee_amount` + `service_fee_asset`, symmetric with the send-side field), and quote expiry.
- New `CrossChainService::prepare_receive` trait method. Orchestra implements it; Boltz returns `InvalidInput`.
- Sizing pads the deposit by `estimate.fee_amount` to absorb Orchestra's per-order settlement charge that only appears at quote time, so small-amount (~\$1) receives no longer trip the drift check on the first try.
- Provider-response guards refuse a same-chain route (Orchestra's on-Spark AMM shouldn't appear as a cross-chain option), a deposit address whose family doesn't match `route.chain`, and a quote whose `amount_in` echoes back materially different from what we requested.
- Drift error names the actionable lever: `Retry with a larger target_overpay_bps or send a bigger amount`.
- Unrelated to the feature: this also folds the several private copies of the current-time helper scattered across core into `utils::time` (`now_secs` / `now_ms` / `try_now_secs`), behavior-preserving.

### Breaking changes

Two enums and two `CrossChainRoutePair` fields are renamed so the names read naturally from both send and receive:

- `SourceAsset` → `SparkAsset` (enum)
- `SourceChain` → `DeliveryMethod` (enum)
- `CrossChainRoutePair.supported_sources` → `accepted_assets`
- `CrossChainRoutePair.supported_source_chains` → `delivery_methods`

Renames propagate through UniFFI / wasm-bindgen / FRB, so downstream integrators referencing the old names will need to update.

### DIFF
--- a/crates/breez-sdk/core/src/cross_chain/orchestra.rs
+++ b/crates/breez-sdk/core/src/cross_chain/orchestra.rs
@@ -12,8 +12,8 @@ use breez_sdk_common::fiat::FiatService;
 use breez_sdk_common::input::CrossChainAddressFamily;
 use chrono::DateTime;
 use flashnet::orchestra::{
-    AmountMode, EstimateRequest, EstimateResponse, OrderStatus, QuoteRequest, QuoteResponse, Route,
-    RouteAsset, StatusResponse, SubmitResponse,
+    AmountMode, EstimateRequest, EstimateResponse, Order, OrderStatus, QuoteRequest, QuoteResponse,
+    Route, RouteAsset, StatusResponse, SubmitResponse,
 };
 use flashnet::{FlashnetError, OrchestraClient, OrchestraConfig, OrchestraConfigResolver};
 use platform_utils::time::{Duration, SystemTime, UNIX_EPOCH};
@@ -30,13 +30,15 @@ use crate::persist::{ConversionFilter, StorageListPaymentsRequest, StoragePaymen
 use crate::{ConversionInfo, ConversionStatus, Payment, PaymentDetails, PaymentStatus, Storage};
 
 use super::{
-    CrossChainFeeMode, CrossChainPrepared, CrossChainProvider, CrossChainProviderContext,
-    CrossChainRouteFilter, CrossChainRoutePair, CrossChainService, SourceAsset, SourceChain,
-    derive_btc_leg_transfer_id, payment_with_conversion_info,
+    CrossChainFeeMode, CrossChainProvider, CrossChainProviderContext, CrossChainReceiveInfo,
+    CrossChainReceivePrepared, CrossChainRouteFilter, CrossChainRoutePair, CrossChainSendPrepared,
+    CrossChainService, DeliveryMethod, SparkAsset, derive_btc_leg_transfer_id,
+    orchestra_storage_adapter::{OrchestraStorageAdapter, OrchestraSwapData},
+    payment_with_conversion_info,
 };
 
 use crate::utils::{
-    payments::fetch_and_process_payment,
+    payments::{fetch_and_process_payment, resolve_and_insert_payment_metadata},
     polling::{PollSchedule, poll_until},
 };
 
@@ -45,24 +47,24 @@ const SOURCE_CHAIN_SPARK: &str = "spark";
 const SOURCE_CHAIN_LIGHTNING: &str = "lightning";
 const SOURCE_CHAIN_BITCOIN: &str = "bitcoin";
 
-/// The Orchestra wire value for a [`SourceChain`].
-fn source_chain_to_wire(source_chain: SourceChain) -> &'static str {
-    match source_chain {
-        SourceChain::Spark => SOURCE_CHAIN_SPARK,
-        SourceChain::Lightning => SOURCE_CHAIN_LIGHTNING,
-        SourceChain::Bitcoin => SOURCE_CHAIN_BITCOIN,
+/// The Orchestra `source_chain` wire value for a [`DeliveryMethod`].
+fn delivery_method_to_wire(delivery_method: DeliveryMethod) -> &'static str {
+    match delivery_method {
+        DeliveryMethod::Spark => SOURCE_CHAIN_SPARK,
+        DeliveryMethod::Lightning => SOURCE_CHAIN_LIGHTNING,
+        DeliveryMethod::Bitcoin => SOURCE_CHAIN_BITCOIN,
     }
 }
 
-/// Parses an Orchestra `source_chain` wire value, or `None` for one that isn't
-/// a local funding chain.
-fn source_chain_from_wire(chain: &str) -> Option<SourceChain> {
+/// Parses an Orchestra `source_chain` wire value into a [`DeliveryMethod`],
+/// or `None` for one that isn't a local delivery rail.
+fn delivery_method_from_wire(chain: &str) -> Option<DeliveryMethod> {
     if chain.eq_ignore_ascii_case(SOURCE_CHAIN_SPARK) {
-        Some(SourceChain::Spark)
+        Some(DeliveryMethod::Spark)
     } else if chain.eq_ignore_ascii_case(SOURCE_CHAIN_LIGHTNING) {
-        Some(SourceChain::Lightning)
+        Some(DeliveryMethod::Lightning)
     } else if chain.eq_ignore_ascii_case(SOURCE_CHAIN_BITCOIN) {
-        Some(SourceChain::Bitcoin)
+        Some(DeliveryMethod::Bitcoin)
     } else {
         None
     }
@@ -73,6 +75,24 @@ const DEFAULT_AFFILIATE_ID: &str = "breez_sdk";
 const SEND_POLL_INITIAL_DELAY_MS: u64 = 500;
 const SEND_POLL_MAX_DELAY_MS: u64 = 2000;
 const SEND_POLL_TIMEOUT_SECS: u64 = 30;
+/// Grace period to keep probing a receive quote before giving up.
+const RECEIVE_GRACE_SECS: u64 = 24 * 60 * 60;
+
+/// One-cent margin (in 6-decimal USDB base units) covering Orchestra's
+/// post-quote rounding drift on USDB deliveries: `estimated_out` rounds up
+/// to the next cent, actual delivery rounds down, so `1_000_000` can arrive
+/// as `990_000`. Applied at both ends of the pipeline: on sizing, so the
+/// deposit is inflated enough for delivery to land at or above `FeesExcluded`
+/// target; on reporting, so `expected_received_amount` doesn't over-promise
+/// the receiver. The pad is meaningful at the $1 boundary (1% of value) and
+/// negligible above roughly $10.
+const USDB_RECEIVE_ROUNDING_MARGIN: u128 = 10_000;
+
+/// Basis points of slack allowed between the source amount we requested
+/// (`ExactIn`) and the `amount_in` Orchestra echoes back on the quote.
+/// Legitimate rounding at provider-side precision boundaries is bounded.
+/// Anything larger is provider misbehavior we refuse to persist.
+const QUOTE_AMOUNT_IN_TOLERANCE_BPS: u32 = 10;
 
 /// Resolves the Orchestra config from Breez server.
 ///
@@ -110,7 +130,7 @@ impl OrchestraConfigResolver for BreezServerOrchestraConfigResolver {
 
 /// Source-side identity of an Orchestra route after `(dest, source)` matching.
 #[derive(Clone, Debug, PartialEq, Eq)]
-struct ResolvedSourceAsset {
+struct ResolvedSparkAsset {
     /// Wire symbol (e.g. `"BTC"`, `"USDB"`).
     asset: String,
     /// Source-asset decimals.
@@ -164,8 +184,10 @@ impl OrchestraService {
         monitor_trigger: &broadcast::Sender<()>,
     ) {
         let storage = Arc::clone(&self.storage);
+        let swap_storage = OrchestraStorageAdapter::new(Arc::clone(&storage));
         let client = Arc::clone(&self.client);
         let spark_wallet = Arc::clone(&self.spark_wallet);
+        let fiat_service = Arc::clone(&self.fiat_service);
         let mut trigger_receiver = monitor_trigger.subscribe();
         let span = tracing::Span::current();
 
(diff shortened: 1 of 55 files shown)


docs/breez-sdk/snippets/rust/src/cross_chain.rs
```diff
@@ -85,3 +85,60 @@ async fn send_payment_cross_chain(
...
+                target_overpay_bps: optional_target_overpay_bps,
+            },
+        })
+        .await?;
+
+    info!("Payment request: {}", response.payment_request);
+    if let Some(info) = response.cross_chain_info {
+        info!("Deposit address: {}", info.deposit_address);
+        info!("Deposit amount: {}", info.deposit_amount);
+        let denom = if info.token_identifier.is_some() { "USDB" } else { "BTC" };
```
danielgranhao: This line concerns me a bit. Do you think it makes sense to state in the docs that the token here can only be USDB like this? (same for all snippets)

crates/breez-sdk/core/src/sdk/api.rs
```diff
@@ -588,10 +588,10 @@ impl BreezSdk {
 }
 
 /// Whether `route` advertises `required` as a fundable source chain. An empty
-/// `supported_source_chains` (e.g. a hand-built route) matches nothing and is
+/// `delivery_methods` (e.g. a hand-built route) matches nothing and is
 /// rejected.
-fn route_supports_source_chain(route: &CrossChainRoutePair, required: SourceChain) -> bool {
-    route.supported_source_chains.contains(&required)
+fn route_supports_source_chain(route: &CrossChainRoutePair, required: DeliveryMethod) -> bool {
```
danielgranhao: nit: can be renamed to e.g. `route_supports_delivery_method`

crates/breez-sdk/core/src/sdk/api.rs
```diff
@@ -720,7 +720,7 @@ mod tests {
         }
     }
 
-    fn route_with_source_chains(chains: Vec<SourceChain>) -> CrossChainRoutePair {
+    fn route_with_source_chains(chains: Vec<DeliveryMethod>) -> CrossChainRoutePair {
```
danielgranhao:
```suggestion
    fn route_with_delivery_methods(methods: Vec<DeliveryMethod>) -> CrossChainRoutePair {
```

crates/breez-sdk/core/src/sdk/api.rs
```diff
@@ -729,35 +729,39 @@ mod tests {
...
             exact_out_eligible: false,
-            supported_sources: vec![],
-            supported_source_chains: chains,
+            accepted_assets: vec![],
+            delivery_methods: chains,
         }
     }
 
     #[test_all]
     fn route_supports_source_chain_checks_membership() {
```
danielgranhao:
```suggestion
    fn route_supports_delivery_method_checks_membership() {
```

crates/breez-sdk/core/src/cross_chain/orchestra.rs
```diff
@@ -449,6 +488,180 @@ impl OrchestraService {
...
+            .status_by_id(order_id, data.read_token.as_deref())
+            .await
+        {
+            Ok(resp) => resp,
+            Err(e) => {
+                debug!(
+                    "Orchestra receive {quote_id}: status request failed for orderId={order_id}: {e}"
+                );
+                return Ok(());
+            }
```
danielgranhao, line 681: We should handle expired read tokens as we do on the send path, so we don't just return here indefinitely.

crates/breez-sdk/core/src/sdk/payments/receive.rs
```diff
@@ -86,7 +94,188 @@ pub(super) async fn receive_payment(
...
+        CrossChainFeeMode, convert_source_amount_to_sats, fetch_btc_usd_rate, rescale_decimals,
+    };
+    let src_decimals = u32::from(route.decimals);
+    match (fee_mode, destination) {
+        (CrossChainFeeMode::FeesIncluded, _) => Ok(amount),
+        (CrossChainFeeMode::FeesExcluded, SparkAsset::Token { .. }) => {
+            rescale_decimals(amount, src_decimals, 6)
+        }
+        (CrossChainFeeMode::FeesExcluded, SparkAsset::Bitcoin) => {
+            let btc_usd = fetch_btc_usd_rate(sdk.fiat_service.as_ref()).await?;
```
danielgranhao: Shouldn't this use `sdk.cross_chain_context.fiat_service()` instead of the raw fiat service?

crates/breez-sdk/core/src/cross_chain/orchestra.rs
```diff
@@ -557,6 +827,43 @@ fn proportional_inflation(
...
+    let applied = reported_pad.min(scaled);
+    (scaled.saturating_add(applied), applied)
+}
+
+/// Whether an Orchestra error on the receive-side `/submit` probe is the
+/// expected "no deposit yet" shape (any 4xx). Anything else (5xx, 429,
+/// transport failure) is a transient provider issue worth logging louder.
+fn is_expected_no_deposit_error(err: &FlashnetError) -> bool {
+    matches!(err, FlashnetError::Network { code: Some(c), .. } if *c < 500)
+}
```
danielgranhao, line 919: 429 is below 500 so a rate limit gets swallowed (goes agaisnt rustdoc above). Maybe `*c < 500 && *c != 429`?

crates/breez-sdk/core/src/cross_chain/orchestra.rs
```diff
@@ -1151,13 +1723,145 @@ fn apply_terminal_status(
...
+        return Ok(false);
+    };
+    let conversion_info = build_orchestra_receive_conversion_info(data, order, fiat_service).await;
+    let metadata = crate::PaymentMetadata {
+        conversion_info: Some(conversion_info),
+        ..Default::default()
+    };
+    // tx_inputs_are_ours = false: on receive, the inbound token tx is funded
+    // by Orchestra's counterparty, not us.
+    resolve_and_insert_payment_metadata(spark_tx_hash, metadata, spark_wallet, storage, false)
```
danielgranhao: This returns Ok whether the metadata lands on the row or only cache so the caller marks row terminal and stops polling. For token destinations, that means a get_token_transactions_by_hashes failure may lose the conversion info. Could we tell the two apart and keep polling?

crates/breez-sdk/core/src/cross_chain/orchestra.rs
```diff
@@ -1151,13 +1723,145 @@ fn apply_terminal_status(
...
+        order_id: order.id.clone(),
+        quote_id: data.quote_id.clone(),
+        read_token: None,
+        chain: data.source_chain.clone(),
+        chain_id: data.source_chain_id.clone(),
+        asset: data.source_asset.clone(),
+        recipient_address: data.recipient_address.clone(),
+        asset_amount_in,
+        estimated_out,
+        delivered_amount,
```
danielgranhao, line 1914: On receive, these 2 fields are Spark-side units while `asset` and `asset_decimals` describe the external side. I think this might not be clear enough. Could the variant docs spell out the receive direction? Three spots in token_conversion/models.rs: the enum header at 163 still says Orchestra is Spark → external only, fee_amount at 228 says "destination asset base units" while receive puts source units there, and estimated_out / delivered_amount at 221-224 don't state a denomination at all.

docs/breez-sdk/snippets/rust/src/cross_chain.rs
danielgranhao: Maybe we can include in the response `destination_asset_symbol` and use that?

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -1612,6 +1612,38 @@ pub enum ReceivePaymentMethod {
...
+        /// USD-stable sources are at parity, so `1 USD = 10^route.decimals`
+        /// (e.g. `1_000_000` for 6-decimal USDC/USDT, `10^18` for 18-decimal
+        /// BSC USDC).
+        ///
+        /// - `FeesExcluded` (default): what the receiver ends up with, sized
+        ///   as if `amount` source units were converted to the Spark-side
+        ///   destination at parity (USDB) or the live BTC/USD rate (Bitcoin).
+        /// - `FeesIncluded`: what the sender deposits. The receiver ends up
+        ///   with that amount minus provider fees.
+        amount: u128,
```
danielgranhao, line 1630: For fees excluded, if someone wants to denominate amount in sats currently they would have to convert sats to source asset only for it to be immediately converted back internally. Do you think it's worth already supporting defining the amount in sats? Maybe it adds too much complexity?

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -1612,6 +1612,38 @@ pub enum ReceivePaymentMethod {
...
+        /// Spark-side asset the receiver wants delivered. When absent, the
+        /// SDK auto-selects: the wallet's active stable-balance token if
+        /// the route supports it, otherwise Bitcoin (sats). When set, the
+        /// value must appear in the route's `accepted_assets`.
+        destination: Option<crate::cross_chain::SparkAsset>,
+        /// How `amount` should be interpreted. When absent, defaults to
+        /// `FeesExcluded`.
+        fee_mode: Option<crate::cross_chain::CrossChainFeeMode>,
```
danielgranhao, line 1638: The rustdocs of `CrossChainFeeMode` are still written only with send in mind.

crates/breez-sdk/core/src/cross_chain/orchestra.rs
```diff
@@ -73,6 +75,24 @@ const DEFAULT_AFFILIATE_ID: &str = "breez_sdk";
 const SEND_POLL_INITIAL_DELAY_MS: u64 = 500;
 const SEND_POLL_MAX_DELAY_MS: u64 = 2000;
 const SEND_POLL_TIMEOUT_SECS: u64 = 30;
+/// Grace period to keep probing a receive quote before giving up.
+const RECEIVE_GRACE_SECS: u64 = 24 * 60 * 60;
```
danielgranhao, line 83: IIUC, each abandoned quote can trigger up to 2880 requests (POST /submit every 30 seconds). 

Not sure about the best way to deal with it, though. Ideally there would be events from orchestra. In their absense, maybe a shorter period where we probe? Or we could reduce polling frequency after expiry?

danielgranhao (commented):
Nice!

crates/breez-sdk/core/src/cross_chain/orchestra.rs
roeierez, line 1734: For the receive won't it suppress the payment PaymentSucceeded/PaymentPending events? In cases sync is running after applying the conversion info (suppressed by the middleware).

crates/breez-sdk/core/src/sdk/payments/receive.rs
```diff
@@ -86,7 +94,189 @@ pub(super) async fn receive_payment(
...
+    // `get_cross_chain_routes` filters the discovery list to USD-stable
+    // sources. Re-check here so a hand-built route can't bypass the USD-par
+    // assumption that `convert_receive_amount_to_provider_units` relies on.
+    if !crate::cross_chain::is_usd_stable_asset(&route.asset) {
+        return Err(SdkError::InvalidInput(format!(
+            "Cross-chain receive source asset must be USD-stable. Got {}",
+            route.asset
+        )));
+    }
+    let slippage = max_slippage_bps.unwrap_or(DEFAULT_CROSS_CHAIN_SLIPPAGE_BPS);
```
roeierez: Should we use the CrossChainConfig::default_slippage_bps ?

crates/breez-sdk/core/src/sdk/payments/receive.rs
```diff
@@ -86,7 +94,189 @@ pub(super) async fn receive_payment(
...
+        target_overpay_bps,
+        sdk.config
+            .cross_chain_config
+            .as_ref()
+            .and_then(|c| c.default_target_overpay_bps),
+    )?;
+
+    let resolved_destination = resolve_receive_destination(sdk, &route, destination).await?;
+
+    let provider_amount = convert_receive_amount_to_provider_units(
```
roeierez, line 159: Worth having a zero check on the provider_amount as well? Given it can be zero after conversion?

crates/breez-sdk/core/src/cross_chain/orchestra.rs
```diff
@@ -449,6 +498,232 @@ impl OrchestraService {
...
+            }
+            OrderStatus::Failed | OrderStatus::Refunded => {
+                info!(
+                    "Orchestra receive {quote_id} → {:?}, closing row without metadata",
+                    order.status
+                );
+                swap_storage.mark_terminal(row).await?;
+            }
+            // Non-terminal status.
+            _ => {}
```
roeierez, line 729: are we missing cases here? What's the case for the other statues and can the order stay stuck forever?

crates/breez-sdk/core/src/cross_chain/orchestra.rs
```diff
@@ -773,9 +1192,213 @@ impl CrossChainService for OrchestraService {
...
+            destination_decimals,
+            token_identifier: destination_token_identifier.clone(),
+            amount_in: quote.amount_in.clone(),
+            expected_amount_out: expected_received_amount.to_string(),
+            fee_amount: Some(quote.total_fee_amount.clone()),
+            expires_at: expires_at_secs,
+        };
+
+        let adapter = OrchestraStorageAdapter::new(Arc::clone(&self.storage));
+        adapter.upsert(&data).await?;
```
roeierez: Should we first call the build_receive_payment_request and then upsert? so we will not end up with a row on cases where validation failed?

roeierez (changes requested):
Looks good

crates/breez-sdk/core/src/cross_chain/orchestra.rs
dangeross: Other statuses are mostly progression states: Processing, Confirming, Bridging, Swapping, Refunding, Delivering with no additional/final information included.

crates/breez-sdk/core/src/cross_chain/orchestra.rs
dangeross: Fixed, but this only applies to spark side payments and the middleware suppressing all payments with conversion_info set. If the orchestra conversion_info becomes terminal later it doesn't emit the update. I tried to address this in another (stale) PR, which I'll take another look at: https://github.com/breez/spark-sdk/pull/965

danielgranhao (approved):
LGTM

roeierez (approved):

PR #965, opened 2026-06-24
Add PaymentUpdated event

dangeross:
Adds `SdkEvent::PaymentUpdated { payment }` so integrators hear about metadata that lands on a payment after it has already settled. Today those updates are silent and the UI stays stale until something else prompts a refresh.

### Why

A cross-chain receive settles as an ordinary Spark transfer, so `PaymentSucceeded` fires before the provider has confirmed the conversion. When the provider's poller later completes the order, the conversion details are written to storage and nothing tells the app, leaving the user looking at what seems to be a plain BTC receive. Cross-chain send and the AMM conversion paths have the same shape.

### What fires

`PaymentUpdated` fires when a payment is already `Completed` or `Failed` and its stored metadata actually changed. It carries the full payment, conversion details included. Payments still in flight are unaffected and keep surfacing through the existing status events.

One rule decides who sees it: if a payment's success was hidden from external listeners, its update is hidden too. `TokenConversionMiddleware` reuses the gate it already applies to the other payment events, so internal AMM legs stay silent while cross-chain payments come through.

`send_payment` writes its own conversion status after the send settles and stays silent, because the enriched payment is returned to the caller synchronously. A listener-only observer on the same instance will not see that one.

### No duplicate events

Storage now reports whether a write actually changed the row, and the event fires only when it did. This matters because the provider monitors rewrite the same values on every poll, and without the check they would emit continuously. Each backend implements the check in whatever way its engine allows, and all of them share one test. Metadata reapplied during the initial sync stays silent, since the status event in the same sync loop already carries the enriched payload.

### Multi-device

Metadata synced from another device now surfaces on the receiving one, so a conversion confirmed on one device updates the payment on the others.

Each session stays silent for its first pull, the catch-up that replays everything missed since the last one, which on a restored wallet is the entire history. Later pulls emit, so nothing is lost when the notification subscription drops and re-establishes. This mirrors the payments sync, which also stays silent on its initial pass.

### Stable balance

The same rule now covers stable balance instead of being decided per call site. Per-receive conversions emit when they finish or fail: the parent is an ordinary inbound payment, already delivered to the app marked as converting, and nothing told the app it was done. On a server there was no signal at all.

One thing worth a look in review:

- A failed metadata write now caches for the next sync instead of failing the conversion task. The funds have already moved by then, so retrying the whole conversion was riskier.

### Boltz

Boltz is wired here but dormant: #1105 unregistered the provider. Wiring it now means re-enabling the provider restores the event in one place, rather than needing this redone.


crates/breez-sdk/wasm/src/persist/mod.rs
```diff
@@ -158,15 +158,20 @@ impl breez_sdk_spark::Storage for WasmStorage {
         &self,
         payment_id: String,
         metadata: breez_sdk_spark::PaymentMetadata,
-    ) -> Result<(), StorageError> {
+    ) -> Result<bool, StorageError> {
```
danielgranhao, line 161: The ts interface lower in this file needs to be updated. Still says `insertPaymentMetadata: (...) => Promise<void>`

crates/breez-sdk/core/src/realtime_sync/storage.rs
```diff
@@ -382,9 +382,20 @@ impl SyncedRecordHandler {
...
+            .storage
+            .insert_payment_metadata(data_id.clone(), metadata)
+            .await?
+        {
+            crate::utils::payments::emit_payment_updated_if_terminal(
+                &self.storage,
+                &self.event_emitter,
+                data_id,
+            )
+            .await;
```
danielgranhao, line 407: When restoring an existing wallet, won't we fire a big `PaymentUpdated` burst? Maybe worth skipping that first pass?

crates/breez-sdk/core/src/stable_balance/mod.rs
```diff
@@ -323,7 +323,18 @@ impl StableBalance {
...
+        // The cleared payments already carry their Failed status; surface it on
+        // the ones whose success reached the app.
+        for payment_id in changed_payment_ids {
+            crate::utils::payments::emit_payment_updated_if_terminal(
+                &self.core.storage,
+                &self.event_emitter,
+                payment_id,
+            )
+            .await;
+        }
```
danielgranhao, line 336: The timeout sweep fails a per-receive task without emitting, so that ending is the one left to cover. Could it move into the conversion worker's drain loop, where the emitter is already in scope next to the Completed write?

danielgranhao (commented):
Looks good

crates/breez-sdk/core/src/stable_balance/queue.rs
```diff
@@ -305,6 +304,8 @@ impl StableBalance {
                     // Register notify future BEFORE checking the queue to avoid missed wakeups
                     let notified = stable_balance.core.queue.notify.notified();
 
+                    stable_balance.expire_deferred_tasks().await;
```
danielgranhao, line 307: Once a task is deferred, pending_task stays Some and push_auto_convert stops notifying, so the worker never wakes to run this and auto-convert stays blocked behind it. Could the Synced arm call notify.notify_one() directly?

crates/breez-sdk/core/src/stable_balance/mod.rs
```diff
@@ -629,26 +650,6 @@ impl EventMiddleware for StableBalanceMiddleware {
         match event {
             // Sync completed → wake the startup gate, sweep timed-out deferred tasks
```
danielgranhao, line 651: Out of date: still says it sweeps here, and the module diagram at line 29 does too.

danielgranhao (approved):
LGTM

danielgranhao:
Superseded by #1157

PR #966, opened 2026-06-25
fix(wasm/passkey): honor preferImmediatelyAvailableCredentials on web

erdemyerebasmaz:
## Problem

The web PRF provider accepted `preferImmediatelyAvailableCredentials` but ignored it: `_performAssertion` called `navigator.credentials.get()` with no `mediation`, so web assertions always ran modal. As Chrome rolls out immediate mediation (enabled by default on some platforms in recent versions), single-CTA hosts switched to the immediate path but still got the account/QR sheet, because the SDK never requested it.

## Fix (two parts)

1. Set `mediation: 'immediate'` when the flag is set and the browser advertises `immediateGet` (via `getClientCapabilities()`). The capability gate is required: passing the unknown enum to `get()` throws `TypeError` on browsers without L3 immediate support.
2. Classify the fast no-credential reject only on the immediate path. `immediate` is the only mediation where a no-credential reject shows no UI; the modal path always shows a picker first, so a NotAllowedError there is a dismiss/timeout, never a fast no-credential. `_mapAssertionError` now applies the fast-fail window only when immediate.

Also logs at debug when the flag is requested but `immediateGet` isn't advertised, so the silent fallback to standard mediation is visible. JS-only beyond the `connect_with_passkey` doc; native unaffected.

### DIFF
--- a/crates/breez-sdk/core/src/passkey/passkey_client.rs
+++ b/crates/breez-sdk/core/src/passkey/passkey_client.rs
@@ -272,11 +272,11 @@ impl PasskeyClient {
     /// register path; every other error (`Cancel`, `Timeout`, ...) propagates
     /// unchanged.
     ///
-    /// Mobile-only: meant for iOS 18+ / Android 9+ where
-    /// `preferImmediatelyAvailableCredentials` is honored. The web
-    /// equivalent (`mediation: 'immediate'`) is not yet stable
-    /// cross-browser, so this is not surfaced on WASM; web hosts call
-    /// [`Self::sign_in`] and catch `CredentialNotFound` themselves.
+    /// Not surfaced on WASM: web hosts drive the same flow via
+    /// [`Self::sign_in`] with `prefer_immediately_available_credentials =
+    /// true` and catch `CredentialNotFound`. The web provider maps the flag
+    /// to `mediation: 'immediate'` where `getClientCapabilities().immediateGet`
+    /// advertises it, falling back to the standard picker elsewhere.
     pub async fn connect_with_passkey(
         &self,
         request: ConnectWithPasskeyRequest,
--- a/crates/breez-sdk/core/src/passkey/passkey_client.rs
+++ b/crates/breez-sdk/core/src/passkey/passkey_client.rs
@@ -275,7 +275,7 @@ impl PasskeyClient {
     /// Not surfaced on WASM: web hosts drive the same flow via
     /// [`Self::sign_in`] with `prefer_immediately_available_credentials =
     /// true` and catch `CredentialNotFound`. The web provider maps the flag
-    /// to `mediation: 'immediate'` where `getClientCapabilities().immediateGet`
+    /// to `uiMode: 'immediate'` where `getClientCapabilities().immediateGet`
     /// advertises it, falling back to the standard picker elsewhere.
     pub async fn connect_with_passkey(
         &self,
--- a/crates/breez-sdk/wasm/js/passkey-prf-provider/index.js
+++ b/crates/breez-sdk/wasm/js/passkey-prf-provider/index.js
@@ -22,10 +22,11 @@ const BREEZ_RP_ID = 'keys.breez.technology';
 /** Default `rpName` for the zero-config {@link PasskeyClient} path. */
 const DEFAULT_RP_NAME = 'Breez';
 
-// WebAuthn collapses "no matching credential" and "user dismissed" into
-// one NotAllowedError, but a no-credential fast-fail resolves before any
-// UI shows while a dismiss takes seconds. Elapsed time below this is
-// classified as no-credential, at or above as user-cancel.
+// A no-credential reject resolves before any UI shows; a dismiss takes
+// seconds. This fast-fail window only applies under `mediation:
+// 'immediate'`: that is the only path where a no-credential reject shows
+// no UI. On the modal path the OS always shows a picker first, so a
+// NotAllowedError there is a dismiss/timeout, never a fast no-credential.
 const NO_CRED_FAST_FAIL_MS = 250;
 
 // iOS and Android tear the biometric sheet down around 55s of inactivity
@@ -42,6 +43,27 @@ function randomBytes(length) {
     const bytes = new Uint8Array(length);
     crypto.getRandomValues(bytes);
     return bytes;
+}
+
+/**
+ * Whether this browser honors `mediation: 'immediate'`, probed via
+ * `getClientCapabilities().immediateGet`. The gate is mandatory: the
+ * unknown `mediation` enum throws a `TypeError` on browsers without
+ * WebAuthn L3 support. Uncached: the probe is cheap and runs at most once
+ * per assertion.
+ * @returns {Promise<boolean>}
+ */
+async function supportsImmediateMediation() {
+    try {
+        if (typeof PublicKeyCredential === 'undefined'
+            || typeof PublicKeyCredential.getClientCapabilities !== 'function') {
+            return false;
+        }
+        const caps = await PublicKeyCredential.getClientCapabilities();
+        return caps?.immediateGet === true;
+    } catch {
+        return false;
+    }
 }
 
 /**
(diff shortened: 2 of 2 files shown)


crates/breez-sdk/wasm/js/passkey-prf-provider/index.js
```diff
@@ -434,6 +459,15 @@ export class PasskeyProvider {
...
 
         const requestOptions = { publicKey };
+        // Immediate mediation: fast-fail with no UI when no credential is
+        // present so a single-CTA host can fall through to register. Only
+        // where the browser advertises it (else `get()` throws on the
+        // unknown `mediation` enum). The fast NotAllowedError this raises
+        // on no-credential is classified as CredentialNotFound below.
+        if (options.preferImmediatelyAvailableCredentials
+            && await supportsImmediateMediation()) {
+            requestOptions.mediation = 'immediate';
```
dangeross: `connect_with_passkey`'s doc (`crates/breez-sdk/core/src/passkey/passkey_client.rs:277-279`) says the web `mediation: 'immediate'` equivalent "is not yet stable cross-browser, so this is not surfaced on WASM." This PR implements it, gated on `getClientCapabilities().immediateGet`, so the "not yet stable" framing is now stale.

The helper itself does stay native-only (it isn't bound on the WASM client), but web hosts now reach this path via `signIn({ preferImmediatelyAvailableCredentials: true })`, so the doc reads as more dismissive of web than the code now is.

erdemyerebasmaz: I'm thinking of surfacing `connect_with_passkey` on WASM now that `immediateGet` has made it to stable Chrome, but it needs some kind of guard mechanism compared to native implementations that checks if `immediateGet` is supported or not.

erdemyerebasmaz: https://github.com/breez/spark-sdk/pull/967

crates/breez-sdk/wasm/js/passkey-prf-provider/index.js
```diff
@@ -434,6 +457,22 @@ export class PasskeyProvider {
...
+        // Immediate UI mode: fast-fail with no UI when no credential is
+        // present so a single-CTA host can fall through to register. Only
+        // where the browser advertises it (`getClientCapabilities().immediateGet`).
+        // The fast NotAllowedError it raises on no-credential is classified
+        // as CredentialNotFound below. Immediate mode is discoverable-only:
+        // a non-empty allowCredentials throws, so clear the pin here.
+        if (options.preferImmediatelyAvailableCredentials) {
+            if (await supportsImmediateMediation()) {
+                requestOptions.uiMode = 'immediate';
+                publicKey.allowCredentials = [];
```
dangeross: `allowCredentials = []` here runs on every immediate-mode assertion, including the pinned later ceremonies of a multi-salt derive, which collides with the pin built just above (lines 432-436). `setup_wallet` always derives `[ACCOUNT_MASTER_SALT, label]` and forwards the flag, so when the authenticator drops `prf.eval.second` the second ceremony is pinned and then un-pinned right here. Immediate mode rejects a non-empty `allowCredentials` (spec safeguard) and the pin requires one, so they are structurally incompatible.

Two ways it surfaces:
- Immediate mode also requires transient activation. The second ceremony runs after the first `get()` has spent the user gesture, so it likely rejects fast with `NotAllowedError` → classified as `CredentialNotFound` → `connect_with_passkey` then registers a duplicate passkey / a different wallet for a user who already has one. This needs only the dual-salt drop, not multiple passkeys.
- If activation survives and the user has more than one passkey, the immediate chooser can resolve to a different credential, so `seeds[0]` (Nostr identity) and `seeds[1]` (wallet mnemonic) derive from different passkeys, silently (`seeds.len()` still matches).

Could we gate immediate mode on `allowList.length === 0` (no pin, no caller `allowCredentials`) and fall back to the modal path otherwise? The fast-fail only matters for the first probe; once a credential is pinned we already know one exists.

crates/breez-sdk/core/src/passkey/passkey_client.rs
```diff
@@ -272,11 +272,11 @@ impl PasskeyClient {
...
     ///
-    /// Mobile-only: meant for iOS 18+ / Android 9+ where
-    /// `preferImmediatelyAvailableCredentials` is honored. The web
-    /// equivalent (`mediation: 'immediate'`) is not yet stable
-    /// cross-browser, so this is not surfaced on WASM; web hosts call
-    /// [`Self::sign_in`] and catch `CredentialNotFound` themselves.
+    /// Not surfaced on WASM: web hosts drive the same flow via
+    /// [`Self::sign_in`] with `prefer_immediately_available_credentials =
+    /// true` and catch `CredentialNotFound`. The web provider maps the flag
+    /// to `mediation: 'immediate'` where `getClientCapabilities().immediateGet`
```
dangeross: This writes the literal `mediation: 'immediate'`, but the web provider sets `uiMode: 'immediate'` (`index.js`). The field was renamed from `mediation: 'immediate'` to `uiMode: 'immediate'` in the Nov 2025 WebAuthn spec update, and the JS switched to `uiMode` in the second commit here, so the doc is now a step behind the code. Update the literal to `uiMode: 'immediate'`? (The same line recurs in #967's re-edit of this doc.)

dangeross (commented):
I'm leaning a lot on Claude here, would also suggest code reviewing locally

crates/breez-sdk/wasm/js/passkey-prf-provider/index.js
erdemyerebasmaz: [fix(wasm/passkey): only use immediate UI mode on the unpinned probe](https://github.com/breez/spark-sdk/pull/966/commits/3e628693935a310699fb9a8a12c410e5fae40ec4)

crates/breez-sdk/core/src/passkey/passkey_client.rs
erdemyerebasmaz: [fix(wasm/passkey): only use immediate UI mode on the unpinned probe](https://github.com/breez/spark-sdk/pull/966/commits/3e628693935a310699fb9a8a12c410e5fae40ec4)

crates/breez-sdk/core/src/passkey/passkey_client.rs
```diff
@@ -272,11 +272,11 @@ impl PasskeyClient {
...
     ///
-    /// Mobile-only: meant for iOS 18+ / Android 9+ where
-    /// `preferImmediatelyAvailableCredentials` is honored. The web
-    /// equivalent (`mediation: 'immediate'`) is not yet stable
-    /// cross-browser, so this is not surfaced on WASM; web hosts call
-    /// [`Self::sign_in`] and catch `CredentialNotFound` themselves.
+    /// Not surfaced on WASM: web hosts drive the same flow via
+    /// [`Self::sign_in`] with `prefer_immediately_available_credentials =
+    /// true` and catch `CredentialNotFound`. The web provider maps the flag
+    /// to `uiMode: 'immediate'` where `getClientCapabilities().immediateGet`
```
dangeross, line 278: Thanks for the `uiMode` fix here. The same literal lives one more place: the public TypeScript interface doc at `crates/breez-sdk/wasm/src/models/passkey_prf_provider.rs:300` (on `DeriveSeedOptions.preferImmediatelyAvailableCredentials`) still says it "maps to WebAuthn `mediation: 'immediate'`".

dangeross: Ok, this is fixed in #967

erdemyerebasmaz: I will merge the PRs and reframe it as a new feature for WASM.

erdemyerebasmaz:
Consolidated into #967, which now carries both the immediate-mediation provider work and the connect_with_passkey web surface. The Swift-BigInt toolchain bump is split into #969.

PR #968, opened 2026-06-26
core: Restore X-Partner-Jwt on SSP requests

mhrheaume:
This header is required for static deposits, where the JWT gets relayed to the SO through the SSP.


dangeross (approved):
LGTM

roeierez (approved):

PR #969, opened 2026-06-26
fix(swift): bump Swift-BigInt to 2.4.0 for current Swift toolchains

erdemyerebasmaz:
This PR bumps `Swift-BigInt` to `2.4.0`.

Swift bindings pinned `Swift-BigInt` at `2.0.0`, which doesn't build with current Swift toolchains, and the package failed to resolve on recent Xcode and broke CI (`CI / Docs / swift`).


dangeross (approved):
LGTM

PR #970, opened 2026-06-26
Support external signers that cannot export keys

dangeross:
A Turnkey wallet whose policy denies `EXPORT_WALLET_ACCOUNT` can't export the key the SDK uses to seed its local encryption (session storage, real-time sync, cross-chain swap secrets, LNURL-auth), nor the static-deposit key needed to claim/refund on-chain deposits.

This declares the signer's capability in the **type system** rather than a runtime flag: a signer either provides the full SDK-signer interface (with ECIES/HMAC) or a signing-only one.

### Signing-only signers

The internal SDK signer is split into three capability traits: `BreezSigner` (signing), `EciesSigner`, and `HmacSigner`. The external interface has two traits:

- `ExternalBreezSigner` — the full signer (signing + ECIES + HMAC), connected via `connect_with_signer` / `SdkBuilder::new_with_signer`.
- `ExternalSigningSigner` — signing only, connected via `connect_with_signing_only_signer` / `SdkBuilder::new_with_signing_only_signer` (with `ConnectWithSigningOnlySignerRequest` / `SigningOnlyExternalSigners`).

Implementing `ExternalSigningSigner` (or connecting via the signing-only path) *is* the declaration that the signer can't run the SDK's local ECIES/HMAC operations. With a signing-only signer the SDK:

- stores session tokens **unencrypted** (auth-only bearer tokens),
- **rejects the build** with `InvalidInput` if real-time sync or cross-chain is configured,
- **errors LNURL-auth** at call time with `Generic` (it needs the HMAC step).

### On-chain deposits (signer-driven)

On-chain (static) deposits are gated by the signer, independent of the full/signing-only choice: deriving a deposit address exports the static-deposit key via `get_static_deposit_public_key`, so a signer that can't export it fails there and an address that could never be claimed or refunded is never issued. `claim_deposit` / `refund_deposit` and sync auto-claim likewise fail naturally at the export (sync records the UTXOs as unclaimed).

### Turnkey

`create_turnkey_signer` returns a full signer (`ExternalBreezSigner`); `create_turnkey_signing_only_signer` returns a signing-only one (`ExternalSigningSigner`) for a wallet under a deny-export policy, which never exports a key.

The full Turnkey signer exports its ECIES/HMAC key lazily on first use (not eagerly at construction), so it never round-trips at connect. If a full signer is used against a deny policy (a misconfiguration: the deny-export path is the signing-only signer), the export fires on first ECIES/HMAC use and surfaces `EncryptionUnavailable`; the 403 denial is memoized so repeated calls fail fast instead of re-exporting and 403-storming Turnkey (transient failures still retry).

### Session-token encryption

`EncryptingSessionStore` encrypts tokens at rest when an `EciesSigner` is present (full signer) and stores them plaintext otherwise (signing-only), tagging each token with its mode (`enc:` / `pln:`). Reconnecting the same wallet with a different signer profile leaves a token written in the other mode; on read the marker mismatches (and a stale ciphertext also fails to decrypt), so it reads as `NotFound` and the caller re-authenticates and re-stores the token in the current mode. Detection is symmetric (both directions caught immediately, not on expiry); the marker rides in the token string (no schema change), and legacy untagged tokens read as `NotFound` and re-auth once on upgrade.

### Tests

- **Unit:** signer routing (`build_signers` yields no `ecies`/`rtsync`/`lnurl_auth` for a signing-only signer, all present for a full/seed signer); build-level rejection (a signing-only build with real-time sync configured fails with `InvalidInput`); LNURL-auth returns `Generic` for a signing-only signer; the signing-only external adapter matches a reference signer across derive / ECDSA / recoverable / Schnorr; and session-store behavior (native + WASM via the shared harness) — encrypt/plaintext round-trips with mode tagging, both mode-switch directions read as `NotFound`, and fail-closed (`encrypt_failure_never_writes_plaintext`: a signer that can't encrypt makes `set_session` error rather than persist a plaintext token).
- **Live deny-policy itests** (`--ignored`, need deny-export creds):
  - `turnkey_no_export_gates_onchain_receive` — signing-only signer + deny-export: connect + Spark receive work, on-chain receive fails at the static-deposit export.
  - `turnkey_export_denied_with_encryption_enabled_fails[_server_mode]` — full signer against a deny policy (misconfigured): `update_user_settings` fails at the encryption-key export (fail-closed, no plaintext fallback), in both client and server mode.

## Configuring a Turnkey policy for testing

Turnkey **root** API keys bypass policies, so reproduce the denial with a **non-root** API user: add an allow-all plus a deny-export policy, then run the SDK with its API keypair.

```json
{
  "policyName": "itest-allow-all",
  "effect": "EFFECT_ALLOW",
  "consensus": "true",
  "condition": "true"
}
```
```json
{
  "policyName": "itest-deny-export",
  "effect": "EFFECT_DENY",
  "consensus": "true",
  "condition": "activity.action == 'EXPORT'"
}
```

`EFFECT_DENY` wins over the allow, so every activity works except export. `activity.action == 'EXPORT'` covers both the encryption-key and static-deposit exports.

```bash
export TURNKEY_ORG_ID=<org/sub-org id>
export TURNKEY_API_PUBLIC_KEY=<non-root user public key>
export TURNKEY_API_PRIVATE_KEY=<non-root user private key>

cargo test -p breez-sdk-itest --features turnkey --test signer_backends -- --ignored --nocapture \
  turnkey_no_export_gates_onchain_receive \
  turnkey_export_denied_with_encryption_enabled_fails \
  turnkey_export_denied_with_encryption_enabled_fails_server_mode
```

### DIFF
--- a/crates/breez-sdk/core/src/models/mod.rs
+++ b/crates/breez-sdk/core/src/models/mod.rs
@@ -677,14 +677,13 @@ pub struct Config {
     /// caller. Cross-chain sends are only supported on mainnet.
     pub cross_chain_config: Option<CrossChainConfig>,
 
-    /// Whether the configured signer can export private keys. Defaults to
-    /// `true`.
+    /// Whether the signer can perform the SDK's local ECIES/HMAC operations.
+    /// Defaults to `true`.
     ///
-    /// Set to `false` when connecting an external signer whose policy forbids
-    /// exporting private keys. The SDK seeds its local encryption from an
-    /// exported key, and claiming or refunding an on-chain deposits also
-    /// requires exporting a key.
-    pub signer_can_export_keys: bool,
+    /// Set to `false` for a signer that can't run them locally (for example a
+    /// policy-restricted enclave). The SDK then keeps session tokens in plaintext
+    /// and disables the features that need local encryption.
+    pub signer_supports_ecies_hmac: bool,
 }
 
 /// Configuration for cross-chain sends.
--- a/crates/breez-sdk/core/src/models/mod.rs
+++ b/crates/breez-sdk/core/src/models/mod.rs
@@ -676,6 +676,15 @@ pub struct Config {
     /// run background work (e.g. web sockets), so enabling is left to the
     /// caller. Cross-chain sends are only supported on mainnet.
     pub cross_chain_config: Option<CrossChainConfig>,
+
+    /// Whether the configured signer can export private keys. Defaults to
+    /// `true`.
+    ///
+    /// Set to `false` when connecting an external signer whose policy forbids
+    /// exporting private keys. The SDK seeds its local encryption from an
+    /// exported key, and claiming or refunding an on-chain deposits also
+    /// requires exporting a key.
+    pub signer_can_export_keys: bool,
 }
 
 /// Configuration for cross-chain sends.
--- a/crates/breez-sdk/core/src/models/mod.rs
+++ b/crates/breez-sdk/core/src/models/mod.rs
@@ -96,6 +96,11 @@ pub struct ConnectWithSignerRequest {
     pub breez_signer: std::sync::Arc<dyn crate::signer::ExternalBreezSigner>,
     /// External high-level Spark signer for the Spark wallet flows.
     pub spark_signer: std::sync::Arc<dyn crate::signer::ExternalSparkSigner>,
+    /// Whether the signer can perform the SDK's local ECIES/HMAC operations.
+    /// Set `false` for a signer that can't run them locally (for example a
+    /// policy-restricted enclave): the SDK then keeps session tokens in plaintext
+    /// and disables the features that rely on ECIES/HMAC.
+    pub supports_ecies_hmac: bool,
     pub storage_dir: String,
 }
 
@@ -676,14 +681,6 @@ pub struct Config {
     /// run background work (e.g. web sockets), so enabling is left to the
     /// caller. Cross-chain sends are only supported on mainnet.
     pub cross_chain_config: Option<CrossChainConfig>,
-
-    /// Whether the signer can perform the SDK's local ECIES/HMAC operations.
-    /// Defaults to `true`.
-    ///
-    /// Set to `false` for a signer that can't run them locally (for example a
-    /// policy-restricted enclave). The SDK then keeps session tokens in plaintext
-    /// and disables the features that need local encryption.
-    pub signer_supports_ecies_hmac: bool,
 }
 
 /// Configuration for cross-chain sends.
--- a/crates/breez-sdk/core/src/models/mod.rs
+++ b/crates/breez-sdk/core/src/models/mod.rs
@@ -96,11 +96,24 @@ pub struct ConnectWithSignerRequest {
     pub breez_signer: std::sync::Arc<dyn crate::signer::ExternalBreezSigner>,
     /// External high-level Spark signer for the Spark wallet flows.
     pub spark_signer: std::sync::Arc<dyn crate::signer::ExternalSparkSigner>,
-    /// Whether the signer can perform the SDK's local ECIES/HMAC operations.
-    /// Set `false` for a signer that can't run them locally (for example a
-    /// policy-restricted enclave): the SDK then keeps session tokens in plaintext
-    /// and disables the features that rely on ECIES/HMAC.
-    pub supports_ecies_hmac: bool,
+    pub storage_dir: String,
+}
+
+/// Request object for connecting to the Spark network using a signing-only
+/// external signer.
+///
+/// Use this instead of [`ConnectWithSignerRequest`] for a signer that can't
+/// perform the SDK's local ECIES/HMAC operations (for example a
+/// policy-restricted enclave). The SDK keeps session tokens in plaintext and
+/// disables the features that rely on ECIES/HMAC.
+#[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct ConnectWithSigningOnlySignerRequest {
+    pub config: Config,
+    /// Signing-only external signer for non-Spark SDK signing.
+    pub breez_signer: std::sync::Arc<dyn crate::signer::ExternalSigningSigner>,
+    /// External high-level Spark signer for the Spark wallet flows.
+    pub spark_signer: std::sync::Arc<dyn crate::signer::ExternalSparkSigner>,
     pub storage_dir: String,
 }
 
--- a/crates/breez-sdk/core/src/sdk/helpers.rs
+++ b/crates/breez-sdk/core/src/sdk/helpers.rs
@@ -183,15 +183,14 @@ pub(crate) fn validate_breez_api_key(api_key: &str) -> Result<(), SdkError> {
 pub(crate) async fn get_deposit_address(
     spark_wallet: &SparkWallet,
     new_address: bool,
+    signer_can_export_keys: bool,
 ) -> Result<String, SdkError> {
     // Never hand out a static deposit address the signer cannot later claim or
     // refund: both require exporting the static-deposit key, so if that export
     // is denied, funds sent here would be stuck.
-    if !spark_wallet.static_deposit_export_available().await? {
-        return Err(SdkError::Signer(
-            "On-chain deposits cannot be claimed or refunded with the current signer, \
-             so no deposit address can be issued"
-                .to_string(),
+    if !signer_can_export_keys {
+        return Err(SdkError::SignerKeyExportUnavailable(
+            "On-chain addresses cannot be issued with the current signer".to_string(),
         ));
     }
     if new_address {
--- a/crates/breez-sdk/core/src/sdk/helpers.rs
+++ b/crates/breez-sdk/core/src/sdk/helpers.rs
@@ -184,6 +184,16 @@ pub(crate) async fn get_deposit_address(
     spark_wallet: &SparkWallet,
     new_address: bool,
 ) -> Result<String, SdkError> {
+    // Never hand out a static deposit address the signer cannot later claim or
+    // refund: both require exporting the static-deposit key, so if that export
+    // is denied, funds sent here would be stuck.
+    if !spark_wallet.static_deposit_export_available().await? {
+        return Err(SdkError::Signer(
+            "On-chain deposits cannot be claimed or refunded with the current signer, \
+             so no deposit address can be issued"
+                .to_string(),
+        ));
+    }
     if new_address {
         Ok(spark_wallet
             .rotate_static_deposit_address()
--- a/crates/breez-sdk/core/src/sdk/helpers.rs
+++ b/crates/breez-sdk/core/src/sdk/helpers.rs
@@ -183,16 +183,7 @@ pub(crate) fn validate_breez_api_key(api_key: &str) -> Result<(), SdkError> {
 pub(crate) async fn get_deposit_address(
     spark_wallet: &SparkWallet,
     new_address: bool,
-    signer_can_export_keys: bool,
 ) -> Result<String, SdkError> {
-    // Never hand out a static deposit address the signer cannot later claim or
-    // refund: both require exporting the static-deposit key, so if that export
-    // is denied, funds sent here would be stuck.
-    if !signer_can_export_keys {
-        return Err(SdkError::SignerKeyExportUnavailable(
-            "On-chain addresses cannot be issued with the current signer".to_string(),
-        ));
-    }
     if new_address {
         Ok(spark_wallet
             .rotate_static_deposit_address()
--- a/crates/breez-sdk/core/src/sdk/lnurl/mod.rs
+++ b/crates/breez-sdk/core/src/sdk/lnurl/mod.rs
@@ -147,14 +147,6 @@ impl BreezSdk {
         &self,
         request_data: LnurlAuthRequestDetails,
     ) -> Result<LnurlCallbackStatus, SdkError> {
-        // LNURL-auth derives the domain linking key via HMAC, which needs the
-        // local encryption key. Fail clearly up front when the signer cannot
-        // export the key that seeds it.
-        if !self.config.signer_can_export_keys {
-            return Err(SdkError::SignerKeyExportUnavailable(
-                "LNURL-auth key cannot be derived with the current signer".to_string(),
-            ));
-        }
         let request: breez_sdk_common::lnurl::auth::LnurlAuthRequestDetails = request_data.into();
         let status = breez_sdk_common::lnurl::auth::perform_lnurl_auth(
             self.lnurl_client.as_ref(),
--- a/crates/breez-sdk/core/src/sdk/lnurl/mod.rs
+++ b/crates/breez-sdk/core/src/sdk/lnurl/mod.rs
@@ -147,6 +147,14 @@ impl BreezSdk {
         &self,
         request_data: LnurlAuthRequestDetails,
     ) -> Result<LnurlCallbackStatus, SdkError> {
+        // LNURL-auth derives the domain linking key via HMAC, which needs the
+        // local encryption key. Fail clearly up front when it is unavailable
+        // (e.g. a Turnkey policy denied the encryption-key export).
+        if !self.encryption_available {
+            return Err(SdkError::Signer(
+                "LNURL-auth requires encryption that the current signer cannot provide".to_string(),
+            ));
+        }
         let request: breez_sdk_common::lnurl::auth::LnurlAuthRequestDetails = request_data.into();
         let status = breez_sdk_common::lnurl::auth::perform_lnurl_auth(
             self.lnurl_client.as_ref(),
--- a/crates/breez-sdk/core/src/sdk/lnurl/mod.rs
+++ b/crates/breez-sdk/core/src/sdk/lnurl/mod.rs
@@ -147,11 +147,17 @@ impl BreezSdk {
         &self,
         request_data: LnurlAuthRequestDetails,
     ) -> Result<LnurlCallbackStatus, SdkError> {
+        // LNURL-auth needs the HMAC step, absent on a signing-only signer.
+        let Some(lnurl_auth_signer) = self.lnurl_auth_signer.as_ref() else {
+            return Err(SdkError::Generic(
+                "LNURL-auth requires a signer that supports HMAC".to_string(),
+            ));
+        };
         let request: breez_sdk_common::lnurl::auth::LnurlAuthRequestDetails = request_data.into();
         let status = breez_sdk_common::lnurl::auth::perform_lnurl_auth(
             self.lnurl_client.as_ref(),
             &request,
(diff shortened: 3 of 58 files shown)


roeierez (approved):
LGTM

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -759,21 +791,31 @@ fn finalize_spark_wallet_config(
...
+    let store: Arc<dyn SessionStore> = if encryption_available {
+        Arc::new(
+            crate::session_store::EncryptingSessionStore::new(inner, signer.clone(), network)
+                .map_err(|e| {
+                    SdkError::Generic(format!("failed to set up session token encryption: {e}"))
+                })?,
+        )
+    } else {
+        inner
+    };
```
roeierez: I am thinking maybe we remove this session encryption all together and make it the integrator responsibility?
If not we should make sure the sdk handles fracefully switching between encryption modes such as if an encrypted session is stored and then signer doesn't support encryption.

crates/breez-sdk/core/src/turnkey/spark_signer.rs
```diff
@@ -479,18 +484,47 @@ impl TurnkeySparkSigner {
             return Ok(secret);
         }
         let path = format!("{}/3'/{index}'", self.base_path());
-        let secret = self
+        let secret = match self
```
roeierez: This request on connect is only for detecting if encryption is supported or not right?
In server mode (where sync is disabled) do we even need encryption? 
My point is that perhaps we can defer the whole encryption detection to the point it is needed like we do in deposit? Not introduced by this PR but currently every connect starts with turnkey roundtrip before even starting to execute the request which will add unnecessary latency.

roeierez (changes requested):

crates/breez-sdk/core/src/sdk/lnurl/mod.rs
```diff
@@ -147,6 +147,14 @@ impl BreezSdk {
...
         request_data: LnurlAuthRequestDetails,
     ) -> Result<LnurlCallbackStatus, SdkError> {
+        // LNURL-auth derives the domain linking key via HMAC, which needs the
+        // local encryption key. Fail clearly up front when it is unavailable
+        // (e.g. a Turnkey policy denied the encryption-key export).
+        if !self.encryption_available {
+            return Err(SdkError::Signer(
+                "LNURL-auth requires encryption that the current signer cannot provide".to_string(),
+            ));
+        }
```
JssDWt: Some Turnkey specifics bleeding in here I think?

crates/breez-sdk/core/src/sdk/helpers.rs
```diff
@@ -184,6 +184,16 @@ pub(crate) async fn get_deposit_address(
...
+    // Never hand out a static deposit address the signer cannot later claim or
+    // refund: both require exporting the static-deposit key, so if that export
+    // is denied, funds sent here would be stuck.
+    if !spark_wallet.static_deposit_export_available().await? {
+        return Err(SdkError::Signer(
+            "On-chain deposits cannot be claimed or refunded with the current signer, \
+             so no deposit address can be issued"
+                .to_string(),
+        ));
+    }
```
JssDWt: I wonder whether it's necessary to have this `static_deposit_export_available` at all? Can't we just let the turnkey signer error with the 403?

crates/breez-sdk/core/src/signer/external.rs
```diff
@@ -116,4 +116,9 @@ pub trait ExternalBreezSigner: Send + Sync {
         message: Vec<u8>,
         path: String,
     ) -> Result<HashedMessageBytes, SignerError>;
+
+    /// Whether the signer can perform local ECIES/HMAC.
+    async fn encryption_available(&self) -> bool {
```
JssDWt: What if:
- We add a flag to the config in the lines of `encrypt_session_storage: Option<bool>`. Defaults to `true`. If set to `false`, we don't encrypt.
- If you try to do LNURL-auth with a signer that doesn't support it, we just error from the signer.
- If you try to create a static deposit address we just error from the signer.

We can drop the `encryption_available` flag and the `static_deposit_export_available` flag then. I think the claiming/refunding will still work. If your signer doesn't support it you'll get an error from the signer. (Note it doesn't make sense to have a static deposit at all if you couldn't create the address in the first place)

crates/breez-sdk/core/src/turnkey/factory.rs
```diff
@@ -63,12 +64,26 @@ pub async fn create_turnkey_signer(config: TurnkeyConfig) -> Result<ExternalSign
...
+        Err(TurnkeyError::Http { status: 403, .. }) => {
+            tracing::warn!(
+                "Turnkey denied the encryption-key export (403): running without local \
+                 encryption (unencrypted session store; features that depend on local \
+                 ECIES/HMAC are unavailable)"
+            );
+            None
+        }
+        Err(e) => return Err(to_signer_err(e)),
+    };
```
JssDWt: I think this should be explicitly requested by the user, not magically resolved.

JssDWt (commented):
I think we can simplify this a great deal if:
- we let the user explicitly signal wallet export support on creating the turnkey signer
- we let the user explicitly disable encryption in the sdk config

The turnkey breez signer can then have the optional encryption enabled based on the user setting. As long as you don't create static deposits and do LNURL-auth, everything should work out of the box then without any signaling flags.

crates/spark/src/header_provider.rs
```diff
@@ -13,6 +14,12 @@ pub enum HeaderProviderError {
 #[macros::async_trait]
 pub trait HeaderProvider: Send + Sync {
     async fn headers(&self) -> Result<HashMap<String, String>, HeaderProviderError>;
+
+    /// Force a fresh authentication on the next request, discarding any cached
+    /// session. Default no-op for providers without a session.
+    async fn reauthenticate(&self) -> Result<(), HeaderProviderError> {
+        Ok(())
+    }
```
JssDWt: The header provider was so clean! Single surpose! It was not necessarily about auth. Like the jwt is also in a header provider. Those are not about auth. They just provide headers. Obviously it makes sense to re-auth when your auth header gets rejected, but we have to find another way if we do.

I wonder whether we can do one of:
- parse the token on get_session to see whether it is a valid token? (they are base64 for example, if encrypted they are not base64) Then get_session could return NotFound if the encrypting provider finds an encrypted token when encryption is off, or vice versa.
- if we really really want to re-auth, pass response status/headers to the header provider and have the header provider decide that it needs to re-auth?

dangeross: Agreed, it complicated it. I've removed the auth recovery and added this NotFound decryption fallback in the EncryptingSessionStore. 
Because of the way session stores are wrapped `CachingSessionStore(EncryptingSessionStore(SessionStore))`:
1. the SessionStore shouldn't invalidate "encrypted data"
2. even if it could, both encrypted and plaintext tokens are base64 encoded bytes

~So it's currently asymmetric: encryption-on immediately forces a re-auth by returning NotFound, encryption-off only re-auths after expiry.~
I've opted for the EncryptingSessionStore to mark the token if it is encrypted or not, then validating based on if ECIES/HMAC is enabled. So now switching in both directions is symmetric and forces a re-auth.

crates/breez-sdk/core/src/turnkey/spark_signer.rs
```diff
@@ -525,6 +525,9 @@ impl ExternalSparkSigner for TurnkeySparkSigner {
         &self,
         index: u32,
     ) -> Result<PublicKeyBytes, SignerError> {
+        // Fail here if the key can't be exported (a deny-export policy), so an
+        // address that could never be claimed or refunded is never issued.
+        self.export_static_deposit_key(index).await?;
         Ok(PublicKeyBytes::from_public_key(
             &self.static_deposit_public_key(index).await?,
         ))
```
danielgranhao: Here `static_deposit_public_key` makes a separate turnkey request when the one from `export_static_deposit_key` should be enough. We can get the pubkey from the exported secret key

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -676,6 +676,14 @@ pub struct Config {
...
     /// caller. Cross-chain sends are only supported on mainnet.
     pub cross_chain_config: Option<CrossChainConfig>,
+
+    /// Whether the signer can perform the SDK's local ECIES/HMAC operations.
+    /// Defaults to `true`.
+    ///
+    /// Set to `false` for a signer that can't run them locally (for example a
+    /// policy-restricted enclave). The SDK then keeps session tokens in plaintext
+    /// and disables the features that need local encryption.
+    pub signer_supports_ecies_hmac: bool,
```
danielgranhao: The name reads as a signer property but lives here in the SDK config. Maybe `enable_signer_ecies_hmac_operations`? Though I'm not sure if this is a good place to put this in the first place. 

What if we move this to `ConnectWithSignerRequest` and carry it on `SignerSource::External { breez, spark, supports_ecies_hmac }`? Then `wrap_session_store` and `validate_signer_capabilities` read it from the signer source instead of config. In this case we would keep name.

Pros would be:
- Kills the seed-path misconfiguration entirely. A seed signer always supports ECIES
- Co-locates the declaration with the signer

danielgranhao (commented):
Looking good

crates/breez-sdk/core/src/models/mod.rs
dangeross: Nice

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -96,6 +96,11 @@ pub struct ConnectWithSignerRequest {
     pub breez_signer: std::sync::Arc<dyn crate::signer::ExternalBreezSigner>,
     /// External high-level Spark signer for the Spark wallet flows.
     pub spark_signer: std::sync::Arc<dyn crate::signer::ExternalSparkSigner>,
+    /// Whether the signer can perform the SDK's local ECIES/HMAC operations.
+    /// Set `false` for a signer that can't run them locally (for example a
+    /// policy-restricted enclave): the SDK then keeps session tokens in plaintext
+    /// and disables the features that rely on ECIES/HMAC.
+    pub supports_ecies_hmac: bool,
```
JssDWt: This is quite good There's two things I'm still not 100% happy with:
- The coupling of ecies and hmac (they are only related because turnkey has to export a key for both of them)
- The fact that ecies and hmac are part of the signer trait, but they are excluded by this boolean

What if we would split ecies and hmac from the breez signer trait? Here's a suggestion:

External api:
```
// signer/external.rs — UNCHANGED. The full signer: ONE trait, all 7 methods.
// Every current impl (Turnkey, DefaultExternalSigner, the JS object) keeps compiling.
#[cfg_attr(feature = "uniffi", uniffi::export(with_foreign, async_runtime = "tokio"))]
#[macros::async_trait]
pub trait ExternalBreezSigner: Send + Sync {
    async fn derive_public_key(&self, path: String) -> Result<PublicKeyBytes, SignerError>;
    async fn sign_ecdsa(&self, m: MessageBytes, path: String) -> Result<EcdsaSignatureBytes, SignerError>;
    async fn sign_ecdsa_recoverable(&self, m: MessageBytes, path: String) -> Result<RecoverableEcdsaSignatureBytes, SignerError>;
    async fn sign_hash_schnorr(&self, hash: Vec<u8>, path: String) -> Result<SchnorrSignatureBytes, SignerError>;
    async fn encrypt_ecies(&self, m: Vec<u8>, path: String) -> Result<Vec<u8>, SignerError>;
    async fn decrypt_ecies(&self, m: Vec<u8>, path: String) -> Result<Vec<u8>, SignerError>;
    async fn hmac_sha256(&self, m: Vec<u8>, path: String) -> Result<HashedMessageBytes, SignerError>;
}

// NEW — the opt-out profile. Standalone; NO `: ExternalBreezSigner` supertrait
// (UniFFI can't do foreign inheritance). Implementing THIS is how you declare
// "I can't encrypt" — the capability is now in the type, no bool anywhere.
#[cfg_attr(feature = "uniffi", uniffi::export(with_foreign, async_runtime = "tokio"))]
#[macros::async_trait]
pub trait ExternalSigningSigner: Send + Sync {
    async fn derive_public_key(&self, path: String) -> Result<PublicKeyBytes, SignerError>;
    async fn sign_ecdsa(&self, m: MessageBytes, path: String) -> Result<EcdsaSignatureBytes, SignerError>;
    async fn sign_ecdsa_recoverable(&self, m: MessageBytes, path: String) -> Result<RecoverableEcdsaSignatureBytes, SignerError>;
    async fn sign_hash_schnorr(&self, hash: Vec<u8>, path: String) -> Result<SchnorrSignatureBytes, SignerError>;
}
```

Internal:
```
// signer/mod.rs
pub trait BreezSigner: Send + Sync {   // signing — always present
    async fn derive_public_key(&self, p: &DerivationPath) -> Result<PublicKey, SdkError>;
    async fn sign_ecdsa(&self, m: Message, p: &DerivationPath) -> Result<Signature, SdkError>;
    async fn sign_ecdsa_recoverable(&self, m: Message, p: &DerivationPath) -> Result<RecoverableSignature, SdkError>;
    async fn sign_hash_schnorr(&self, h: &[u8], p: &DerivationPath) -> Result<schnorr::Signature, SdkError>;
}
pub trait EciesSigner: Send + Sync {   // optional axis 1
    async fn encrypt_ecies(&self, m: &[u8], p: &DerivationPath) -> Result<Vec<u8>, SdkError>;
    async fn decrypt_ecies(&self, m: &[u8], p: &DerivationPath) -> Result<Vec<u8>, SdkError>;
}
pub trait HmacSigner: Send + Sync {    // optional axis 2 — independent of ECIES
    async fn hmac_sha256(&self, key_path: &DerivationPath, input: &[u8]) -> Result<Hmac<sha256::Hash>, SdkError>;
}
```

Builder:
```
enum ExternalBreez {                                  // pure Rust, lives in sdk_builder.rs
    Full(Arc<dyn ExternalBreezSigner>),
    SigningOnly(Arc<dyn ExternalSigningSigner>),
}

struct Signers {
    base:  Arc<dyn BreezSigner>,
    ecies: Option<Arc<dyn EciesSigner>>,   // ← the two axes, independent
    hmac:  Option<Arc<dyn HmacSigner>>,    //   even though today's profiles set both-or-neither
    spark: Arc<dyn SparkSigner>,
    rtsync:     Option<Arc<RTSyncSigner>>,
    lnurl_auth: Option<Arc<LnurlAuthSignerAdapter>>,
}

// build_signers — presence replaces validate_signer_capabilities + the encrypt bool:
let (base, ecies, hmac) = match breez {
    ExternalBreez::Full(e) => {
        let a = Arc::new(FullAdapter::new(e));                    // impls all three internal traits
        (a.clone() as _, Some(a.clone() as _), Some(a as _))
    }
    ExternalBreez::SigningOnly(e) => (Arc::new(SigningAdapter::new(e)) as _, None, None),
};
// Seed source → all three Some (BreezSignerImpl already does everything; split its 7 impls mechanically).

// rtsync/cross_chain need `ecies` → build them iff ecies.is_some(); hard-error at build if the
//   feature is configured but ecies is None (keep #970's fail-loud, now a presence check).
// lnurl_auth needs `hmac` → Some(adapter) iff hmac.is_some(); else lnurl_auth() returns a typed error.
// session store: ecies.is_some() → encrypt; else plaintext. Keep #970's pln:/enc: mode tag.
```
```
impl SdkBuilder {
    pub fn new_with_signer(config: Config, breez: Arc<dyn ExternalBreezSigner>, spark: …) -> Self { /* Full */ }
    pub fn new_with_signing_only_signer(config: Config, breez: Arc<dyn ExternalSigningSigner>, spark: …) -> Self { /* SigningOnly */ }
}
```

docs/breez-sdk/src/guide/external_signer.md
```diff
@@ -29,6 +29,10 @@ If you need full control over the signing process, you can implement the [Extern
 
 The default implementations of the two interfaces, [DefaultExternalSigner](https://github.com/breez/spark-sdk/blob/main/crates/breez-sdk/core/src/signer/default_external.rs) and [DefaultExternalSparkSigner](https://github.com/breez/spark-sdk/blob/main/crates/breez-sdk/core/src/signer/default_external_spark.rs), can be used as a reference for what's expected.
 
+### Signers Without Local ECIES/HMAC Support
+
+Some external signers can't perform the SDK's local ECIES/HMAC operations (for example, a policy-restricted enclave that won't release key material). For these, implement {{#name ExternalSigningSigner}} instead of {{#name ExternalBreezSigner}}, then connect with {{#name connect_with_signing_only_signer}}. The SDK keeps session tokens in plaintext and disables the features that rely on those operations.
```
danielgranhao: Do you think it's worth mentioning that when implementing `ExternalSigningSigner`, rtsync must be disabled in config? Build will fail otherwise

crates/breez-sdk/core/src/turnkey/spark_signer.rs
```diff
@@ -525,9 +525,12 @@ impl ExternalSparkSigner for TurnkeySparkSigner {
...
         index: u32,
     ) -> Result<PublicKeyBytes, SignerError> {
-        Ok(PublicKeyBytes::from_public_key(
-            &self.static_deposit_public_key(index).await?,
-        ))
+        // Export the static-deposit key up front so a signer that can't export
+        // it (a deny-export policy) fails here, not later. The pubkey is derived
+        // from the exported secret, so there's no second Turnkey round-trip.
+        let secret = self.export_static_deposit_key(index).await?;
+        let public_key = PublicKey::from_secret_key(&Secp256k1::new(), &secret);
```
danielgranhao: Minor optimization: might be worth moving to a global or shared secp context to avoid the repeated allocation

danielgranhao (commented):
Just 2 minor comments. Otherwise LGTM

docs/breez-sdk/src/guide/external_signer.md
JssDWt: Should we list the features that are unavailable when using this?
- real-time sync
- LNURL-auth
- cross-chain
- not encrypting session tokens I think is worth a proper warning? !important!

JssDWt (commented):
Looking good!

danielgranhao (approved):
LGTM

crates/breez-sdk/core/src/signer/mod.rs
```diff
@@ -1,8 +1,15 @@
+use std::sync::LazyLock;
+
 use crate::SdkError;
 use bitcoin::bip32::DerivationPath;
 use bitcoin::hashes::{Hmac, sha256};
 use bitcoin::secp256k1::{self, Message, ecdsa::RecoverableSignature};
 
+/// Shared secp256k1 context for the SDK-layer signers. `Secp256k1::new()`
+/// allocates a large precomputation table, so it's built once for the process.
+pub(crate) static SECP: LazyLock<secp256k1::Secp256k1<secp256k1::All>> =
```
JssDWt: I think this is a bit much! A field on the turnkey signer will do.

JssDWt (approved):
LGTM. Nit on the Secp256k1 initialization

PR #972, opened 2026-06-26
Client signing send

roeierez:
**Problem**

When the SDK runs on a server with a Turnkey signer, the key that authorizes a payment belongs to the end user on their device, not the server. Today send_payment does all the signing inline in a single call, so there's no way for the server to pause, get the user to approve the payment, and then continue. This PR adds support for that, across every kind of send: spark, lightning, tokens, and on-chain (cooperative exit).

**Approach**

We split a send so the one piece that needs the user's key is produced on its own and signed off-device:

- **Prepare (server):** prepare_send_payment describes the payment and its fees. It does no signing and produces nothing to approve.
- **Build the approval (server):** build_unsigned_transfer_package returns the single item the user needs to sign.
- **Approve (user's device):** the front-end has the user sign that item.
- **Submit (server):** publish_signed_transfer_package takes the signed item and completes the payment.

Everything needed to finish the payment travels inside the request/response, so the server keeps no state in between: no memory, no database. That means the approval can happen out-of-process and even survive a restart or land on a different instance.

This is fully opt-in: a normal send_payment (no client signing) behaves exactly as before.

For on-chain (cooperative exit) sends, the user picks the confirmation speed when building the approval: the fee, and therefore the exact thing being signed, depends on it.

**The front-end loop:**

```ts
const prep = await sdk.prepareSendPayment({ paymentRequest, amount })

while (true) {
  const unsigned = await sdk.buildUnsignedTransferPackage({ prepareResponse: prep, options }) // options: speed for on-chain

  // The package self-describes the payment, so the approver (even a remote signer
  // that only receives this package) can show what's being signed.
  showApproval(describe(unsigned))

  const signature = await approveOnDevice(unsigned)  // user signs on their device
  const result = await sdk.publishSignedTransferPackage({ signedPackage: { unsigned, signature } })

  if (result.type === "swapCompleted") continue
  break                                           // result.payment is the sent payment
}
```

**API**

New methods:

- `build_unsigned_transfer_package(BuildUnsignedTransferPackageRequest) -> UnsignedTransferPackage`: builds the single item to approve from a prepare response. No signing, no side effects. The user should sign the result.
- `publish_signed_transfer_package(PublishSignedTransferPackageRequest { prepare_response, signed_package }) -> PublishSignedTransferPackageResponse`: completes the approved step. Returns `PaymentSent { payment }` for a Transfer or Token, or `SwapCompleted` when a denomination swap ran (re-build and submit again).

prepare_send_payment and send_payment keep their existing shape: client signing adds methods rather than changing the normal send.

**Denomination swaps**

Sometimes the wallet can't send the exact amount with the funds it has and must first re-shape them (a "denomination swap"). That swap also needs the user's approval, so it can't be done silently on the server.

Rather than hiding it inside prepare (which would give prepare surprising side effects), the swap is just another package the user approves and submits. publish_signed_transfer_package reports back SwapCompleted, and the integrator keeps approving and submitting until the payment is sent. prepare stays read-only.


crates/spark/src/tree/tests.rs
```diff
@@ -282,6 +282,60 @@ pub async fn test_reserve_leaves(store: &dyn TreeStore) {
     assert!(!reservation.id.is_empty());
 }
 
+pub async fn test_reserve_leaves_by_ids(store: &dyn TreeStore) {
```
dangeross, line 285: `try_select_leaves` doesn't have the shared-suite coverage `test_reserve_leaves_by_ids` has here, so the Postgres/MySQL impls run untested. Worth the same shared test + harness wiring? (Same for `select_token_outputs` / `reserve_token_outputs_by_outpoints` in `token/tests.rs`.)

crates/breez-sdk/wasm/src/models/mod.rs
```diff
@@ -1040,6 +1040,73 @@ pub struct LnurlWithdrawResponse {
...
+        amount_sat: u64,
+        fee_sat: u64,
+        target: TransferTarget,
+    },
+    Token {
+        prepare_token_transaction: crate::signer::ExternalPrepareTokenTransactionRequest,
+        token_context: Vec<u8>,
+        token_identifier: String,
+        amount: u128,
+        fee: u128,
```
dangeross, line 1084: `amount`/`fee` are bare `u128`, so they cross to JS as `bigint` and `JSON.stringify`-ing the package throws (`deserialize_u128` then rejects a number/string back). A Token package can't JSON round-trip the way the stateless flow needs. Every other wire `u128` here stringifies (`SparkAddress.fee`, `CrossChainAddress`, `ConversionSide`); the itests miss it since they never serialize.

```suggestion
        #[tsify(type = "string")]
        #[serde(with = "serde_u128_as_string")]
        amount: u128,
        #[tsify(type = "string")]
        #[serde(with = "serde_u128_as_string")]
        fee: u128,
```

crates/breez-sdk/wasm/js/postgres-tree-store/index.cjs
```diff
@@ -594,6 +594,105 @@ class PostgresTreeStore {
...
+        id: r.id,
+        value: Number(r.value),
+      }));
+
+      const selected = this._selectLeavesByTargetAmounts(slimLeaves, targetAmounts);
+      if (selected !== null && selected.length > 0) {
+        const fullLeaves = await this._fetchFullLeavesByIds(
+          this.pool,
+          selected.map((l) => l.id)
+        );
```
dangeross: Rust `try_select_leaves` runs the slim select + resolve in one transaction and errors if the resolve returns fewer leaves than selected. Here the resolve is a separate `this.pool` call and `_fetchFullLeavesByIds` has no count check, so a concurrent `set_leaves` deleting a selected leaf silently returns an under-funded `exact` set, and the user signs a package whose leaves no longer cover the amount. One transaction + a count check like the Rust path? (Same in the mysql store.)

crates/breez-sdk/wasm/src/models/mod.rs
roeierez: Done

crates/breez-sdk/wasm/js/postgres-tree-store/index.cjs
roeierez: Done

crates/spark/src/tree/tests.rs
roeierez: Added to the shared tests.

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -1439,6 +1439,25 @@ pub struct LnurlPayResponse {
...
 }
 
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct BuildUnsignedLnurlPayPackageRequest {
+    pub prepare_response: PrepareLnurlPayResponse,
+}
+
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct PublishSignedLnurlPayPackageRequest {
+    pub prepare_response: PrepareLnurlPayResponse,
```
dangeross: Is it possible to mirror the `PublishSignedTransferPackageRequest` and instead include only needed fields for the publish from the `prepare_response` somewhere in the unsigned transfer package? (`pay_request` / `comment` / `success_action`)

crates/breez-sdk/core/src/sdk/lnurl/mod.rs
```diff
@@ -27,6 +29,21 @@ impl BreezSdk {
         pay::send(self, request).await
     }
 
+    pub async fn build_unsigned_lnurl_pay_package(
```
dangeross, line 32: These aren't yet mirrored in the wasm / flutter bindings

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -1588,7 +1680,7 @@ pub struct PrepareSendPaymentRequest {
     pub fee_policy: Option<FeePolicy>,
 }
 
-#[derive(Debug, Clone, Serialize)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
dangeross: `prepare_response` was dropped from the publish request, so this `Deserialize` (plus the ones added with it) should no longer be needed right? (still needed on `SendOnchainFeeQuote` and `SendOnchainSpeedFeeQuote` for `TransferTarget::CoopExit`)

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -2146,3 +2251,31 @@ pub struct UnregisterWebhookRequest {
     /// The unique identifier of the webhook to unregister.
     pub webhook_id: String,
 }
+
+#[cfg(test)]
+mod client_signing_tests {
+    use super::*;
+
+    #[test]
+    fn prepare_response_json_round_trips() {
```
dangeross: Is this test still needed now the round-trip with the `prepare_response` no longer happens?

crates/spark/src/tree/mod.rs
```diff
@@ -357,12 +357,15 @@ pub struct SelectLeavesOptions {
     ///
     /// Default: 60 seconds
     pub max_wait_for_pending: Duration,
+
+    pub allow_swap: bool,
```
JssDWt: I believe this field is never set to `false`, so it can be removed?

crates/spark-wallet/src/wallet/external_signing.rs
```diff
@@ -0,0 +1,365 @@
...
+        approved_transfer: PreparedTransfer,
+    ) -> Result<WalletTransfer, SparkWalletError> {
+        let reservation = self
+            .tree_service
+            .reserve_leaves_by_ids(&leaf_ids, ReservationPurpose::Payment)
+            .await?;
+
+        let transfer = with_reserved_leaves(
+            self.tree_service.as_ref(),
+            self.transfer_service.submit_transfer_with_prepared(
```
JssDWt, line 101: There appear to be some differences between the old and new spark transfer (transfer_leaves_to vs submit_transfer_with_prepared) and lightning (pay_lightning_invoice vs submit_lightning_send) send paths. The main thing I'm spotting now is that the payment observer is not invoked for this path. So the code for the sends is duplicated. It would be great if we could fit the 'normal' send path in this new build/publish send path too, so code is reused, and we won't accidentally update only one of these paths in the future.

crates/breez-sdk/core/src/sdk/payments/mod.rs
```diff
@@ -88,6 +91,35 @@ impl BreezSdk {
         Box::pin(send::orchestrate_send(self, request, false, None)).await
     }
 
+    pub async fn build_unsigned_transfer_package(
```
JssDWt, line 94: I'm OK with this name, but another suggestion:

`build_unsigned_payment_step`
and
`publish_signed_payment_step`

crates/spark-mysql/src/tree_store.rs
```diff
@@ -515,6 +515,106 @@ impl TreeStore for MysqlTreeStore {
...
+        target_amounts: Option<&TargetAmounts>,
+    ) -> Result<LeafSelection, TreeServiceError> {
+        let target_amount = target_amounts.map_or(0, TargetAmounts::total_sats);
+        let max_target = Self::slim_max_target(target_amounts);
+        let max_target_signed: i64 = i64::try_from(max_target).unwrap_or(i64::MAX);
+
+        let mut conn = self.pool.get_conn().await.map_err(map_err)?;
+        let mut tx = conn.start_transaction(tx_opts()).await.map_err(map_err)?;
+
+        let slim_rows: Vec<(String, i64)> = tx
```
JssDWt, line 556: The slim_rows query can probably be shared with try_reserve_leaves too (same in other database)

JssDWt (commented):
It is looking pretty good. Maybe later we can do an effort to inverse the position of the signer inside spark-wallet in general. So calls to the signer always go from the top-level, like in this new flow.

crates/breez-sdk/core/src/models/mod.rs
roeierez: Done

crates/breez-sdk/core/src/sdk/lnurl/mod.rs
roeierez: Done

crates/breez-sdk/core/src/models/mod.rs
roeierez: Removed

crates/breez-sdk/core/src/models/mod.rs
roeierez: Removed

crates/spark/src/tree/mod.rs
roeierez: Yes, removed.

crates/breez-sdk/core/src/sdk/payments/mod.rs
roeierez: I had a discussion with Roy about this. I think the transfer package is important here and also use in spark sdk in this context. I think it serves well the flow we are implementing here.

crates/spark-mysql/src/tree_store.rs
roeierez: Sure, done.

crates/spark-wallet/src/wallet/external_signing.rs
roeierez: Observer is now invoked on both submit paths via shared helpers.
To fully unify both flows I think we need the signer inversion from your summary comment, that's a large change.
If you agree I would happily look at this in a follow up.

crates/breez-sdk/core/src/sdk/payments/client_signing.rs
```diff
@@ -0,0 +1,285 @@
...
+        .parse::<SparkAddress>()
+        .map_err(|_| SdkError::InvalidInput("Invalid spark address".to_string()))?;
+    let prepared = sdk
+        .spark_wallet
+        .prepare_token_package(
+            vec![TransferTokenOutput {
+                token_id: token_identifier.clone(),
+                amount,
+                receiver_address: spark_address,
+                spark_invoice: None,
```
danielgranhao: Doesn't this break token sends to a spark invoice?

For a token-denominated Spark invoice, `build_unsigned_transfer_package`
routes `SendPaymentMethod::SparkInvoice` into `build_spark_package` with
`receiver = spark_invoice_details.invoice` (the raw invoice string). Because a
`token_identifier` is set, it forwards to `build_token_package`, which parses the
invoice-bearing address and hands it to `prepare_token_package`.

Downstream, `TokenService::build_partial_token_transfer` calls
`receiver_address.to_address_string()`, which returns `Err("Invoice addresses cannot be converted to address strings")` for any address carrying invoice fields causing the build to error out.

Suggested fix: in the token branch, strip the invoice from the address (as the
sats path does) and pass `spark_invoice: Some(invoice)` + an `execute_before`
derived from the invoice expiry.

crates/breez-sdk/core/src/sdk/payments/client_signing.rs
```diff
@@ -0,0 +1,285 @@
...
+    Ok(())
+}
+
+pub(in crate::sdk) async fn build_unsigned_transfer_package(
+    sdk: &BreezSdk,
+    prepare_response: &PrepareSendPaymentResponse,
+    options: Option<&BuildTransferPackageOptions>,
+) -> Result<UnsignedTransferPackage, SdkError> {
+    reject_conversion(prepare_response)?;
+    match &prepare_response.payment_method {
```
danielgranhao, line 76: Fee mode is getting ignored here and works as if it's always `FeesExcluded`. For `FeesIncluded` we'll pay the wrong amount

crates/spark-wallet/src/wallet/external_signing.rs
danielgranhao: Another difference seems to be missing call to `renew_reservation_timelocks`. On normal path we call it after reserving.

roeierez: @JssDWt I decided to go for the spark wallet's flow unifications as part of this PR, so this should be addressed.

crates/breez-sdk/core/src/sdk/payments/client_signing.rs
roeierez: Done

crates/breez-sdk/core/src/sdk/payments/client_signing.rs
roeierez: Done

crates/spark-wallet/src/wallet/external_signing.rs
roeierez: @dangeross added leaves refresh timelocks.

crates/spark-wallet/src/wallet/external_signing.rs
```diff
@@ -0,0 +1,385 @@
...
+            Ok(nodes) => nodes,
+            Err(e) => {
+                let _ = self.tree_service.cancel_reservation(reservation).await;
+                return Err(e.into());
+            }
+        };
+
+        self.tree_service
+            .finalize_reservation(reservation.id.clone(), Some(claimed.as_slice()))
+            .await?;
```
dangeross: I'm just wondering if we could log-and-continue here (and in `publish_coop_exit_package`) instead of propagating the error? The other finalize calls (`with_reserved_leaves` etc.) do that, but I'm not sure what's correct once the send has already committed.

crates/spark/src/token/token_service.rs
```diff
@@ -988,6 +1069,77 @@ impl TokenService {
...
+
+        let token_transaction = with_reserved_token_outputs(
+            self.token_output_service.as_ref(),
+            self.broadcast_token_transfer(identity_public_key, prepared, signature),
+            &reservation,
+        )
+        .await?;
+
+        self.update_token_outputs_for_transaction(&token_transaction, &identity_public_key)
+            .await?;
```
dangeross: I'm wondering if we can just log-and-continue here instead of propagating? The transfer is already broadcast, and the token store should self-heal on the next sync.

crates/spark-wallet/src/wallet/external_signing.rs
```diff
@@ -0,0 +1,385 @@
...
+    }
+
+    pub async fn publish_transfer_package(
+        &self,
+        transfer_id: TransferId,
+        receiver_public_key: PublicKey,
+        leaf_ids: Vec<TreeNodeId>,
+        spark_invoice: Option<String>,
+        approved_transfer: PreparedTransfer,
+    ) -> Result<WalletTransfer, SparkWalletError> {
```
danielgranhao, line 89: Here we don't validate the spark's invoice expiry time, which could have expired during package signing. We could re-run `parse_and_validate_spark_invoice` on the spark_invoice.

crates/breez-sdk/core/src/sdk/payments/client_signing.rs
```diff
@@ -0,0 +1,354 @@
...
+    let coop_fee_quote: CoopExitFeeQuote = fee_quote.clone().into();
+    let fee_sat = coop_fee_quote.fee_sats(&exit_speed);
+    let receiver_sat = if prepare_response.fee_policy == FeePolicy::FeesIncluded {
+        let receiver = amount_sat.saturating_sub(fee_sat);
+        let dust_limit_sats = crate::utils::bitcoin_dust::get_dust_limit_sats(&address.address)?;
+        if receiver < dust_limit_sats {
+            return Err(SdkError::InvalidInput(format!(
+                "Amount is below the minimum of {dust_limit_sats} sats required for this address"
+            )));
+        }
```
danielgranhao: Here we check for dust limit only on FeesIncluded. On fees excluded we should check as well to match the normal path (send/bitcoin_address.rs:49), which checks dust unconditionally

crates/breez-sdk/core/src/sdk/payments/send/mod.rs
```diff
@@ -8,12 +8,130 @@ use crate::{
...
+            "LNURL pay packages must be published with publish_signed_lnurl_pay_package"
+                .to_string(),
+        ));
+    }
+    match publish_signed_package_inner(sdk, signed_package).await? {
+        None => Ok(PublishSignedTransferPackageResponse::SwapCompleted),
+        Some(res) => {
+            sdk.event_emitter
+                .emit(&SdkEvent::from_payment(res.payment.clone()))
+                .await;
```
danielgranhao, line 144: If publishing is retried after a successful attempt, we will emit a duplicate success event because of the idempotency guard in `publish_signed_package_inner` -> `deferred_transfer_send`. Not a critical issue but on `orchestrate_send`, we gate on `!suppress_payment_event`

crates/breez-sdk/core/src/sdk/payments/send/bolt11.rs
```diff
@@ -97,6 +98,37 @@ pub(super) async fn send(
...
+        bolt11.to_string(),
+        Some(amount_to_send),
+        signed.to_prepared_transfer()?,
+    ))
+    .await?;
+
+    let payment = sdk
+        .lightning_sender
+        .payment_from_pay_result(result, u128::from(amount_to_send), 0)
+        .await?;
```
danielgranhao, line 147: Here we record a different amount than `send_payment` for the same `FeesIncluded`payment (`send_signed` passes the receiver amount `amount_to_send`; send passes `prepare_response.amount`). 

 The two flows should agree. Though I think it's the normal path is the one that should change to match this one so can be done as a follow-up PR.

crates/spark/src/token/token_service.rs
```diff
@@ -988,6 +1069,77 @@ impl TokenService {
...
+
+        self.build_partial_token_transfer(
+            identity_public_key,
+            &token_id,
+            selected.outputs,
+            receiver_outputs,
+            execute_before_unix_micros,
+        )
+        .await
```
danielgranhao, line 1176: In the normal path we have a loop on `NeededTooManyOutputs` so that we optimize token outputs to consolidate. On the client signing path we surface that too many outputs error. Do we want to implement a similar approach to leaves where the consolidation is returned step is returned at publish (like SwapCompleted)? Or leave this out for now?

danielgranhao (commented):
Looking good

crates/spark/src/token/token_service.rs
dangeross: Also worth noting there's no way to manually recover here either: `optimize_token_outputs` isn't exposed publicly the way `optimize_leaves` is.

crates/breez-sdk/breez-itest/tests/client_signing_send.rs
```diff
@@ -0,0 +1,825 @@
...
+            }
+            PublishSignedTransferPackageResponse::PaymentSent { payment } => {
+                info!("coop-exit iteration {iterations}: transfer ready, sent");
+                assert_eq!(payment.payment_type, PaymentType::Send);
+                assert!(
+                    matches!(
+                        payment.status,
+                        PaymentStatus::Completed | PaymentStatus::Pending
+                    ),
+                    "Coop-exit payment should be completed or pending"
```
dangeross, line 990: Unlike the other client-signing send tests, this one never checks the recipient: Bob only mints the address, his events/balance aren't read, and `Pending` is accepted. A coop-exit returns `Pending`, so it passes as long as `publish` returns a Send payment, and would stay green if the withdrawal used the wrong amount or address. Intentionally left shallow for now?

crates/breez-sdk/core/src/sdk/payments/send/mod.rs
```diff
@@ -4,16 +4,192 @@ pub(in crate::sdk::payments) mod cross_chain;
...
+    let cache = ObjectCacheRepository::new(sdk.storage.clone());
+    let res = match (&signed_package.unsigned, &signed_package.signature) {
+        (
+            UnsignedTransferPackage::Swap {
+                prepare_transfer, ..
+            },
+            TransferSignature::Transfer { .. },
+        ) => {
+            if let Ok(Some(_)) = cache
+                .fetch_published_package(&prepare_transfer.transfer_id)
```
dangeross: This replay guard (and the token/transfer ones below) returns the existing payment when the same package is published twice, and `make token and swap publishes replay-safe` is one of this PR's fix commits. No itest re-publishes a package, so the exact regression it guards against isn't exercised end to end. Worth a publish-twice test?

crates/spark-wallet/src/wallet/external_signing.rs
roeierez: Done

crates/spark/src/token/token_service.rs
roeierez: Done

crates/spark-wallet/src/wallet/external_signing.rs
roeierez: Done

crates/breez-sdk/core/src/sdk/payments/client_signing.rs
roeierez: Done

crates/breez-sdk/core/src/sdk/payments/send/mod.rs
roeierez: Done

crates/spark/src/token/token_service.rs
roeierez: True. Added a swap operation (user level) also for tokens. Until turnkey add policy for that will require two signatures but seems reasonable consider this is very rare case.

crates/breez-sdk/core/src/sdk/payments/send/bolt11.rs
roeierez: Good catch, only lightning was affected (lnurl got away from it with its own logic). I changed the normal path to align with the client signing path.

crates/breez-sdk/breez-itest/tests/client_signing_send.rs
roeierez: By mistake. Enhanced the test with relevant details.

crates/breez-sdk/core/src/sdk/payments/send/mod.rs
roeierez: Added test.

crates/breez-sdk/core/src/sdk/payments/send/bolt11.rs
danielgranhao: Ah, seems that LNURL also has an issue. For non-conversion `FeesIncluded` LNURL, pay.rs passes `amount_override = receiver_amount + overpayment`, so `amount_to_send` here (and the persisted `displayed_amount`) becomes `receiver_amount + overpayment` when it should be only `receive_amount`.

For the normal path this is a regression. Before we used `prepare_response.amount` (which was `receiver_amount`)

dangeross (approved):
Looks good from my side 👍

danielgranhao (approved):
LGTM

PR #974, opened 2026-07-01
perf(spark): batch FROST signing and derive owned-leaf pubkeys locally

danielgranhao:
## Summary

Reduces signer round-trips on the FROST signing hot paths (send, claim, coop-exit, timelock renewal, and leaf refresh). On a remote signer each `sign_frost` / `get_public_key_for_leaf` is a network round-trip, so the previous per-leaf-variant loops cost one round-trip per job. This batches the signing calls and, where the value is not a payout destination, derives owned-leaf signing pubkeys locally from persisted tree data. It also adds the Turnkey-side handling (request chunking + rate pacing) that the larger batched activities require.

### Signing hot paths (spark)

- **Batch FROST signing** into a single `sign_frost(Vec<FrostJob>)` call on the send (`sign_refunds`), claim, coop-exit, and timelock-renewal (`sign_signing_jobs`) paths. Jobs are built up front, then reattached to their per-variant outputs in order.
- **Derive owned-leaf signing pubkeys locally** as `verifying_public_key - signing_keyshare.public_key` instead of asking the signer, on the **send and coop-exit** paths. There the derived value is used only as the signing key (a wrong value fails closed as an invalid signature), never as a payout destination. Not applied to claim (incoming leaf uses the new key) or timelock (see Security).
- **Warm the claim leaf-pubkey cache** by running the key-tweak (`prepare_claim`) step before signing refunds, so per-leaf `get_public_key_for_leaf` in the claim path is a cache hit rather than a round-trip.
- **Fetch refresh-leaf pubkeys concurrently** in `refresh_leaves` (unbounded, order-preserving). A rate-limited signer (Turnkey) paces the calls internally, so no per-call-site concurrency cap is imposed.
- **Encode refund variants as struct fields**: `LeafRefundJobs` / `SignedLeafRefunds` replace the `RefundVariant` enum and the parallel jobs/pending lists. One `sign_leaf_refunds` helper flattens every leaf's jobs, signs the batch, and rebuilds in field order, confining the sole reliance on share ordering to one length-checked place. Shared by the send / claim / coop-exit paths.

### Turnkey remote signer

- **Chunk batched `sign_frost`** under Turnkey's per-activity limit: a single enclave `sign_frost` fails past ~250-300 jobs, so submissions are chunked at 100 jobs per activity (`MAX_FROST_JOBS_PER_ACTIVITY`).
- **Adaptive per-suborg rate pacer**: a client-level pacer holds requests to a configurable `max_rps` (default: Turnkey's documented 10 RPS), backing off on 429 and recovering on success, shared across all concurrent callers of one suborganization. Replaces the per-call-site concurrency caps.

### Security

- **Timelock renewal fetches its signing key from the signer**, not derived. The renewed exit refund pays to this key, so deriving it from persisted (coordinator-supplied) tree data would let a malicious coordinator steer the refund destination. Renewal is infrequent and off the send hot path, so the round-trip cost is negligible.

## Testing

- `cargo clippy -p spark` and `cargo clippy -p breez-sdk-spark --features turnkey` clean
- `cargo test -p spark` (247 tests) passing
- `cargo test -p breez-sdk-spark --features turnkey` (turnkey transport / pacer tests) passing

### DIFF
--- a/crates/breez-sdk/core/src/turnkey/spark_signer.rs
+++ b/crates/breez-sdk/core/src/turnkey/spark_signer.rs
@@ -25,7 +25,6 @@ use std::collections::{BTreeMap, HashMap};
 use std::str::FromStr;
 use std::sync::Arc;
 
-use futures::{StreamExt, TryStreamExt};
 use tokio::sync::Mutex;
 
 use bitcoin::bip32::DerivationPath;
@@ -274,13 +273,8 @@ fn ffi_public_key_map(
 /// to 300 jobs, far below its ~11 MB request-body cap. The batched send/coop-exit/
 /// claim/timelock paths can exceed that (a full-balance coop-exit signs every
 /// wallet leaf, one to three jobs each), so submissions are chunked with wide
-/// margin. Measured by the probe in `breez-itest/tests/turnkey_frost_batch_limit`.
+/// margin. Exercised by `breez-itest/tests/turnkey_sign_frost_chunking`.
 const MAX_FROST_JOBS_PER_ACTIVITY: usize = 100;
-
-/// How many chunk activities to submit concurrently. Kept low so a many-chunk
-/// batch stays within Turnkey's 10 RPS per-suborganization rate limit (each
-/// activity is a submit plus a few status polls).
-const SIGN_FROST_CHUNK_CONCURRENCY: usize = 4;
 
 pub(crate) struct TurnkeySparkSigner {
     client: Arc<TurnkeyClient>,
@@ -450,21 +444,17 @@ impl TurnkeySparkSigner {
         jobs: &[FrostJob],
     ) -> Result<Vec<FrostShareResult>, SignerError> {
         let sign_with = self.spark_identity_address().await?;
-        // One activity per chunk (see MAX_FROST_JOBS_PER_ACTIVITY), submitted with
-        // bounded concurrency and reassembled in job order. FROST jobs are
-        // independent, so splitting the batch is signing-equivalent to one call.
-        let chunks: Vec<Vec<FrostJob>> = jobs
-            .chunks(MAX_FROST_JOBS_PER_ACTIVITY)
-            .map(<[FrostJob]>::to_vec)
-            .collect();
-        let chunk_shares: Vec<Vec<FrostShareResult>> = futures::stream::iter(chunks)
-            .map(|chunk| {
-                let sign_with = sign_with.clone();
-                async move { self.sign_frost_chunk(sign_with, &chunk).await }
-            })
-            .buffered(SIGN_FROST_CHUNK_CONCURRENCY)
-            .try_collect()
-            .await?;
+        // One activity per chunk (see MAX_FROST_JOBS_PER_ACTIVITY), all fired
+        // concurrently and reassembled in job order. FROST jobs are independent,
+        // so splitting the batch is signing-equivalent to one call. The Turnkey
+        // client paces submissions to the per-suborg rate limit, so no concurrency
+        // cap is needed here.
+        let chunk_futures = jobs.chunks(MAX_FROST_JOBS_PER_ACTIVITY).map(|chunk| {
+            let sign_with = sign_with.clone();
+            async move { self.sign_frost_chunk(sign_with, chunk).await }
+        });
+        let chunk_shares: Vec<Vec<FrostShareResult>> =
+            futures::future::try_join_all(chunk_futures).await?;
         Ok(chunk_shares.into_iter().flatten().collect())
     }
 
--- a/crates/breez-sdk/core/src/turnkey/spark_signer.rs
+++ b/crates/breez-sdk/core/src/turnkey/spark_signer.rs
@@ -564,6 +564,10 @@ impl ExternalSparkSigner for TurnkeySparkSigner {
         ))
     }
 
+    fn is_remote(&self) -> bool {
+        true
+    }
+
     async fn get_static_deposit_public_key(
         &self,
         index: u32,
--- a/crates/breez-sdk/core/src/turnkey/spark_signer.rs
+++ b/crates/breez-sdk/core/src/turnkey/spark_signer.rs
@@ -959,6 +959,7 @@ mod tests {
             network: Network::Regtest,
             account_number: Some(0),
             retry: None,
+            max_rps: None,
         };
         Arc::new(TurnkeyClient::new(&config, http).unwrap())
     }
--- a/crates/breez-sdk/core/src/turnkey/spark_signer.rs
+++ b/crates/breez-sdk/core/src/turnkey/spark_signer.rs
@@ -25,6 +25,7 @@ use std::collections::{BTreeMap, HashMap};
 use std::str::FromStr;
 use std::sync::Arc;
 
+use futures::{StreamExt, TryStreamExt};
 use tokio::sync::Mutex;
 
 use bitcoin::bip32::DerivationPath;
@@ -268,6 +269,19 @@ fn ffi_public_key_map(
         .collect()
 }
 
+/// Maximum FROST jobs per `SPARK_SIGN_FROST` activity. Turnkey's enclave fails a
+/// single `sign_frost` with HTTP 500 (an internal ~15s timeout) past roughly 250
+/// to 300 jobs, far below its ~11 MB request-body cap. The batched send/coop-exit/
+/// claim/timelock paths can exceed that (a full-balance coop-exit signs every
+/// wallet leaf, one to three jobs each), so submissions are chunked with wide
+/// margin. Measured by the probe in `breez-itest/tests/turnkey_frost_batch_limit`.
+const MAX_FROST_JOBS_PER_ACTIVITY: usize = 100;
+
+/// How many chunk activities to submit concurrently. Kept low so a many-chunk
+/// batch stays within Turnkey's 10 RPS per-suborganization rate limit (each
+/// activity is a submit plus a few status polls).
+const SIGN_FROST_CHUNK_CONCURRENCY: usize = 4;
+
 pub(crate) struct TurnkeySparkSigner {
     client: Arc<TurnkeyClient>,
     account: u32,
@@ -436,8 +450,33 @@ impl TurnkeySparkSigner {
         jobs: &[FrostJob],
     ) -> Result<Vec<FrostShareResult>, SignerError> {
         let sign_with = self.spark_identity_address().await?;
-        let mut signatures = Vec::with_capacity(jobs.len());
-        for job in jobs {
+        // One activity per chunk (see MAX_FROST_JOBS_PER_ACTIVITY), submitted with
+        // bounded concurrency and reassembled in job order. FROST jobs are
+        // independent, so splitting the batch is signing-equivalent to one call.
+        let chunks: Vec<Vec<FrostJob>> = jobs
+            .chunks(MAX_FROST_JOBS_PER_ACTIVITY)
+            .map(<[FrostJob]>::to_vec)
+            .collect();
+        let chunk_shares: Vec<Vec<FrostShareResult>> = futures::stream::iter(chunks)
+            .map(|chunk| {
+                let sign_with = sign_with.clone();
+                async move { self.sign_frost_chunk(sign_with, &chunk).await }
+            })
+            .buffered(SIGN_FROST_CHUNK_CONCURRENCY)
+            .try_collect()
+            .await?;
+        Ok(chunk_shares.into_iter().flatten().collect())
+    }
+
+    /// Signs one chunk of FROST jobs as a single `SPARK_SIGN_FROST` activity,
+    /// returning the shares in job order.
+    async fn sign_frost_chunk(
+        &self,
+        sign_with: String,
+        chunk: &[FrostJob],
+    ) -> Result<Vec<FrostShareResult>, SignerError> {
+        let mut signatures = Vec::with_capacity(chunk.len());
+        for job in chunk {
             signatures.push(frost_job_to_request(job).map_err(to_spark_err)?);
         }
         let result: SparkSignFrostResult = self
@@ -454,10 +493,10 @@ impl TurnkeySparkSigner {
             )
             .await
             .map_err(to_spark_err)?;
-        if result.signatures.len() != jobs.len() {
+        if result.signatures.len() != chunk.len() {
             return Err(SignerError::Generic(format!(
                 "turnkey sign_frost: expected {} shares, got {}",
-                jobs.len(),
+                chunk.len(),
                 result.signatures.len()
             )));
         }
--- a/crates/breez-sdk/core/src/turnkey/transport.rs
+++ b/crates/breez-sdk/core/src/turnkey/transport.rs
@@ -6,9 +6,10 @@
 //! the activity reaches a terminal status.
 
 use std::collections::HashMap;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
+use platform_utils::time::Instant;
 use platform_utils::tokio::time::sleep;
 use platform_utils::{HttpClient, HttpResponse};
 use serde::Serialize;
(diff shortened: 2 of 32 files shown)


crates/spark/src/utils/frost.rs
```diff
@@ -1,11 +1,53 @@
...
+/// leaf is mid-transfer, so its stored SE share need not pair with the new key.
+pub(crate) fn derive_leaf_signing_public_key(
+    node: &TreeNode,
+    secp: &Secp256k1<All>,
+) -> Result<PublicKey, SignerError> {
+    let se_share = node.signing_keyshare.public_key.negate(secp);
+    node.verifying_public_key
+        .combine(&se_share)
+        .map_err(|e| SignerError::Generic(format!("failed to derive leaf signing key: {e}")))
+}
```
dangeross, line 30: Is it worth adding a unit test for this?

crates/spark/src/utils/refund.rs
```diff
@@ -131,35 +135,61 @@ pub async fn sign_refunds(
...
+    // Sign every job in a single batched call. Remote-signer backends (e.g.
+    // Turnkey) collapse this into one round-trip instead of one per leaf-variant.
+    let shares = spark_signer.sign_frost(jobs).await?;
+    if shares.len() != pending.len() {
+        return Err(SignerError::Generic(format!(
+            "sign_frost returned {} shares, expected {}",
+            shares.len(),
+            pending.len()
+        )));
+    }
```
dangeross: `coop_exit`, `transfer`, and `signing_job` all route through `sign_frost_batch`; the send path reimplements it inline here (call + length check + zip). The helper is generic over the pending type, so it drops in as-is and deletes the check. Fold it in so the four paths can't drift?

crates/spark/src/utils/frost.rs
```diff
@@ -1,11 +1,53 @@
...
+/// Signs a batch of FROST jobs in a single `sign_frost` call, pairing each
+/// returned share with its caller-side metadata (order preserved). Remote signer
+/// backends (e.g. Turnkey) collapse this into one round-trip instead of one per
+/// job, so callers should build all jobs up front rather than signing per item.
+pub(crate) async fn sign_frost_batch<T>(
+    spark_signer: &Arc<dyn SparkSigner>,
+    jobs: Vec<FrostJob>,
+    pending: Vec<T>,
+) -> Result<Vec<(T, FrostShareResult)>, SignerError> {
+    let shares = spark_signer.sign_frost(jobs).await?;
```
dangeross, line 58: The pubkey fetch in `refresh_leaves` is capped at 16, but this batches every job into one unbounded `sign_frost` call, so a large send or coop-exit becomes a single request whose payload scales with leaf count. Is the batch bounded anywhere upstream, or are these paths assumed small enough that one request is always fine?

dangeross (commented):
Looking good

crates/spark/src/utils/frost.rs
danielgranhao: Done

crates/spark/src/utils/refund.rs
danielgranhao: Done

crates/spark/src/utils/frost.rs
danielgranhao: Good catch. I didn't find a documented max on turnkey's docs but experimentally I verified that around 250-300 jobs in a batch, the activity fails with an internal server error. I've split batches into 100 job chunks to make sure we don't bump into that.

crates/breez-sdk/core/src/turnkey/spark_signer.rs
```diff
@@ -268,6 +269,19 @@ fn ffi_public_key_map(
         .collect()
 }
 
+/// Maximum FROST jobs per `SPARK_SIGN_FROST` activity. Turnkey's enclave fails a
+/// single `sign_frost` with HTTP 500 (an internal ~15s timeout) past roughly 250
+/// to 300 jobs, far below its ~11 MB request-body cap. The batched send/coop-exit/
+/// claim/timelock paths can exceed that (a full-balance coop-exit signs every
+/// wallet leaf, one to three jobs each), so submissions are chunked with wide
+/// margin. Measured by the probe in `breez-itest/tests/turnkey_frost_batch_limit`.
```
dangeross: This `breez-itest/tests/turnkey_frost_batch_limit` doesn't exist

dangeross (commented):
Looks good

crates/spark/src/services/timelock_manager.rs
```diff
@@ -188,7 +190,9 @@ impl TimelockManager {
         info!("Renewing node: {:?}", node.id);
         let mut signing_jobs = Vec::new();
 
-        let signing_public_key = self.spark_signer.get_public_key_for_leaf(&node.id).await?;
+        // Owned node: recover the signing key locally from the tree node instead
+        // of a signer round-trip (see derive_leaf_signing_public_key).
+        let signing_public_key = derive_leaf_signing_public_key(node, &Secp256k1::new())?;
```
JssDWt: I think all 3 signing public keys in this file should be taken from the signer directly, rather than derived. The reason is `refresh_leaves` is not the only place where leaves are inserted into storage. So it is not guaranteed that the verifying public key and the signing keyshare public key are correct here. If the coordinator wants to attack you, they could replace the signing keyshare public key with one chosen by them. I think that makes you pay the attacker. It also wouldn't have the 1/N trust guarantee then, because you query only a single operator. So I think the rule should be, wherever you would actually use this signing public key in a cryptographic operation, we have to take it from the signer.

How would that affect performance? Should be OK I think, because renewing doesn't happen all that often right?

crates/spark/src/services/coop_exit.rs
```diff
@@ -474,15 +475,17 @@ impl CoopExitService {
...
     > {
-        let mut cpfp_jobs = Vec::new();
-        let mut direct_jobs = Vec::new();
-        let mut direct_from_cpfp_jobs = Vec::new();
+        // Build every leaf-variant connector-refund FROST job up front, then sign
+        // the whole batch in one call. A per-job loop would cost one sign_frost
+        // network round-trip per leaf-variant on a remote signer.
+        let mut jobs = Vec::new();
+        let mut pending = Vec::new();
+        let secp = Secp256k1::new();
```
JssDWt: Let's put secp in a field on the CoopExitService

crates/spark/src/utils/frost.rs
```diff
@@ -1,11 +1,53 @@
...
+/// The user's own signing public key for an OWNED leaf, derived from persisted
+/// tree data instead of asking the signer: `verifying_public_key -
+/// signing_keyshare.public_key`. FROST composes the group verifying key as the
+/// user's verifying share plus the operators' aggregate share, and
+/// `refresh_leaves` validates this relation for every Available leaf, so the
+/// user's share is recoverable locally. This avoids a per-leaf signer round-trip
+/// on the send/coop-exit/timelock hot paths (for a remote signer with a cold
+/// in-memory cache, e.g. a per-request server instance, that would otherwise be
+/// one network call per leaf). Only valid for owned leaves: an incoming (claim)
+/// leaf is mid-transfer, so its stored SE share need not pair with the new key.
```
JssDWt: Comment leaking rest of codebase. (violates rule 6 of comments in CLAUDE.md)

crates/spark/src/services/transfer.rs
```diff
@@ -833,94 +839,71 @@ impl TransferService {
...
+        let mut direct_jobs = Vec::new();
+        let mut direct_from_cpfp_jobs = Vec::new();
+        for (entry, share) in signed {
+            let (variant, job) = entry.into_user_signed_job(share)?;
+            match variant {
+                RefundVariant::Cpfp => cpfp_jobs.push(job),
+                RefundVariant::Direct => direct_jobs.push(job),
+                RefundVariant::DirectFromCpfp => direct_from_cpfp_jobs.push(job),
+            }
+        }
```
JssDWt: This same structure is in a few places. It contains a loop+match patterns and zipping two equal-length lists. I'm not a fan of both. loop+match usually means you can put something in a struct with fields.

I had Claude make this suggestion:

### Suggestion

Keep the single `sign_frost` round-trip, but carry structure through named fields instead
of a flat list. Encode the variant in the field name (drop `RefundVariant`):

```rust
struct RefundJob { job: FrostJob, pending: PendingRefundSignature } // no `variant`
struct LeafRefundJobs   { cpfp: RefundJob, direct: Option<RefundJob>, direct_from_cpfp: Option<RefundJob> }
struct SignedLeafRefunds{ cpfp: SignedTx,  direct: Option<SignedTx>,  direct_from_cpfp: Option<SignedTx> }
```

One helper owns flatten → sign (once) → reattach; `jobs()`/`rebuild()` iterate the same
fields in the same order so they can't drift:

```rust
async fn sign_leaf_refunds(signer, leaves: Vec<LeafRefundJobs>) -> Result<Vec<SignedLeafRefunds>> {
    let jobs: Vec<FrostJob> = leaves.iter().flat_map(LeafRefundJobs::jobs).collect();
    let shares = signer.sign_frost(jobs).await?; // length-check as today
    let mut shares = shares.into_iter();
    leaves.into_iter().map(|l| l.rebuild(&mut shares)).collect()
}
```

Call sites become one-liners (three-way split is just a field projection, no tag/match):

```rust
let (cpfp, direct, dfc) = sign_leaf_refunds(signer, leaf_jobs).await?.into_signed_tx_groups();        // send
let (cpfp, direct, dfc) = sign_leaf_refunds(signer, leaf_jobs).await?.into_user_signed_job_groups()?; // coop-exit / claim
```

**Deletes:** `RefundVariant`, `PendingRefundSignature.variant`, `into_user_signed_job`'s tuple,
all three `match variant` loops, and the twin-`jobs`/`pending` lists.

**Timelock (`signing_job.rs`):** smaller fix — fold `operator_commitments` into the `SigningJob`
metadata so it rides with its job instead of being a third list zipped twice.

Net: one round-trip preserved; no variant enum; no hand-maintained parallel lists; the one
remaining "results in job order" reliance is confined to a single length-checked helper.

JssDWt (commented):
Looking good. Great performance improvement. I have a few comments.

crates/spark/src/services/transfer.rs
danielgranhao: Thanks, done

crates/spark/src/utils/frost.rs
danielgranhao: fixed

crates/spark/src/services/coop_exit.rs
danielgranhao: done

crates/spark/src/services/timelock_manager.rs
danielgranhao: Agreed. Nice catch. Reverted

crates/breez-sdk/core/src/turnkey/spark_signer.rs
danielgranhao: Fixed

crates/spark/src/services/models.rs
```diff
@@ -145,6 +150,218 @@ impl TryFrom<&SignedTx> for operator_rpc::spark::UserSignedTxSigningJob {
...
+struct LeafRefundShape {
+    has_direct: bool,
+    has_direct_from_cpfp: bool,
+}
+
+impl LeafRefundJobs {
+    /// Moves this leaf's jobs and pending metadata into the shared batch buffers
+    /// in field order (cpfp, then direct, then direct-from-cpfp), returning the
+    /// variant shape needed to regroup the shares. Moves rather than clones: each
+    /// job is consumed by the single `sign_frost` batch.
```
dangeross: Restates the mechanism (no `.clone()` already says it moves) and asserts a downstream fact. The sentence above already states what `flatten_into` does.

```suggestion
    /// variant shape needed to regroup the shares.
```

crates/spark/src/services/models.rs
```diff
@@ -145,6 +150,218 @@ impl TryFrom<&SignedTx> for operator_rpc::spark::UserSignedTxSigningJob {
...
+        .next()
+        .ok_or_else(|| SignerError::Generic("sign_frost returned too few shares".to_string()))
+}
+
+/// Signs every leaf's refund jobs in one batched `sign_frost` call, then
+/// reattaches the shares. Flatten and rebuild walk the same fields in the same
+/// order, and the count is length-checked inside `sign_frost_batch`, so the
+/// single reliance on share ordering is confined here. Remote signer backends
+/// (e.g. Turnkey) collapse the whole batch into one round-trip instead of one
+/// per job.
```
dangeross: I think this leaks signer implementation specific behavior.
```suggestion
/// single reliance on share ordering is confined here.
```

crates/spark/src/utils/frost.rs
```diff
@@ -1,11 +1,65 @@
...
+            pending.len()
+        )));
+    }
+    Ok(pending.into_iter().zip(shares).collect())
+}
+
+/// Signs a batch of FROST jobs in a single `sign_frost` call, pairing each
+/// returned share with its caller-side metadata (order preserved). Remote signer
+/// backends (e.g. Turnkey) collapse the whole batch into one round-trip instead
+/// of one per job.
```
dangeross: Same `Turnkey` leak in signer-agnostic code.

```suggestion
/// returned share with its caller-side metadata (order preserved).
```

crates/spark/src/utils/refund.rs
```diff
@@ -109,131 +115,99 @@ pub async fn sign_refunds(
...
+        leaf_jobs.push(LeafRefundJobs {
+            cpfp,
+            direct,
+            direct_from_cpfp,
+        });
     }
 
+    // Sign every leaf's jobs in one batched call, then split by variant. Remote
+    // signer backends (e.g. Turnkey) collapse this into one round-trip instead of
+    // one per leaf-variant.
```
dangeross: Same `Turnkey` leak.

```suggestion
    // Sign every leaf's jobs in one batched call, then split by variant.
```

crates/spark/src/services/coop_exit.rs
```diff
@@ -474,15 +477,16 @@ impl CoopExitService {
         ),
         ServiceError,
     > {
-        let mut cpfp_jobs = Vec::new();
-        let mut direct_jobs = Vec::new();
-        let mut direct_from_cpfp_jobs = Vec::new();
+        // Build every leaf-variant connector-refund FROST job up front, then sign
+        // the whole batch in one call. A per-job loop would cost one sign_frost
+        // network round-trip per leaf-variant on a remote signer.
```
dangeross: Not sure it belongs here
```suggestion
        // Build every leaf-variant connector-refund FROST job up front, then sign
        // the whole batch in one call.
```

crates/spark/src/services/transfer.rs
```diff
@@ -781,9 +787,10 @@ impl TransferService {
         ),
         ServiceError,
     > {
-        let mut cpfp_jobs = Vec::new();
-        let mut direct_jobs = Vec::new();
-        let mut direct_from_cpfp_jobs = Vec::new();
+        // Build every leaf-variant claim-refund FROST job up front, then sign the
+        // whole batch in one call. A per-job loop would cost one sign_frost
+        // network round-trip per leaf-variant on a remote signer.
```
dangeross: Same here.
```suggestion
        // Build every leaf-variant claim-refund FROST job up front, then sign the
        // whole batch in one call.
```

crates/spark/src/utils/signing_job.rs
```diff
@@ -45,39 +46,42 @@ pub async fn sign_signing_jobs(
     signing_commitments: Vec<BTreeMap<Identifier, SigningCommitments>>,
     network: Network,
 ) -> Result<Vec<SignedJob>, SignerError> {
-    let mut signed_txs = Vec::new();
-
+    // Build every renewal-tx FROST job up front, then sign the whole batch in one
+    // call: remote signers turn each sign_frost into a network round-trip, so a
+    // per-job loop would cost one round-trip per renewal.
```
dangeross: Same here.

```suggestion
    // Build every renewal-tx FROST job up front, then sign the whole batch in one
    // call.
```

crates/spark/src/tree/service.rs
```diff
@@ -259,27 +259,50 @@ impl TreeService for SynchronousTreeService {
...
+            .collect();
 
-            let our_node_pubkey = self.spark_signer.get_public_key_for_leaf(&leaf.id).await?;
-            let other_node_pubkey = leaf.signing_keyshare.public_key;
-            let verifying_pubkey = leaf.verifying_public_key;
+        // Fetch our signing pubkey for each Available leaf concurrently, so a
+        // remote signer resolves them in parallel instead of one blocking
+        // round-trip per leaf. A rate-limited signer (e.g. Turnkey) paces the
+        // calls internally, so no concurrency cap is imposed here. Order is
+        // preserved for the zip below.
```
dangeross: Leaking of signer implementation specific behavior.

dangeross (commented):
Looks good to me. Just nits on comments leaking signer implementation specifics in agnostic code

dangeross (approved):
LGTM

crates/spark/src/tree/service.rs
```diff
@@ -259,27 +259,48 @@ impl TreeService for SynchronousTreeService {
...
+        // Order is preserved for the zip below.
+        // TODO: validate each leaf once and persist the result, to skip
+        // re-checking unchanged leaves on every refresh.
+        let signer = &self.spark_signer;
+        let our_pubkeys: Vec<PublicKey> = futures::future::try_join_all(
+            available_leaves
+                .iter()
+                .map(|leaf| async move { signer.get_public_key_for_leaf(&leaf.id).await }),
+        )
+        .await?;
```
roeierez, line 308: We are still doing network requests here, just concurrently right? 
Isn't that going to be a problem with large amount of leaves for each refresh? Other than many requests seems like it may force payments to wait? (due to the rate limiter)
Assuming leaves enters the store only after verification, can we just lookup for the leaf in the store and if exists we can skip fetching the public key from the signer?

crates/breez-sdk/core/src/turnkey/transport.rs
```diff
@@ -181,22 +286,52 @@ impl TurnkeyClient {
...
+        // Set on the first attempt, after the initial paced slot. Queue wait
+        // (waiting our turn behind other concurrent callers) is deliberately
+        // excluded from the timeout: it must not consume the request's retry
+        // budget, or a deeply-queued request would arrive with none left.
+        let mut started: Option<Instant> = None;
         loop {
+            // Pace emission to the adaptive rate before every attempt, so
+            // concurrent callers self-throttle to Turnkey's per-suborg RPS limit
+            // rather than bursting into 429s.
+            let wait = self.pacer.reserve();
```
roeierez: If the request was canceled by the caller it seems the slot is not reclaimed? Perhaps we need a drop guard here?

roeierez (changes requested):
Looks good!

crates/breez-sdk/core/src/turnkey/transport.rs
danielgranhao: Done

crates/spark/src/tree/service.rs
danielgranhao: Right. I was leaving it for later, but you're right that we should do it now. Done

crates/spark/src/tree/service.rs
```diff
@@ -259,27 +259,71 @@ impl TreeService for SynchronousTreeService {
...
+
+        // A leaf enters the store only after this ownership check has passed, so
+        // one already stored with the same verifying and signing-keyshare keys is
+        // known-good and skipped. Only leaves new or changed since the last
+        // refresh need their pubkey fetched, which is a network round-trip on some
+        // signers: re-deriving every leaf each refresh would flood the signer and
+        // stall payments behind the rate limiter on large wallets. Remaining
+        // fetches run concurrently, leaving the signer to bound its own
+        // concurrency; order is preserved for the zip below.
+        let stored_leaves = self.state.get_leaves().await?;
```
roeierez: I am a bit concerned to fetch and load all leaves into memory from the db on refresh.
I have seen users with tens of thousands of leaves. Also most lots of users won't use turnkey.
Perhaps fetch only the leaves ids? Also can we scope it only to the kind of signers actually use network requests?

roeierez (commented):
Looks good. One comment.

crates/spark/src/tree/service.rs
danielgranhao: Done

roeierez (approved):
LGTM

PR #975, opened 2026-07-01
Turnkey signer provisioning for network-free re-init

danielgranhao:
## Problem

Server deployments build an SDK, and its signer, per request. Creating a Turnkey signer costs sequential network round-trips: materialize the enclave identity account, then export the SDK-layer ECIES/HMAC key. #970 made the export lazy, but the memo is in-memory per instance, so every per-request instance that encrypts (sync, session storage, LNURL-auth) re-pays the export. Laziness defers the cost; it can't amortize it across instances.

## Change

Split the factory into two entry points:

- **`provision_turnkey_signer(config)`** — run once, at user creation. Materializes the identity account, exports the encryption key, and captures the wallet's stable Spark values, returning an opaque, versioned, wallet-bound `TurnkeyProvisionedSigner` blob to persist (encrypted).
- **`create_turnkey_signer(config, provisioned)`** — with a blob, rebuilds the signer with **no network**: seeds the encryption backend and the Spark signer's identity pubkey / Spark address, and skips the identity create. Without a blob (mobile/CLI), behavior is unchanged (eager identity create, lazy export/fetch).

### What the blob carries

- **Encryption key** (or a record that export is policy-denied) → seeds the Breez signer's ECIES/HMAC backend.
- **Identity public key + Spark address** → seeds the Spark signer's memos. These are stable and deterministic per wallet, so a provisioned per-request signer serves them from memory instead of re-fetching (`get_wallet_account` / `create_wallet_accounts`) on every request.

Dynamic per-leaf and static-deposit keys are intentionally **not** in the blob (they change with the wallet's leaves) and stay lazy.

### Absent-key disambiguation

The blob's encryption state is a two-state enum, resolved authoritatively at provisioning time:

- `Key(..)` → seed the local ECIES/HMAC backend, no network.
- `Unavailable` (export 403'd under a deny policy) → the built signer reports `EncryptionUnavailable` **without ever probing**, so a provisioned no-export wallet doesn't re-attempt the export per request.

### Staleness

A blob that does not match the config (wallet / network / account, or an older layout version) is rejected with the new `SignerError::ProvisioningOutdated`, so the caller re-provisions (e.g. after an SDK upgrade or a wallet policy change).

### Relationship to lazy export (#970)

Lazy export/fetch is kept for the **unprovisioned** path. Persistence seeds the existing memos (`EncryptionBackend::{Seeded,Denied,Lazy}` for the Breez signer; `TurnkeySparkSigner::new_seeded` for the Spark signer) rather than replacing the lazy machinery, so the integration is a pre-fill and the delta stays small.

## Caller shape

```ts
// once, at user creation
const { bytes } = await provisionTurnkeySigner(config)
await store.set(userId, bytes)

// per request — no network in createTurnkeySigner
const signers = await createTurnkeySigner(config, await store.get(userId))
await connectWithSigner({ config, signers })
```

## Commits

1. **provisioning for network-free re-init** — the `provision`/`create` split, encryption seeding, deny handling, staleness.
2. **seed Spark identity pubkey and address** — closes a gap found in server-mode measurement: the first commit removed the export from the hot path, but the Spark signer still lazy-fetched its identity pubkey / Spark address per request. Carrying them in the blob (bumping `PROVISION_VERSION` to 2) makes provisioned per-request init truly network-free for signer setup.

### DIFF
--- a/crates/breez-sdk/core/src/turnkey/factory.rs
+++ b/crates/breez-sdk/core/src/turnkey/factory.rs
@@ -6,14 +6,15 @@ use std::sync::Arc;
 use bitcoin::secp256k1::{PublicKey, SecretKey};
 use platform_utils::create_http_client;
 use serde::{Deserialize, Serialize};
-use spark_wallet::SparkAddress;
 
 use crate::ExternalSigners;
 use crate::error::SignerError;
 use crate::signer::breez::BreezSignerImpl;
 use crate::signer::{ExternalBreezSigner, ExternalSparkSigner};
 
-use super::accounts::{encryption_key_path, spark_address_format, xpriv_from_secret};
+use super::accounts::{
+    encryption_key_path, spark_address_format, spark_identity_address, xpriv_from_secret,
+};
 use super::breez_signer::{EncryptionBackend, TurnkeyBreezSigner};
 use super::config::TurnkeyConfig;
 use super::error::TurnkeyError;
@@ -41,7 +42,7 @@ fn identity_path(account: u32) -> String {
 /// Layout version of [`TurnkeyProvisionedSigner`]'s bytes. Bumped when the blob
 /// contents change so an older blob is rejected as outdated rather than
 /// misread; the caller re-provisions to upgrade.
-const PROVISION_VERSION: u16 = 2;
+const PROVISION_VERSION: u16 = 3;
 
 /// The encryption-key verdict captured at provisioning time.
 #[derive(Serialize, Deserialize)]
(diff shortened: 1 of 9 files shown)


crates/breez-sdk/core/src/turnkey/factory.rs
```diff
@@ -31,63 +39,413 @@ fn identity_path(account: u32) -> String {
...
+/// Dynamic per-leaf keys are not here either: they change with the wallet's
+/// leaves and stay lazy.
+#[derive(Serialize, Deserialize)]
+struct ProvisionBlob {
+    version: u16,
+    network: u8,
+    account: u32,
+    organization_id: String,
+    wallet_id: String,
+    encryption: ProvisionedEncryption,
```
roeierez, line 76: Persisting the encryption key (that encrypts the sessions) effectively invalidates the session encryption.
I would like to suggest exploring another approach for the problem:

The only "hot path" where the encryption is needed is in session encryption (sync, lnurl auth and cross chain are not likely to be used on this deployment) and I think by removing the session encryption (which I personally don't think it is the sdk job to do, especially when we provide a trait that users can pass their own implementation with encryption) then we end up with a green light for the lazy approach.
I think also the identity pub key can be given as an optional parameter to the TurnkeyConfig and allows us to skip the fetch from turnkey.

@danielgranhao LMK what you think about this approach.

roeierez (changes requested):

danielgranhao:
Closing in favor of #982

PR #976, opened 2026-07-02
Fix cross-chain send amount contract for stable-balance conversions

dangeross:
For a cross-chain send with no `token_identifier`, `amount` is in **sats**. But when stable balance auto-injects a `ToBitcoin` conversion, the dispatcher treated `amount` as **destination base units** (fiat-inverted), causing:
- tiny sends rejected at prepare (`Amount out of range: 358 sats (min: 1000)`), and
- `insufficient funds` at send (conversion targeted the pre-fee amount, not the provider invoice).

### Fix
Route the `conversion + no token_identifier` case to a new `prepare_sats_denominated_with_conversion` dispatcher: size the provider leg in **sats** and fund it via a token→sats conversion covering `invoice + source_transfer_fee`, with `response.amount` set to the invoice so the send-stage conversion targets it.

Only the previously-broken path changes; direct and token-denominated sends are untouched.

### DIFF
--- a/crates/breez-sdk/core/src/sdk/payments/prepare/cross_chain.rs
+++ b/crates/breez-sdk/core/src/sdk/payments/prepare/cross_chain.rs
@@ -205,24 +205,48 @@ async fn estimate_and_validate_conversion(
     )
     .await?;
 
-    if let Some(ref ce) = estimate
-        && let ConversionType::ToBitcoin {
-            from_token_identifier,
-        } = &ce.options.conversion_type
-    {
-        let balances = sdk.spark_wallet.get_token_balances().await?;
-        let have = balances
-            .get(from_token_identifier)
-            .map_or(0u128, |b| b.balance);
-        if have < ce.amount_in {
-            return Err(SdkError::InvalidInput(format!(
-                "Insufficient {from_token_identifier} balance for conversion: have {have}, need {}.",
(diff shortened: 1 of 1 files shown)


crates/breez-sdk/core/src/sdk/payments/prepare/cross_chain.rs
```diff
@@ -384,6 +402,97 @@ fn resolve_direct_overpay_amount(amount: u128, fee_policy: FeePolicy, overpay_bp
...
+            .get(from_token_identifier)
+            .map_or(0u128, |b| b.balance);
+        if have < estimate.amount_in {
+            return Err(SdkError::InvalidInput(format!(
+                "Insufficient {from_token_identifier} balance for cross-chain send: \
+                 have {have}, need {} (= {amount} sats + AMM/provider spread + bridge fees).",
+                estimate.amount_in
+            )));
+        }
+    }
```
danielgranhao: This seems to be duplicating lines 208–223 and 669–692. Could be worth extracting a `ensure_token_balance_covers(sdk, estimate) -> Result<(), SdkError>` helper

danielgranhao (commented):
Looks good. Just a nit

danielgranhao (approved):

JssDWt (approved):

PR #978, opened 2026-07-04
fix(turnkey): create leaf accounts as Bitcoin P2TR

JssDWt:
Correctness fix for the Turnkey signer.

Each wallet leaf gets its own account on the Turnkey side. We were creating a generic (compressed-key) account, but a leaf's funds are actually held at a Bitcoin P2TR (taproot) address. This change creates the leaf account as the matching Bitcoin P2TR account, so the account type matches how the funds are held.

It does not change any existing behaviour: the underlying key is the same, so signing and balances are unaffected. It only ensures the right type of account is created on Turnkey.


dangeross (approved):
LGTM

danielgranhao (approved):

PR #981, opened 2026-07-07
docs: Add using external signer with SdkBuilder section

dangeross:
Add a section to the External Signer page on advanced building of an SDK with external signers


roeierez (approved):
LGTM

PR #982, opened 2026-07-08
Turnkey network-free re-init via config identity pubkey; drop SDK session encryption

danielgranhao:
Replaces #975.

**Problem.** Server deployments build an SDK and its Turnkey signer per request. Signer setup costs sequential Turnkey round-trips: it materializes the identity account and exports the SDK-layer ECIES key. #970 made the export lazy, but the memo is per-instance, so every per-request signer re-pays it. Laziness defers the cost, it cannot amortize it across instances.

**Why not #975.** #975 amortized by persisting a provisioning blob (encryption key, identity pubkey, Spark address). But persisting the encryption key defeats the session encryption it exists for (feedback: [discussion_r3537771334](https://github.com/breez/spark-sdk/pull/975#discussion_r3537771334)).

**This alternative** removes the same cost without persisting any secret:

- **Drop SDK session encryption.** The ECIES export sat on the hot path only to encrypt session tokens. Encrypting them is not the SDK's job when a pluggable `SessionStore` already lets integrators do it their own way, so the SDK now stores tokens verbatim and the export leaves the hot path entirely, leaving nothing to persist. Integrators who want at-rest encryption wrap the backend's store via `with_session_store` / `default_session_store`.
- **Skip the identity fetch via config.** A new optional `TurnkeyConfig.identity_public_key` (a stable, non-secret, per-wallet value) lets init serve the identity key and its locally derived Spark address from memory, skipping both `create_account` round-trips. Unset keeps the prior lazy behavior.

Net: per-request signer setup is network-free, and session-encryption integrity is intact because no key is persisted.

**Mixed-version safety.** A transitional `LegacyTokenGuard` treats tokens an earlier version wrote encrypted and tag-prefixed as absent, so a rolling deployment re-authenticates per user on first access instead of serving a stale token until its TTL.


crates/breez-sdk/core/src/session_store/legacy_guard.rs
```diff
@@ -0,0 +1,142 @@
...
+}
+
+impl LegacyTokenGuard {
+    pub(crate) fn new(inner: Arc<dyn spark_wallet::SessionStore>) -> Self {
+        Self { inner }
+    }
+}
+
+#[macros::async_trait]
+impl spark_wallet::SessionStore for LegacyTokenGuard {
```
roeierez: Would be cleaner If we can avoid adding this wrapper. I think better to have a good recovery in case the session token is invalid.

roeierez (changes requested):
Looks good! Still reviewing. added a comment after one pass.

crates/breez-sdk/core/src/persist/backend/mod.rs
```diff
@@ -66,6 +66,32 @@ pub fn custom_storage(storage: Arc<dyn Storage>) -> Arc<dyn StorageBackend> {
...
+/// [`SdkBuilder::with_session_store`](crate::SdkBuilder::with_session_store),
+/// keeping the backend's persistence. A typical use is at-rest encryption (the
+/// SDK does not encrypt tokens itself): wrap it in a store that encrypts on
+/// write and decrypts on read. `identity` is the wallet identity public key
+/// bytes (the same value the SDK derives from the signer).
+#[cfg_attr(feature = "uniffi", uniffi::export(async_runtime = "tokio"))]
+pub async fn default_session_store(
+    backend: Arc<dyn StorageBackend>,
+    network: Network,
+    identity: Vec<u8>,
```
dangeross, line 83: Doesn't having to provide the `identity` cause a chicken-and-egg situation with a Turnkey signer? We can't use `default_session_store()` until we've done an SDK connect loop one time.

danielgranhao: Can't the identity be fetched from signer before providing the signer to SDK? Something like:

```
const signers = await createTurnkeySigner(config)          // no config.identityPublicKey yet
const identity = await signers.sparkSigner.getIdentityPublicKey()
const store = await defaultSessionStore(backend, network, identity)  // now scoped correctly
const sdk = await connectWithSigner({ ..., signers, sessionStore: encWrap(store) })
```

roeierez: Do you think we can cache it internally by turnkey config? So user initialization doesn't change but before creating the account we lookup for the identity public key by the org id, wallet id and account number?

crates/breez-sdk/core/src/session_store/legacy_guard.rs
danielgranhao: Done in [be09b72](https://github.com/breez/spark-sdk/pull/982/commits/be09b7264b011259bc65ce7d2ad6a643ebd70697)

roeierez (approved):
LGTM

crates/breez-sdk/core/src/persist/backend/mod.rs
danielgranhao: In an offline discussion, we agreed it's fine for the integrator to handle providing the identity pubkey. Caching would require having storage on the turnkey constructor.

docs/breez-sdk/snippets/wasm/sdk_building.ts
```diff
@@ -225,3 +230,41 @@ const exampleWithPaymentObserver = (builder: SdkBuilder): SdkBuilder => {
...
+  config.apiKey = '<breez api key>'
+
+  // Get the backend's own session store, then wrap it to encrypt at rest.
+  // identity is the wallet identity public key, hex.
+  const storageConfig = defaultStorage('./.data')
+  const inner = await defaultSessionStore(storageConfig, 'mainnet', identity)
+  const sessionStore = new EncryptingSessionStore(inner)
+
+  let builder = SdkBuilder.new(config, seed)
+  builder = builder.withStorageBackend(storageConfig)
```
dangeross, line 266: Within this span `storageConfig` is used twice: `defaultSessionStore` (line 262) consumes it (wasm-bindgen passes `WasmStorageConfig` by value and zeroes the handle), so `withStorageBackend(storageConfig)` (line 266) then throws `null pointer passed to rust` at runtime. Confirmed against the built node package. Snippet CI is `tsc` + eslint with no execution, so it stays green while the example crashes.

crates/spark/src/operator/rpc/spark_rpc_client.rs
```diff
@@ -600,51 +604,88 @@ impl SparkRpcClient {
...
+    /// gRPC client from it, then issue exactly one request. It is invoked at most
+    /// twice. For streaming RPCs only the stream open is covered; an auth failure
+    /// surfacing while the stream is polled is not retried here.
+    ///
+    /// Concurrent refreshes for the same operator are not coalesced: several
+    /// in-flight calls rejected at once each re-authenticate independently. This
+    /// only happens when a token is server-rejected (rare, and self-correcting as
+    /// each stores a fresh valid token), so the redundant round-trips do not
+    /// justify a single-flight lock on the auth hot path.
+    async fn call_with_auth_retry<T, F, Fut>(&self, call: F) -> Result<T>
```
dangeross, line 644: The operator half of this recovery is unverified (no test module here, and unlike the SSP side the transport is a concrete tonic channel, so no easy unit mock). Feasible itest: build with a custom `SessionStore` (via `with_session_store`) pre-seeded with a garbage token at a far-future expiration, then make an operator call and assert it succeeds. That covers the force-refresh end to end, doubles as the only coverage for `with_session_store`, and confirms the operator returns `Unauthenticated` for a bad token, which the self-heal depends on.

dangeross (commented):
Looks good. Just test coverage / wasm usage comments

docs/breez-sdk/snippets/wasm/sdk_building.ts
danielgranhao: Good catch. Fixed by adjusting wasm bindings to take storageConfig by ref

crates/spark/src/operator/rpc/spark_rpc_client.rs
danielgranhao: Good idea. Done.

dangeross (approved):
LGTM

PR #983, opened 2026-07-09
client signing docs

roeierez:


docs/breez-sdk/snippets/rust/src/client_signing.rs
```diff
@@ -0,0 +1,175 @@
...
+            }
+        }
+        UnsignedTransferPackage::Token {
+            prepare_token_transaction,
+            token_identifier,
+            amount,
+            fee,
+            ..
+        } => {
+            info!("Approve sending {amount} of token {token_identifier} (fee {fee})");
```
dangeross: When this Token package is a consolidation (`is_swap = true`), `build_token_package()` sets `amount = 0` and `fee = 0`, so this line prints "Approve sending 0 of token X (fee 0)" for what is actually a fund-reshaping step, not a send. The sats path just above distinguishes exactly this case with its own `Swap` message, but the Token arm drops `is_swap` via `..` and always frames it as a send.

The Tokens section of the guide documents what `is_swap` means, so a user hitting this path is signing something the prompt describes wrong. Should the snippet branch on `is_swap` here and show a "Consolidating token outputs" message, matching the sats flow? Same pattern in the other 8 language snippets.

dangeross (approved):
LGTM

danielgranhao (approved):
LGTM

PR #984, opened 2026-07-09
docs: Breez SDK - Spark theme reskin (theme-only)

dsbreez:
Update the mdBook chrome with the Breez SDK design: navy default theme, fixed brand header, design-mockup code panels and syntax palette, vendored JetBrains Mono (OFL).

No Markdown content, preprocessor, or build-workflow changes.


docs/breez-sdk/theme/header.hbs
```diff
@@ -0,0 +1,43 @@
...
+    <a href="https://t.me/breezsdk" target="_blank" rel="noopener" aria-label="Telegram">
+      <img src="{{ path_to_root }}assets/icons/telegram-icon.svg" alt="">
+    </a>
+  </nav>
+
+  <!-- brand-header.js relocates #sidebar-toggle, #theme-toggle, #theme-list and
+       #search-toggle into this container so they read as one bar. -->
+  <div id="brand-controls" class="brand-controls"></div>
+
+  <a class="brand-cta" href="https://breez.technology/request-api-key/" target="_blank" rel="noopener">
```
dangeross: Does this need the hash id to jump to the form?
https://breez.technology/request-api-key/#contact-us-form-sdk

dsbreez: fixed!

dangeross (approved):
LGTM

erdemyerebasmaz (changes requested):
Yay @dsbreez 👏 

Deployed and tested on my machine and this is looking good! 

A few things to address, and a couple notes:

- Search bar needs top padding and has to be aligned to main content. The search results need to be contained to the width of the search bar. They should not be behind right navigation side menu.

- If the docs is opened on a small browser window, the side menus can be hidden and they do not re-appear after resizing to full screen. It requires a refresh on a larger sized window for them to be visible again. We need an always visible toggle, like the previous `sidebar-toggle` to show/hide the left side menu.

- This is not a regression but, sidebar resize handle does not work at all, it can be removed or it can be made functional to expand and shrink the sidebar within a reasonable limit, like up to 1/3rd of screen.

- I also think clicking Breez logo should open the docs considering we already have a Home button to go to Breez website.

erdemyerebasmaz (changes requested):
Search bar changes are great!

Left side menu still does not re-appear automatically after the window is resized, and I'm not sure about the toggle being on the right side on the top bar. It should be somewhere on the left side imo, maybe to the left of Breez logo?

The resize handle works great, but maybe it should just hide the side menu after it's shrinked below 1/6th of screen, or 200px, so it's content is still legible.

erdemyerebasmaz (approved):
LGTM!! 👏

PR #986, opened 2026-07-10
LNURL per-domain partner attribution for lightning-address receives

dangeross:
Lightning-address receives are not attributed to a partner when created by the LNURL server.

### Change
Attribute each server-created invoice to its domain's Spark partner:

- **Schema**: a dedicated `domain_attribution` table holding the admin-set `api_key` and the server-cached `jwt`.
- **Per-request attribution**: every receive creates the invoice with a short-lived, background-less `SparkWallet` bound to the request's domain (sharing the process signer, pre-warmed session, HTTP client, and connection pool), so its SSP receive call carries the right `x-partner-jwt`. Its provider serves the domain's own partner JWT, or the default key's when the domain has none (or its own token isn't cached yet). A build failure falls back to the shared wallet, so attribution never fails a receive.
- **JWT cache** `JwtCache` + `JwtHeaderProvider`: a shared background hydrator keeps a token warm for every domain with its own api key (per-domain backoff + fetch timeout, api-key-rotation aware) and one for the default key, persisting the per-domain tokens so restarts and sibling instances start warm. The provider's `headers()` is **serve-only**: it reads the in-memory cache, never does I/O, and never returns `Err`, so it can't delay or fail an invoice.
- **Default API key** `--default-api-key` / `BREEZ_LNURL_DEFAULT_API_KEY`: covers any domain without its own api key (and any whose own token isn't cached yet), so no receive is left unattributed; such domains get a startup warning. The shared fallback wallet carries this key's JWT too.
- **Mainnet only**: the JWT endpoint is mainnet-only, so regtest/testnet skip attribution.

### Configuring
Allowlist the domain, then set its API key in the attribution table (picked up within ~1 min, no restart):

```sql
INSERT INTO domain_attribution (domain, api_key) VALUES ('example.com', '<breez-api-key>')
ON CONFLICT (domain) DO UPDATE SET api_key = excluded.api_key, jwt = NULL;
```

Clearing `jwt` on an api-key change forces a fresh fetch for the new key. Leaving a stale cached `jwt` in place would keep attributing to the old partner until that token expires.


crates/breez-sdk/lnurl/src/partner_jwt.rs
```diff
@@ -0,0 +1,631 @@
...
+                continue;
+            };
+            // Cache the persisted token if not yet expired
+            if let Some(exp) = jwt_exp(&token)
+                && now < exp
+            {
+                self.cache.write().await.insert(
+                    domain,
+                    CachedToken {
+                        api_key: api_key.clone(),
```
danielgranhao, line 169: `load_persisted` can't detect a key rotated during downtime. The persisted JWT is tagged here with the domain's current key, but the stored token contains no record of which key minted it. If a domain's api key is rotated while this instance is down, on restart, the stale token is tagged with the new key, passes the same-key freshness gate in hydrate_once, and keeps attributing to the old partner until it expires (never refetched). The live rotation path handles this correctly via the new api_key field; only this warm-start path is blind to it.

It's pretty narrow (rotation has to coincide with downtime and a persisted token), and it self-heals at expiry, so I'm fine with accepting it. To fully close it, we could persist the minting key alongside the JWT and warm the cache only when it matches the current key.

crates/breez-sdk/lnurl/src/partner_jwt.rs
```diff
@@ -0,0 +1,631 @@
...
+                "Bearer session".to_string(),
+            )]))
+        }
+    }
+
+    #[tokio::test]
+    async fn composes_with_real_combined_header_provider() {
+        use spark::header_provider::CombinedHeaderProvider;
+
+        let provider = test_provider(&[("a.com", Some("key-a"))]);
```
danielgranhao: The whole feature relies on `CURRENT_DOMAIN` surviving to `headers()` in the real SSP request, but that's only asserted directly by CombinedHeaderProvider, not through the wallet's request dispatch. This depends on there being no task-spawn between with_domain and the header fetch, an implementation detail of the spark crate with no guard here. If the SSP client is ever refactored to dispatch on a pooled/spawned task, attribution breaks silently with zero test failures. Consider a test that points the wallet's SSP client at a local axum endpoint (same pattern as hydrate_fetches_caches_serves_and_persists) and asserts the inbound request carries x-partner-jwt to avoid future regressions.

crates/breez-sdk/lnurl/src/sqlite/repository.rs
```diff
@@ -855,3 +872,63 @@ impl crate::webhooks::WebhookRepository for LnurlRepository {
             .map_err(|e| WebhookRepositoryError::General(e.into()))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::LnurlRepository as SqliteRepository;
+    use crate::repository::LnurlRepository;
+    use sqlx::sqlite::SqlitePoolOptions;
+
```
danielgranhao, line 884: Can we move this to shared_tests and run it for postgres as well?

danielgranhao (commented):
Looking good

crates/breez-sdk/lnurl/src/partner_jwt.rs
dangeross: Added `ssp_request_carries_partner_jwt_for_scoped_domain` test

crates/breez-sdk/lnurl/src/sqlite/repository.rs
dangeross: Added to `shared_tests`

crates/breez-sdk/lnurl/src/partner_jwt.rs
dangeross: Going to leave as is as it's a pretty slim case

crates/breez-sdk/lnurl/migrations/postgres/20260709000000_domain_attribution.sql
```diff
@@ -0,0 +1,6 @@
+-- Per-domain Spark partner attribution, colocated on the allowlist row:
+--   api_key: the Breez API key (admin-set), exchanged for a partner JWT.
+--   jwt:     the cached partner JWT (server-written), so restarts and sibling
+--            instances start warm instead of re-fetching every token.
+ALTER TABLE allowed_domains ADD COLUMN api_key TEXT;
+ALTER TABLE allowed_domains ADD COLUMN jwt TEXT;
```
JssDWt: I would prefer the api key / jwt to go in a dedicated table.

crates/breez-sdk/lnurl/src/partner_jwt.rs
```diff
@@ -0,0 +1,715 @@
...
+
+/// Shared map of allowed domains to their (optional) Breez API key, kept in
+/// sync with the DB by the `domains` refresher.
+pub type DomainKeys = Arc<RwLock<HashMap<String, Option<String>>>>;
+
+tokio::task_local! {
+    /// The domain whose invoice is currently being created. Set by
+    /// [`with_domain`] and read by [`PerDomainJwtHeaderProvider::headers`].
+    static CURRENT_DOMAIN: String;
+}
```
JssDWt: I think this part is crossing some complexity boundaries that should be unnecessary. Perhaps a spark wallet per invoice request is a solution? With a header provider per domain? So you can pick the right provider based on the current domain then?

dangeross: Yeah, I was struggling with this a bit too. I think the spark wallet can be made lightweight enough to do per request handling.

crates/breez-sdk/lnurl/src/state.rs
```diff
@@ -24,11 +24,44 @@ pub struct State<DB> {
...
+            c.provider_for(domain.to_string()) as Arc<dyn spark::header_provider::HeaderProvider>
+        });
+        let wallet = spark_wallet::SparkWallet::new(
+            self.spark_config.clone(),
+            Arc::new(spark_wallet::SparkSignerAdapter::new(self.signer.clone())),
+            self.session_store.clone(),
+            Arc::new(spark::tree::InMemoryTreeStore::default()),
+            Arc::new(spark::token::InMemoryTokenOutputStore::default()),
+            Arc::clone(&self.connection_manager),
+            None,
```
roeierez, line 59: This seems to create ssp_http_client on every wallet creation (request). Better to reuse it then it is truly light wallet creation.

roeierez (changes requested):

crates/breez-sdk/lnurl/src/main.rs
```diff
@@ -243,30 +299,16 @@ where
             Arc::new(InMemoryTreeStore::default()),
             Arc::new(InMemoryTokenOutputStore::default()),
             Arc::clone(&connection_manager),
+            Some(Arc::clone(&ssp_http_client)),
             None,
-            None,
-            None,
+            default_provider,
             None,
```
roeierez: Seems like the we don't use the default jwt in so requests?

crates/breez-sdk/lnurl/src/routes.rs
```diff
@@ -754,8 +767,32 @@ where
...
+                Err(e) => {
+                    warn!(
+                        "failed to build attributed wallet for '{domain}', using the shared wallet: {e}"
+                    );
+                    std::sync::Arc::clone(&state.wallet)
+                }
+            }
+        } else {
+            std::sync::Arc::clone(&state.wallet)
+        };
```
roeierez: IIUC the state wallet is only used as a fallback when no  api key is provided. Perhaps it is better to get rid of the shared wallet and add a fallback in the header providers to return the result of the default provider in case of no api key? 
Apart of being simple to reason about it serves as a better way to ensure we will always use a jwt in the request.

roeierez (changes requested):

crates/breez-sdk/lnurl/src/routes.rs
```diff
@@ -696,6 +695,20 @@ where
...
+    /// the LNURL-pay response. A domain with its own api key creates the invoice
+    /// with a per-request wallet bound to that key, so its SSP receive call
+    /// carries the domain's `x-partner-jwt`; other domains use the shared wallet.
+    #[allow(clippy::too_many_lines)]
+    async fn handle_invoice_inner(
+        state: State<DB>,
+        params: LnurlPayCallbackParams,
+        username: String,
+        domain: String,
+    ) -> Result<Json<Value>, (StatusCode, Json<Value>)> {
```
danielgranhao: nit: we can revert this now

crates/breez-sdk/lnurl/src/partner_jwt.rs
```diff
@@ -0,0 +1,797 @@
...
+        backoff: &mut Backoff,
+        now: u64,
+    ) -> Option<CachedToken> {
+        let fresh = current_exp.is_some_and(|exp| now.saturating_add(REFRESH_LEAD_SECS) < exp);
+        if fresh || now < backoff.next_retry {
+            return None;
+        }
+        match fetch_jwt(&self.http, &self.jwt_url, api_key).await {
+            Ok(token) => {
+                let exp = jwt_exp(&token).unwrap_or(0);
```
danielgranhao: Could we treat an unparseable expiry as a fetch failure instead?

crates/breez-sdk/lnurl/src/partner_jwt.rs
```diff
@@ -0,0 +1,797 @@
...
+                *backoff = Backoff::default();
+                Some(CachedToken {
+                    api_key: api_key.to_string(),
+                    token,
+                    exp,
+                })
+            }
+            Err(e) => {
+                // The api key is a secret, so it is not logged.
+                warn!("could not fetch a partner JWT: {e}");
```
danielgranhao: Could we log the domain or api key here?

crates/breez-sdk/lnurl/src/postgresql/repository.rs
```diff
@@ -843,3 +864,60 @@ impl crate::webhooks::WebhookRepository for LnurlRepository {
...
+        let pool = sqlx::PgPool::connect(&url).await.ok()?;
+        crate::postgresql::run_migrations(&pool).await.ok()?;
+        sqlx::query("DELETE FROM domain_attribution")
+            .execute(&pool)
+            .await
+            .ok()?;
+        sqlx::query("DELETE FROM allowed_domains")
+            .execute(&pool)
+            .await
+            .ok()?;
```
danielgranhao: Let's replace the `?` in all these statements with `expect` otherwise any issue with test infra will result in test passing even when `LNURL_TEST_POSTGRES_URL` is set

crates/breez-sdk/lnurl/src/routes.rs
```diff
@@ -696,6 +695,20 @@ where
 
         let username = sanitize_username(&identifier);
         let domain = sanitize_domain(&state, &host).await?;
+        Self::handle_invoice_inner(state, params, username, domain).await
+    }
+
+    /// Create the lightning-address invoice for `username` on `domain` and build
+    /// the LNURL-pay response. A domain with its own api key creates the invoice
+    /// with a per-request wallet bound to that key, so its SSP receive call
+    /// carries the domain's `x-partner-jwt`; other domains use the shared wallet.
```
danielgranhao: Doc is out of date but if we inline `handle_invoice_inner` as suggested in another comment, that will be resolved

crates/breez-sdk/lnurl/src/partner_jwt.rs
```diff
@@ -0,0 +1,797 @@
...
+        self.0
+            .list_domains()
+            .await
+            .map(|domains| {
+                domains
+                    .into_iter()
+                    .filter_map(|d| d.jwt.map(|jwt| (d.domain, jwt)))
+                    .collect()
+            })
+            .unwrap_or_default()
```
danielgranhao: Could we warn on failure?

crates/breez-sdk/lnurl/src/partner_jwt.rs
```diff
@@ -0,0 +1,797 @@
...
+const HYDRATE_INTERVAL: Duration = Duration::from_secs(30);
+/// Bound on a single JWT fetch, so a slow endpoint can't stall the hydrator.
+const FETCH_TIMEOUT: Duration = Duration::from_secs(10);
+/// Cap on the per-target fetch backoff.
+const BACKOFF_MAX_SECS: u64 = 5 * 60;
+
+/// Shared map of allowed domains to their own Breez API key, or `None` when
+/// the domain has none and falls back to the default. Kept in sync with the
+/// DB by the `domains` refresher.
+pub type DomainKeys = Arc<RwLock<HashMap<String, Option<String>>>>;
```
danielgranhao: Could it be `Arc<RwLock<crate::domains::DomainMap>>` to avoid restating `DomainMap`?

danielgranhao (commented):
LGTM, just nits

danielgranhao (approved):

roeierez (approved):
LGTM

crates/breez-sdk/lnurl/src/main.rs
```diff
@@ -45,6 +51,10 @@ mod webhook_notify;
 mod webhooks;
 mod zap;
 
+fn default_user_agent() -> String {
+    concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION")).to_string()
```
JssDWt, line 55: Do these have to be set at runtime? Should we use built_info like in other crates?

JssDWt (approved):
LGTM, potential nit

crates/breez-sdk/lnurl/src/main.rs
dangeross: These are compile time variables added by cargo.

I did start with built + git2 here to match the other crates but dropped it. The reason those use `built` is for GIT_VERSION, but that doesn't survive the fly deploy: the builder image doesn't have git installed, and src is a submodule + src/.git is in .dockerignore, so there's no repo in the build context. `built` would just fall back to PKG_VERSION and give the same "lnurl/0.1.0", while pulling libgit2-sys into every docker build for nothing.

I tested it to be sure: even with git installed + the submodule + built/git2 running in the docker build, GIT_VERSION comes out None. So env! gets us the same version without the build time git dep.

PR #988, opened 2026-07-10
Tighten JWT handling

dangeross:
Tightens 3 cases where the partner JWT is missing from requests:

### 1. Persist the JWT across restarts

The provider was built with `storage: None`, leaving its persistence code inactive: the JWT was re-fetched from scratch on every launch, so clients had an unattributed cold-start window on each start.

Construction is now two-phase: `new()` to initialize and `start(storage)` to bind storage then spawn the refresh task. `SdkBuilder::build` binds storage only with no provided shared context; a shared server context stays in-memory. Clients now warm-start from cache and persist refreshes.

### 2. Resolve composed headers sequentially

`CombinedHeaderProvider` resolved `[auth, jwt]` with `try_join_all` (parallel), so the best-effort JWT was snapshotted immediately and gained nothing from the auth handshake latency. It is now sequential: the auth provider's first call performs a challenge/response handshake, so running it first gives the background JWT fetch that round-trip to land before its header is read. The JWT header read is a cache only call so adds little latency.

### 3. Serve a valid token through the refresh window

`headers()` gated on `is_expired` (the 5-minute refresh grace), so a still-valid token was withheld for its last 5 minutes, emitting no header during every refresh cycle. Serve is now split from refresh: `is_past_serve_expiry` withholds
only within a 30-second clock-skew margin of real `exp`; the refresh loop still fires 5 minutes early (unchanged).

### DIFF
--- a/crates/breez-sdk/core/src/jwt_header_provider.rs
+++ b/crates/breez-sdk/core/src/jwt_header_provider.rs
@@ -1,5 +1,5 @@
 use std::collections::HashMap;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex, OnceLock};
 
 use base64::{Engine, engine::general_purpose::URL_SAFE_NO_PAD};
 use breez_sdk_common::utils::now;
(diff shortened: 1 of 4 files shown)


crates/breez-sdk/core/src/jwt_header_provider.rs
```diff
@@ -319,6 +361,35 @@ mod tests {
...
+    fn test_is_past_serve_expiry_far_future() {
+        // Well beyond the skew margin: servable.
+        assert!(!is_past_serve_expiry(u64::from(now()) + 3600));
+    }
+
+    #[test]
+    fn test_is_past_serve_expiry_within_refresh_grace_still_served() {
+        // Expires in 2 minutes: inside the 5-minute refresh lead but outside the
+        // 30s serve margin, so it must still be served. This is the case the old
+        // grace check dropped, and is the core regression guard for this fix.
```
roeierez: This comment referring to the history lead to the fix. Better change or remove the history part.

roeierez (approved):
LGTM with one small NIT.

danielgranhao (approved):
LGTM

PR #989, opened 2026-07-13
perf(transfer): Coalesce concurrent claims per transfer.

mhrheaume:
The wallet drives transfer claims from several unsynchronized triggers (incoming-transfer event, stream reconnect, periodic sync, explicit API call, and counter-swap). With no per-transfer guard, overlapping triggers submit identical claims for the same transfer within ~50ms and race each other at the coordinator: one wins and the losers get spurious errors after doing a full verify -> sign -> finalize round of wasted work.

Add a single-flight set on TransferService keyed by transfer id. claim_transfer reserves a slot on entry and returns Ok(vec![]) when a claim for that transfer is already in flight. An RAII guard releases the slot on completion, so a failed claim is retried by the next trigger. This sits at the one chokepoint every claim path passes through, covering both the legacy and package-based finalize RPCs.

### DIFF
--- a/crates/spark/src/services/transfer.rs
+++ b/crates/spark/src/services/transfer.rs
@@ -1,5 +1,8 @@
 use std::time::Duration;
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::{HashMap, HashSet},
+    sync::{Arc, Mutex},
+};
 
 use crate::Network;
 use crate::address::SparkAddress;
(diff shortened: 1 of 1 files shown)


danielgranhao:
Thanks a lot for the PR. Closing in favor of https://github.com/breez/spark-sdk/pull/991 which coalesces triggers and also adds some improvements for multi-instance races

PR #991, opened 2026-07-13
fix(transfer): Coalesce concurrent transfer claims per transfer

danielgranhao:
## Context

A Spark maintainer reported the SDK submitting the same transfer claim 2-4 times
concurrently (sibling claims ~50ms apart). One claim wins; each duplicate runs a
full verify / sign / finalize round that fails at the coordinator. The root cause
is that claims have no in-flight guard, and in a single SDK instance up to three
independent tasks can claim the same transfer concurrently:

- the wallet event loop, which claims on incoming-transfer events and runs the
  stream-reconnect sweep (those two are serialized with each other),
- the SDK sync loop, which runs the periodic and on-demand sync sweeps (also
  serialized with each other), and
- the wait-for-payment pollers (one claimer per concurrent wait call).

Supersedes #989.

## Changes

`TransferService` now holds a per-transfer claim lock. A trigger that finds a
claim already in flight waits for it and adopts its outcome by reading the
finalized leaves off the coordinator, so every caller gets the real leaves
(instead of the draft's `Ok(vec![])`). If the in-flight claim failed, the
waiter claims the transfer itself while still holding the lock, so a failed
claim is retried by the next trigger immediately.

## Notes

- Races between separate instances of the same wallet are out of scope here:
  they are already handled by the pre-existing fallback that adopts the
  coordinator's finalized leaves when a failed claim turns out to be completed.
- Claims for different transfers are unaffected: locks are per transfer id, and
  a transfer's lock entry is dropped once its last claim finishes.

### DIFF
--- a/crates/spark/src/services/transfer.rs
+++ b/crates/spark/src/services/transfer.rs
@@ -1,3 +1,4 @@
+use std::sync::{Mutex, MutexGuard, PoisonError};
 use std::time::Duration;
 use std::{collections::HashMap, sync::Arc};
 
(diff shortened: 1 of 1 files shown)


crates/spark/src/services/transfer.rs
```diff
@@ -485,13 +570,59 @@ impl TransferService {
...
+    ) -> Result<Vec<TreeNode>, ServiceError> {
+        let _permit = match claim_lock.lock.try_lock() {
+            Ok(permit) => permit,
+            Err(_) => {
+                debug!(
+                    "Claim already in flight for transfer {}, waiting for it",
+                    transfer.id
+                );
+                let permit = claim_lock.lock.lock().await;
+                if let Some(leaves) = self.finalized_leaves_if_already_claimed(&transfer.id).await {
```
roeierez: Do you think we can save the network request for the transfer if we set the transfer result on the claim guard? And all other waiters take it from there?

danielgranhao: Done

dangeross (approved):
LGTM

roeierez (approved):
LGTM

PR #992, opened 2026-07-14
Unilateral exit in the sdk

JssDWt:
Replaces #795

Expose unilateral exit from the breez sdk. With a user experience that I would call 'as good as it gets'. 

`prepare_unilateral_exit` selects leaves for you to exit and gives you a quote, so you have an idea on what the exit will cost. 
`unilateral_exit` takes the response from `prepare_unilateral_exit`, utxos that fund the CPFP transactions, and a signer to sign for those CPFP outputs. So unilateral exit will always be called with a specific set of leaves (picked by `prepare_unilateral_exit`)

## Notes

- Building an exit still requires the operators to be online. Recovery with the operators offline, from local state, will be handled in a follow-up (very soon). 
- User brings their own utxos to fund the CPFP fees.
- Exiting several leaves at once starts with a fan-out transaction that splits one funding UTXO into one output per branch. Shared tree ancestors are broadcast once, and all refunds are collected into a single sweep. Funding one UTXO per branch (`per_branch_funding`) skips the fan-out and its fee.
- `LeafSelection::Auto` keeps only leaves whose value exceeds their own marginal exit cost; `Specific` exits exactly the leaves you name. When nothing is worth exiting, the quote comes back empty.
- A `CpfpSigner` is injected to sign the CPFP outputs
- A final sweep sends the recovered funds to a destination Bitcoin address you supply.
- Every CPFP change output pays back to the first funding UTXO's address, so that one address is reused across the exit.
- The SDK builds and signs the set and returns it, the caller broadcasts. Each `UnilateralExitTransaction` carries its `depends_on` txids and `csv_timelock_blocks` so the caller can order and time the broadcasts.
- Every call inspects the current CPFP chain state on-chain and picks up from wherever the exit currently is, instead of restarting. It rediscovers each confirmed node's on-chain CPFP change output and funds the next step from it, and it recognises an already-broadcast fan-out from chain state, so no earlier transaction is ever passed back in.
- Re-running with a higher `fee_rate_sat_per_vbyte` rebuilds the not-yet-confirmed transactions to replace the earlier ones by RBF; already-confirmed steps are left as they are.

Moved bitcoind for the integration tests to `bitcoin/bitcoin:31.0` for its 1P1C package relay and TRUC policy. I verified that image contains bitcoind byte-equal with the release on bitcoincore.org.


crates/breez-sdk/wasm/src/chain_service.rs
```diff
@@ -31,6 +31,16 @@ impl BitcoinChainServiceHandle {
...
 
+    #[wasm_bindgen(js_name = "getAddressFundedTxoCount")]
+    pub async fn get_address_funded_txo_count(&self, address: String) -> Result<JsValue, JsValue> {
+        let count = self
+            .inner
+            .get_address_funded_txo_count(address)
+            .await
+            .map_err(|e| JsValue::from_str(&e.to_string()))?;
+        serde_wasm_bindgen::to_value(&count).map_err(|e| JsValue::from_str(&e.to_string()))
+    }
```
dangeross: Missing a `get_outspend` fn?

docs/breez-sdk/src/guide/unilateral_exit.md
```diff
@@ -0,0 +1,154 @@
...
+
+{{#tabs unilateral_exit:unilateral-exit}}
+
+### The signer
+
+The CPFP children and the fan-out spend your funding UTXOs, so they have to be signed. The SDK does not hold your funding keys; it hands each unsigned transaction to a signer you provide.
+
+The built-in single-key signer covers the common case: it signs P2WPKH and P2TR inputs from one secret key. For {{#enum CpfpInput::P2tr}} funding, pass the **internal, untweaked (BIP86)** key, not the tweaked on-chain output key: the tweaked key derives a scriptPubKey that does not match the UTXO, so the transaction is rejected at broadcast. For anything else (a multisig, a hardware wallet, or keeping key material out of the SDK entirely) implement the `CpfpSigner` interface and describe the funding with {{#enum CpfpFundingKind::Custom}} (in the quote) and {{#enum CpfpInput::Custom}} (in the build). Those carry the funding `script_pubkey_hex` and an upper-bound `signed_input_weight` so the fee stays exact for any witness program. The signer receives a serialized PSBT, signs the inputs that are not already finalized, and returns the serialized signed PSBT:
```
dangeross:
```suggestion
The built-in single-key signer covers the common case: it signs P2WPKH and P2TR inputs from one secret key. For {{#enum CpfpInput::P2tr}} funding, pass the **internal, untweaked (BIP86)** key, not the tweaked on-chain output key: the tweaked key derives a scriptPubKey that does not match the UTXO, so the transaction is rejected at broadcast. For anything else (a multisig, a hardware wallet, or keeping key material out of the SDK entirely) implement the `CpfpSigner` interface and describe the funding with {{#enum CpfpFundingKind::Custom}} (in the quote) and {{#enum CpfpInput::Custom}} (in the build). Those carry the funding {{#name script_pubkey_hex}} and an upper-bound {{#name signed_input_weight}} so the fee stays exact for any witness program. The signer receives a serialized PSBT, signs the inputs that are not already finalized, and returns the serialized signed PSBT.
```

docs/breez-sdk/src/guide/unilateral_exit.md
```diff
@@ -0,0 +1,154 @@
...
+
+Whichever signer you use, the funding inputs must be **native SegWit** (a witness-program script; P2WPKH or P2TR with the built-in signer, any other witness program with a custom one). The exit refers to each transaction by an id it computes before signing, which only stays stable when the signature lives in the witness (native SegWit) rather than in the input script; legacy scripts are rejected, so your signer only ever has to sign native SegWit inputs.
+
+{{#tabs unilateral_exit:custom-cpfp-signer}}
+
+<div class="warning">
+<h4>Flutter</h4>
+Flutter cannot pass a foreign <code>CpfpSigner</code>, so it exposes two exit calls. {{#name unilateral_exit}} takes the funding secret key bytes and uses the built-in single-key signer. {{#name unilateral_exit_with_signer}} takes a <code>signPsbt</code> callback that receives the serialized PSBT, signs the inputs that are not already finalized (any scheme), and returns the serialized signed PSBT.
```
dangeross: I think stick to dart syntax as its Flutter specific
```suggestion
Flutter cannot pass a foreign <code>CpfpSigner</code>, so it exposes two exit calls. <code>unilateralExit</code> takes the funding secret key bytes and uses the built-in single-key signer. <code>unilateralExitWithSigner</code> takes a <code>signPsbt</code> callback that receives the serialized PSBT, signs the inputs that are not already finalized (any scheme), and returns the serialized signed PSBT.
```

crates/breez-sdk/core/src/chain/mod.rs
```diff
@@ -71,6 +77,79 @@ pub struct RecommendedFees {
     pub minimum_fee: u64,
 }
 
+/// The spend status of a transaction output (Esplora `/tx/:txid/outspend/:vout`).
+///
+/// An enum so an unspent output carries no spurious `txid`/`vin` and a spent one
+/// always names its spender. It (de)serializes to the flat Esplora shape
+/// `{ "spent": bool, "txid"?, "vin"?, "status"? }`, so chain services (including
+/// foreign ones over WASM/UniFFI) keep returning that object.
```
dangeross: Is it needed? Seems more internal/Claude reasoning

crates/breez-sdk/core/src/chain/mod.rs
```diff
@@ -71,6 +77,79 @@ pub struct RecommendedFees {
     pub minimum_fee: u64,
 }
 
+/// The spend status of a transaction output (Esplora `/tx/:txid/outspend/:vout`).
```
dangeross: Do we need to specifically mention Esplora?

crates/breez-sdk/core/src/signer/single_key_signer.rs
```diff
@@ -0,0 +1,243 @@
...
+
+use crate::error::SignerError;
+
+use super::cpfp::CpfpSigner;
+
+/// A CPFP signer backed by a single private key. Signs P2WPKH and P2TR key-path
+/// inputs only; taproot script-path spends are not supported.
+#[cfg_attr(feature = "uniffi", uniffi::export)]
+#[allow(clippy::needless_pass_by_value)]
+pub fn single_key_cpfp_signer(
```
dangeross, line 20: I don't have a strong opinion, but would `default_cpfp_signer` be more consistent?

crates/breez-sdk/wasm/src/signer.rs
dangeross, line 1193:
```suggestion
    signLeafRefundSpend(leafId: ExternalTreeNodeId, sighash: Uint8Array): Promise<SchnorrSignatureBytes>;
    signFrost(jobs: ExternalFrostJob[]): Promise<ExternalFrostShareResult[]>;
```

dangeross, line 1326: Missing a `sign_leaf_refund_spend` fn?

dangeross (commented):
Progress so far

crates/spark-wallet/src/wallet.rs
```diff
@@ -1335,22 +1332,286 @@ impl SparkWallet {
...
+        for leaf_id in leaf_ids {
+            let Some(leaf) = tree_nodes.get(leaf_id).cloned() else {
+                // Offline sourcing from local storage is deferred; a leaf the
+                // operators omit stays absent and is reported un-exitable.
+                warn!("Requested leaf {leaf_id} not returned by the operators; cannot exit it");
+                continue;
+            };
+            // Clone so the future owns (not borrows) the tree service: a
+            // `&self` borrow across the await isn't `Send` and would reject the
+            // uniffi-exported async callers.
```
dangeross: Remove the Claude reasoning?

crates/spark-wallet/src/wallet.rs
```diff
@@ -1335,22 +1332,286 @@ impl SparkWallet {
...
+    pub async fn prepare_unilateral_exit_plan(
+        &self,
+        fee_rate_sat_per_kw: u64,
+        selection: ExitLeafSelection,
+        inputs: Vec<CpfpInput>,
+        destination_script_len: usize,
+    ) -> Result<PreparedUnilateralExit, SparkWalletError> {
+        let (leaf_ids, filter) = self.resolve_leaf_selection(selection).await?;
+        let tree_nodes = self.source_exit_tree_nodes(&leaf_ids).await?;
+        let plan = spark::services::plan_exit(
```
dangeross: nit: I think because `coop_exit::*` and `unilateral_exit::*` are flattened into the spark services module it might be clearer to name the unilateral types/fns more explicitly

crates/spark/src/services/unilateral_exit.rs
```diff
@@ -300,141 +803,122 @@ fn create_tx_cpfp_psbt(
...
-                vout: utxo.vout,
-            },
+    let total_input_value: u64 = funding_inputs
+        .iter()
+        .map(|i| i.witness_utxo.value.to_sat())
+        .fold(0u64, u64::saturating_add);
+    let change_script_pubkey = funding_inputs[0].witness_utxo.script_pubkey.clone();
+    let first_signed_input_weight = funding_inputs[0].signed_input_weight;
+
+    let rbf_sequence = Sequence(0xffff_fffd);
```
dangeross: Use `Sequence::ENABLE_RBF_NO_LOCKTIME` for consistency?

crates/xtask/src/main.rs
```diff
@@ -860,8 +866,14 @@ fn wasm_clippy_cmd(fix: bool, rest: Vec<String>) -> Result<()> {
...
     let sh = prepare_itest_images()?;
 
-    // Run the integration tests
-    cmd!(sh, "cargo test -p spark-itest --no-fail-fast").run()?;
+    // Run the integration tests with limited parallelism to avoid overwhelming
+    // Docker with too many concurrent container startups. Each test spins up
+    // its own bitcoind + operator containers.
+    cmd!(
+        sh,
+        "cargo test -p spark-itest -p breez-sdk-itest --no-fail-fast -- --test-threads=2"
```
dangeross: Is it worth separating out the breez-itest into a separate xtask command (and updating the makefile)?

crates/spark/src/services/unilateral_exit.rs
```diff
@@ -1,297 +1,800 @@
...
-            ));
-        }
+    let change_script = &inputs[0].witness_utxo.script_pubkey;
+    let params = LeafExitCostParams {
+        initial_cpfp_input_weight: Weight::from_wu(
+            inputs
+                .iter()
+                .map(|i| i.signed_input_weight)
+                .fold(0u64, u64::saturating_add),
+        ),
```
dangeross: This charges every branch's first CPFP child for the sum of all N supplied inputs, but each child spends only its own branch's funding (a single UTXO under the recommended `per_branch_funding`). `quote_exit` sizes it as one input, so a multi-leaf exit funded at exactly the quoted per-branch amounts is short per branch and fails with `InsufficientCpfpFunds`. The all-inputs weight fits the fan-out transaction's own fee, not the per-branch children. Should we be matching the quote's single-input sizing here?

(Claude found this, I need to double check)

roeierez: I think a test with exact funding will cover this case here.

crates/spark/src/services/unilateral_exit.rs
```diff
@@ -1,297 +1,800 @@
...
+    }
+    let mut remaining: Vec<&CpfpInput> = inputs.iter().collect();
+    remaining.sort_by(|a, b| {
+        b.witness_utxo
+            .value
+            .cmp(&a.witness_utxo.value)
+            .then_with(|| a.outpoint.cmp(&b.outpoint))
+    });
+    let mut sorted_leaves: Vec<&SelectedLeaf> = selected_leaves.iter().collect();
+    sorted_leaves.sort_by(|a, b| {
```
roeierez, line 675: Funding amounts are computed in value order (https://github.com/breez/spark-sdk/pull/992/changes#diff-8581e8357f8dbe03c226dd8c055459a8138112dee99762fe2dd76b2726399d10R481), but this sorts by cost. build_exit then charges each shared ancestor to the first branch in this order (spark-wallet/unilateral_exit.rs:914), so when the richest leaf isn't the costliest (depends also on the number of nodes in the branch), a branch gets underfunded and fails the dust check. Perhaps we shoiuld return per_branch_funding in value order.

roeierez (changes requested):

crates/breez-sdk/core/src/signer/single_key_signer.rs
JssDWt: I have the feeling this will not be the only cpfp signer we will create, and it will depend on the inputs you provide. For example, I can see a 'multiple key cpfp signer' for when each of your utxos has a different address.

crates/spark-wallet/src/unilateral_exit.rs
```diff
@@ -0,0 +1,2514 @@
...
+                        node_id: Some(node.id.clone()),
+                        txid: node_txid,
+                        csv_timelock_blocks: csv_timelock(&base_tx),
+                        base_tx,
+                        to_sign,
+                        depends_on,
+                        status,
+                    });
+                }
+                first_in_branch = false;
```
danielgranhao: Should this be moved up so it's inside the `if emitted.insert(node_txid)` block?

As is, `depends_on` omits the fan-out for every branch after the first because we flip `first_in_branch` even when a node is skipped. 

A unit test that covers fan-out and shared ancestor would also be useful for catching this.

crates/spark-wallet/src/wallet.rs
```diff
@@ -1335,22 +1332,286 @@ impl SparkWallet {
...
+        &self,
+        selection: ExitLeafSelection,
+    ) -> Result<(Vec<TreeNodeId>, LeafFilter), SparkWalletError> {
+        match selection {
+            ExitLeafSelection::Auto => {
+                let mut leaf_ids: Vec<TreeNodeId> = self
+                    .tree_service
+                    .list_leaves()
+                    .await?
+                    .available
```
danielgranhao, line 1399: Unless I'm missing something, `Auto` mode cannot resume a partially-exited leaf because it would be filtered out here

crates/breez-sdk/core/src/sdk/unilateral_exit.rs
```diff
@@ -0,0 +1,731 @@
...
+                build.refund_outputs,
+                build.cpfp_change_inputs,
+                destination,
+                fee_rate_sat_per_kw,
+            )
+            .await?;
+        let actual_sweep_fee = sweep_fee(&sweep_psbt);
+        let total_fee_sat = build_fee_sat.saturating_add(actual_sweep_fee);
+        let sweep_txid = sweep_psbt.unsigned_tx.compute_txid();
+        let sweep_status = tx_confirmation(chain, sweep_txid).await;
```
danielgranhao: For fresh exits a sweep will always be reported as `Unverified` because `chain.get_transaction_status` in `tx_confirmation` will 404. Maybe we can distinguish "not found" from a transport error and map "not found" to `Unconfirmed` instead?

crates/breez-sdk/core/src/sdk/unilateral_exit.rs
```diff
@@ -0,0 +1,731 @@
...
+        return Ok(serialize_hex(&psbt.extract_tx_unchecked_fee_rate()));
+    }
+    let out_bytes = signer
+        .sign_psbt(psbt.serialize())
+        .await
+        .map_err(|e| SdkError::Signer(format!("Sweep signer error: {e}")))?;
+    let out_psbt = bitcoin::Psbt::deserialize(&out_bytes)
+        .map_err(|e| SdkError::Generic(format!("Failed to deserialize signed sweep PSBT: {e}")))?;
+    Ok(serialize_hex(&out_psbt.extract_tx_unchecked_fee_rate()))
+}
```
danielgranhao, line 882: Not strictly necessary, but it would be nice to validate all inputs were signed otherwise that will only be caught at broadcast time

danielgranhao (commented):
Still going through it but just pushing what I have rn

crates/spark-wallet/src/unilateral_exit.rs
```diff
@@ -0,0 +1,2514 @@
...
+                }
+                return;
+            }
+            ChainResult::Unavailable => {
+                // Chain lookup failed. Fall back to operator status: OnChain =>
+                // assume cpfp-confirmed (skip the child, continue) so we don't
+                // double-spend an already-spent anchor. It can't tell cpfp from a
+                // direct/foreign confirmation; a wrong guess surfaces on the next
+                // lookup, and any refund is still adopted independently.
+                if node.status == TreeNodeStatus::OnChain {
```
roeierez, line 723: I am still wondering what we should do here...
Currently if chain service failed to query the node tx status we fallback to the operator. Not something will cause loosing funds, just invalid tx chain.
The alternative is to return error which not ideal as well.
Currently it at least gives the process a chance to complete. If I didn't miss anything then I think this is fine.

crates/spark-wallet/src/unilateral_exit.rs
```diff
@@ -0,0 +1,2514 @@
...
+
+    // Drop supplied inputs a prior run's CPFP child already spent. Only a confirmed
+    // spend counts: an unconfirmed spender is our own replaceable child (rebuilt via
+    // RBF on resume). Gated to tracked-change branches; skipped under a fan-out.
+    let mut spent_funding: HashSet<OutPoint> = HashSet::new();
+    if plan.fan_out_psbt.is_none() {
+        for (leaf_id, funding) in &plan.per_branch_funding {
+            if !branch_has_tracked_change(&node_map, leaf_id, &nodes) {
+                continue;
+            }
```
danielgranhao: If a node is treated as confirmed but its CPFP change couldn't be resolved (chain service down, operator says OnChain), the spent-funding check below is skipped. The rebuilt downstream children then spend the original funding UTXO, which the confirmed node's child already spent, so they're double-spends that can't be broadcast. Most of them still come back as plain `Unconfirmed`, so the caller has no hint the build is bad. 

Should this error out (or at least flag the affected txs) instead of silently returning them?

crates/breez-sdk/core/src/sdk/unilateral_exit.rs
```diff
@@ -0,0 +1,731 @@
...
+            round,
+            queries = queries.len(),
+            "resolve_exit_observations: round"
+        );
+        // Each query is answered exactly once (a failed lookup records
+        // `Unavailable`), so the loop always progresses and terminates.
+        for query in queries {
+            let result = execute_chain_query(chain, &query).await;
+            observed.push(Observation { query, result });
+        }
```
danielgranhao, line 554: Is it worth parallelizing this with some concurrency cap?

crates/spark-itest/src/fixtures/spark_so.rs
```diff
@@ -76,8 +76,22 @@ fn generate_self_signed_certificate(host_names: &[String]) -> Result<(String, St
...
 impl SparkSoFixture {
+    /// Stops every operator container (and its postgres backend), simulating a
+    /// full operator outage. bitcoind is untouched.
+    pub async fn stop(&self) -> Result<()> {
+        for op in &self.operators {
+            op.container.stop().await?;
+            op.postgres.stop().await?;
+        }
+        Ok(())
+    }
```
danielgranhao: This isn't used yet. I'm fine with keeping but consider dropping and only adding later when it gets used

crates/spark/src/services/unilateral_exit.rs
```diff
@@ -1,297 +1,800 @@
...
+}
+
+pub struct LeafExitCostParams {
+    /// Weight of the first CPFP child's inputs in a leaf's chain.
+    pub initial_cpfp_input_weight: Weight,
+    /// Weight of each subsequent child's single (chained-change) input.
+    pub single_cpfp_input_weight: Weight,
+    pub change_script_len: usize,
+    pub change_dust_limit: u64,
+    pub total_cpfp_budget: u64,
```
roeierez: Somehow it feels to me that the `total_cpfp_budget` doesn't belong here. Perhaps I get it wrong but seems like this field is used to validate enough balance at the end and it seems cleaner if `evaluate_leaf_exit_costs` will not mix the cost calculation and the affordability. One signal for it is that from the quote_exit we pass u64::MAX. One step for cost and then on the plan_exit when we have the real input budget we can actually validate the balance.

crates/spark/src/services/unilateral_exit.rs
JssDWt: Went for one input here, and at build time, report the missing amounts for each branch in an error, and aggregate the amounts in the error to report the right amount to the user.

crates/breez-sdk/core/src/sdk/unilateral_exit.rs
JssDWt: The most uncertain thing about this whole exit operation are the chain queries, because the chain service might be rate limited. For that reason I had not considered any parallelization of the chain service queries. Having in mind this is not performance-sensitive too. Then again if you would be rate limited, you would probably be rate limited regardless of parallelization here. You think we should do it?

roeierez: FWIW I prefer simplicity over speed on the unilateral exit flow. So I am ok with sequential queries.

crates/spark-wallet/src/wallet.rs
JssDWt: This now combines available and non-available leaves. What remains is to actually query non-available leaves from the operator at all times, which I will do in a followup.

crates/breez-sdk/core/src/sdk/unilateral_exit.rs
JssDWt: I don't think the transport error 404 should be 'unconfirmed' necessarily, because 404 doesn't always mean 'the transaction is not found'. However, I think the chain call is unnecessary here. It's unverified now if a parent is unverified, otherwise unconfirmed (since we're constructing it here).

crates/spark-wallet/src/unilateral_exit.rs
JssDWt: I do some better flagging for unverified now. I prefer to not error, to leave as much chance for a healthy exit as possible.

crates/spark/src/services/unilateral_exit.rs
```diff
@@ -1,297 +1,808 @@
...
+        // paid from the swept value, not this funding UTXO, so estimated_cost's
+        // sweep component (the quote's headroom) must not inflate the hard gate.
+        let required = selected[0].cpfp_cost.saturating_add(change_dust_limit);
+        let available = inputs
+            .iter()
+            .map(|i| i.witness_utxo.value.to_sat())
+            .fold(0u64, u64::saturating_add);
+        if available < required {
+            return Err(ServiceError::InsufficientCpfpBudget {
+                required_sat: required,
```
dangeross: I have a feeling because the cpfp cost is sized on the first input it can pass this check here, but later during the build the first child will get all inputs and a higher cost. The knock on effect being an insufficient budget (only in a narrow window, fee diff between 1 and N inputs on the first child).

Is it worth adding a single-leaf test with more than one UTXO?

crates/breez-sdk/cli/src/command/i_know_what_im_doing.rs
```diff
@@ -0,0 +1,177 @@
...
+            FundingKindArg::P2wpkh => CpfpFundingKind::P2wpkh,
+            FundingKindArg::P2tr => CpfpFundingKind::P2tr,
+        }
+    }
+}
+
+/// Expert-only commands that build raw transactions for you to broadcast
+/// yourself. Misuse can strand or lose funds.
+#[derive(Clone, Debug, Subcommand)]
+pub enum IKnowWhatImDoingCommand {
```
dangeross: 😅 I don't think we have to name the module/enum this. We could maybe put a **warning** prompt/confirmation when handling the command?

dangeross (commented):
Lookin good

docs/breez-sdk/snippets/react-native/unilateral_exit.ts
```diff
@@ -0,0 +1,65 @@
...
+  console.log(`Recovering ${quote.recoverableValueSat} sats for ${quote.totalFeeSat} sats in fees`)
+  console.log(`Fund a single UTXO of at least ${quote.singleUtxoFundingSat} sats`)
+  // ANCHOR_END: prepare-unilateral-exit
+  return quote
+}
+
+const exampleBuildExit = async (sdk: BreezSdk, quote: PrepareUnilateralExitResponse) => {
+  // ANCHOR: unilateral-exit
+  const secretKeyBytes = Buffer.from('your-secret-key-hex', 'hex')
+  const signer = singleKeyCpfpSigner(secretKeyBytes.buffer)
```
danielgranhao: secretKeyBytes.buffer is the pooled 8 KB slab under Node and the standard buffer polyfill (verified by running it), so this throws 'Invalid secret key'. Can be fixed by slicing by byteOffset/byteLength instead

crates/breez-sdk/breez-itest/tests/unilateral_exit.rs
```diff
@@ -0,0 +1,2743 @@
...
+}
+
+/// The sweep opts into RBF (BIP125): every input signals it, and once broadcast
+/// with confirmed ancestors bitcoind reports the sweep as `bip125-replaceable`.
+/// So an under-fee sweep sitting unconfirmed can be replaced by a higher-fee
+/// rebuild. bitcoind rejects replacement of non-signaling txs by default, so
+/// this would fail if the sweep inputs fell back to the default (non-RBF) sequence.
+#[apply(each_backend)]
+#[test_log::test(tokio::test)]
+async fn test_sweep_is_rbf_replaceable(#[case] backend: SignerBackend) -> Result<()> {
```
danielgranhao, line 807: Is there a reason to stop at checking the replaceable flag and not ensure that we can actually replace by re-quoting with a higher rate and broadcasting again? Not a blocker

danielgranhao (commented):
Looks good

crates/breez-sdk/cli/src/command/i_know_what_im_doing.rs
JssDWt: I added it to the cli. I didn't like the idea that unilateral exit is on the first level there when you type 'help'. It's not something you should use, unless you have good reason to. So I think a subcommand is nice, just to hide it a little bit (like people should when they build apps). I can also call this subcommand 'advanced' maybe?

dangeross: I agree with the subcommand. Think advanced is fine.

I think if I saw a `i-know-what-im-doing` command, it's the first thing I'd try 🙂

roeierez (approved):
Other than the pending comments of Ross and Daniel, LGTM.

crates/breez-sdk/cli/src/command/i_know_what_im_doing.rs
dangeross: To be addressed still? or leave as is?

crates/spark-wallet/src/unilateral_exit.rs
danielgranhao: @JssDWt I think one issue remains. If the anchor outspend lookup succeeds but the child tx body lookup returns `Unavailable`, the change stays `None`. The branch then fails `branch_has_tracked_change`, so the spent-funding check is skipped, and the rebuilt next child reuses the re-supplied UTXO that the confirmed child already spent. In this case, we would end up returning `Unconfirmed` when we ideally should return `Unverified` (the node is chain-verified, so it is not in operator_confirmed and the current flagging does not catch it).

crates/breez-sdk/cli/src/command/i_know_what_im_doing.rs
JssDWt: Ah right, will fix

crates/spark-wallet/src/unilateral_exit.rs
JssDWt: 73fcd107bf1b783383faf03645e78cb997a3b6be

crates/breez-sdk/cli/src/command/i_know_what_im_doing.rs
JssDWt: @dangeross fixed now

dangeross (approved):
LGTM. 

I think it would be good to see multi-leaf/fan-out end-to-end tested also like test_full_exit_and_sweep

danielgranhao (approved):
LGTM

PR #993, opened 2026-07-14
Add shared CLI scenarios for bindings

erdemyerebasmaz:
## Context

The CLI ports compile in CI, but their behavior isn't verified. This PR adds shared end-to-end JSON scenarios that are executed by the WASM, Swift, and Kotlin ports, using the Rust CLI as the reference implementation.

## Key Changes

- Added shared JSON scenarios covering core wallet flows: info and sync, receive and parse, deposit claim, Spark payments, and LNURL-pay.
- Added end-to-end test runners for the Rust CLI and the WASM, Swift, and Kotlin ports (`make cli-itest`, `make wasm-itest`, `make swift-itest`, `make kotlin-itest`).
- Added a smoke test for the WASM npm package.
- Added tests that pin the CLI command surface shared by all ports.
- Added `--lnurl-domain` to target a local LNURL server during tests.
- Integrated the new test suites into CI and the CLI sync workflow.


crates/breez-sdk/cli/tests/harness/session.rs
```diff
@@ -0,0 +1,223 @@
...
+            }
+            stable_len = len;
+        }
+
+        let transcript = self.transcript.lock().expect("transcript lock");
+        // The chunk ends where the marker's own output begins. The marker is
+        // preceded by the CLI's error prefix on the same line, so cut at the
+        // start of that line.
+        let chunk_end = transcript[..marker_pos].rfind('\n').unwrap_or(0);
+        let chunk = transcript[self.cursor..chunk_end].to_string();
```
dangeross: nit: `chunk_end` can land before `self.cursor` when a step emits no newline between the cursor and the marker (`rfind` then matches a `\n` from an earlier step), and the slice on the next line panics. Not reachable today since commands print newline-terminated output, but the JS runner already guards it (`Math.max(chunkEnd, cursor)` in `scenario.js`), so worth keeping the two in sync:

```suggestion
        let chunk_end = transcript[..marker_pos].rfind('\n').unwrap_or(0).max(self.cursor);
        let chunk = transcript[self.cursor..chunk_end].to_string();
```

.github/workflows/ci.yaml
```diff
@@ -143,9 +143,79 @@ jobs:
...
+
+      - name: Load lnurl server image
+        if: steps.lnurl-image.outputs.cache-hit == 'true'
+        run: docker load -i /tmp/breez-lnurl-built.tar
+
+      # Shares this job's toolchain, build cache, faucet limits, and lnurl
+      # image with the breez itests. Runs first so a breez-itest failure
+      # never masks the scenario signal.
+      - name: Run CLI scenarios (rust binding tests)
+        run: make cli-itest
```
dangeross: As a step in the breez itest job, a scenario failure surfaces as a `Breez integration tests` failure, so a dev hunts for a broken itest rather than the CLI scenario (and scenario flakes read as itest flakes). The build/faucet/lnurl reuse is nice here. Worth its own job/check like the other bindings so the signal points at the right suite?

dangeross (commented):
Looking good

.github/workflows/ci.yaml
erdemyerebasmaz: Addressed in 1cbbb196b

crates/breez-sdk/cli/tests/harness/session.rs
erdemyerebasmaz: Addressed in 1cbbb196b

crates/breez-sdk/cli/tests/scenarios/README.md
```diff
@@ -0,0 +1,116 @@
...
+expected output. They are the canonical behavioral tests for the SDK's CLI
+surface, and they are **shared verbatim across languages**: each language CLI
+port gets a thin runner that executes the same files, so a behavior is
+defined once and enforced everywhere.
+
+Current runners:
+
+| Runner | Drives | Entry point | CI home |
+|---|---|---|---|
+| Rust | `crates/breez-sdk/cli` binary | `crates/breez-sdk/cli/tests/scenarios.rs` (`make cli-itest`) | step in `Breez integration tests` (shares its toolchain, faucet limits, and lnurl image) |
```
dangeross: `cli-itest` is its own job now (`Rust CLI binding tests`), so this cell is stale.

.github/workflows/ci.yaml
```diff
@@ -143,9 +143,106 @@ jobs:
...
+            && docker save breez-lnurl-built:latest -o /tmp/breez-lnurl-built.tar \
+            || echo "lnurl image was not built in this run"
+
+  cli-itest:
+    name: Rust CLI binding tests
+    runs-on: ubuntu-latest
+    env:
+      FAUCET_USERNAME: ${{ secrets.FAUCET_USERNAME }}
+      FAUCET_PASSWORD: ${{ secrets.FAUCET_PASSWORD }}
+      FAUCET_CONCURRENCY: "2"
```
dangeross: `FAUCET_CONCURRENCY` is only read by `breez-itest`, and the scenario harness has no semaphore, so it's a no-op here. Same in `wasm-itest` and `cli-ci.yml`. The real bound is `--test-threads=2` in the make target.

Worth dropping it, or do we want an actual limit now that four jobs hit the faucet at once?

crates/breez-sdk/cli/tests/harness/engine.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    let path = Path::new(env!("CARGO_MANIFEST_DIR"))
+        .join("tests/scenarios")
+        .join(format!("{name}.json"));
+    let raw = std::fs::read_to_string(&path)
+        .with_context(|| format!("failed to read {}", path.display()))?;
+    let scenario: Scenario =
+        serde_json::from_str(&raw).with_context(|| format!("invalid scenario {name}"))?;
+
+    if scenario.requires.iter().any(|r| r == "docker") && !docker_available() {
```
dangeross, line 45: Only `docker` is checked here (same in `scenarios.test.js`), so `requires: ["faucet"]` does nothing. The suite-wide `FAUCET_USERNAME` gate is what skips them. A typo like `"dokcer"` runs the scenario instead of skipping it.

`fixtures` bails on an unknown name. Should `requires` do the same?

packages/wasm/itest/smoke.test.js
```diff
@@ -0,0 +1,174 @@
...
+
+  let lnurl
+  const wallets = []
+  try {
+    if (haveDocker) {
+      lnurl = await LnurlFixture.start()
+    }
+    const alice = await connectWallet('alice')
+    const bob = await connectWallet('bob', lnurl?.httpUrl)
+    wallets.push(alice, bob)
```
dangeross: nit: if `connectWallet('bob')` throws, alice never makes it into `wallets`, so she isn't disconnected and her temp dir stays behind. Pushing each as it's created covers it:

```suggestion
    const alice = await connectWallet('alice')
    wallets.push(alice)
    const bob = await connectWallet('bob', lnurl?.httpUrl)
    wallets.push(bob)
```

crates/breez-sdk/cli/tests/scenarios/README.md
```diff
@@ -0,0 +1,116 @@
...
+When adding a language runner, prefer attaching its `make <lang>-itest` step
+to the CI job that already has that language's toolchain and the faucet
+secrets rather than adding a fresh job (the rust runner rides along with the
+breez itests for exactly that reason).
+
+## The sync contract
+
+- **Scenarios are data, never ported.** Adding or changing behavior means
+  changing the Rust CLI and the scenario in the same PR; the rust runner
+  (`make cli-itest`, a step in the `Breez integration tests` CI job) proves
```
dangeross:
```suggestion
  (`make cli-itest`, a step in the `Rust CLI binding tests` CI job) proves
```

crates/breez-sdk/cli/tests/scenarios/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+The Rust runner is generic: `SCENARIO_CLI` (a command line) and `SCENARIO_CLI_CWD`
+point it at any CLI port, so most languages need no runner code at all, just a
+make target. The JS runner exists separately because it also hosts the npm-API
+smoke suite.
+
+When adding a language runner, prefer attaching its `make <lang>-itest` step
+to the CI job that already has that language's toolchain and the faucet
+secrets rather than adding a fresh job (the rust runner rides along with the
+breez itests for exactly that reason).
```
dangeross: Maybe drop this? Clearer to have its own CI job

crates/breez-sdk/cli/tests/scenarios/README.md
erdemyerebasmaz: Fixed in 3a0abbd.

.github/workflows/ci.yaml
erdemyerebasmaz: Dropped in 3a0abbd. A per-process semaphore could not bound cross-job concurrency anyway; the per-job bound stays `--test-threads=2` (`--test-concurrency=1` for JS) and both faucet clients retry with backoff.

crates/breez-sdk/cli/tests/harness/engine.rs
erdemyerebasmaz: Done in 3a0abbd: unknown `requires` names now bail in both runners, validated before the faucet gate so a typo fails even the hermetic no-faucet run.

packages/wasm/itest/smoke.test.js
erdemyerebasmaz: Done in 3a0abbd, plus `connectWallet` now cleans up after itself on partial failure.

crates/breez-sdk/cli/tests/scenarios/README.md
erdemyerebasmaz: Applied in 3a0abbd.

crates/breez-sdk/cli/tests/scenarios/README.md
erdemyerebasmaz: Dropped in 3a0abbd.

.github/workflows/ci.yaml
```diff
@@ -143,9 +143,104 @@ jobs:
...
+        uses: ./.github/actions/setup-build
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Restore lnurl server image
+        id: lnurl-image
+        uses: actions/cache@v4
+        with:
+          path: /tmp/breez-lnurl-built.tar
+          key: lnurl-image-${{ runner.os }}-${{ hashFiles('crates/breez-sdk/lnurl/**') }}
```
dangeross: Four jobs share this key now (`breez-test`, this one, `wasm-itest`, and the kotlin/swift legs of `cli-ci.yml`). On a cold cache they each build the image in parallel and all four race to save it, so we pay for the build four times and three get a "cache already exists" warning.

Do you think its worth adding a single build job the dependent jobs `needs:`?

.github/workflows/ci.yaml
```diff
@@ -143,9 +143,104 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Restore lnurl server image
+        id: lnurl-image
+        uses: actions/cache@v4
+        with:
+          path: /tmp/breez-lnurl-built.tar
+          key: lnurl-image-${{ runner.os }}-${{ hashFiles('crates/breez-sdk/lnurl/**') }}
```
dangeross: Doesn't this hashed key only catch changes in `crates/breez-sdk/lnurl/**`? But if there are any changes to it's local dependencies (`spark`, `spark-wallet`, `platform-utils`, and `lnurl-models`) the cache won't be invalidated.

erdemyerebasmaz: Fixed in 12b2f51.

.github/workflows/ci.yaml
erdemyerebasmaz: I don't think so: the parallel builds only cost free public-repo minutes, and a `needs:` job couldn't reach the kotlin leg inside the reusable workflow anyway.

dangeross (approved):
LGTM

PR #995, opened 2026-07-15
LNURL: Require a default API key on mainnet

dangeross:
Makes `--default-api-key` (`BREEZ_LNURL_DEFAULT_API_KEY`) mandatory on mainnet: the server now fails to start without it. This ensures a self-hosted LNURL server attributes all of its lightning-address receives to its own partner, rather than leaving any unattributed.

- On mainnet a missing or blank key is a startup error.
- README updated (new Partner Attribution section, flags table, example commands) and unit tests added for the key resolution.


roeierez (approved):

JssDWt (approved):
@roeierez Has to be listed as a breaking change in the release notes.

PR #996, opened 2026-07-15
Fix doubled balance from leaf counted in two maps

danielgranhao:
## What

`InMemoryTreeStore` could report double the real wallet balance for several seconds after a transfer was claimed. This collapses the store's two leaf maps into one, making the double-count structurally impossible.

## Root cause

`LeavesState` kept unreserved leaves in two separate maps, `leaves` and `missing_operators_leaves`, and `Leaves::balance()` summed both. Nothing enforced that the maps were disjoint. `add_leaves` inserted into `leaves` without removing the id from `missing_operators_leaves`, so a leaf that a refresh had flagged as missing-from-an-operator, and that an in-flight claim then re-added, was counted once as available and once as missing.

This is exactly what a deposit claim triggers: a refresh flags the just-claimed leaves as missing (the coordinator reports them, a lagging operator has not indexed them yet), then the claim adds the same leaves. The wrong value is sticky because `missing_operators_leaves` is only rebuilt by `set_leaves`, which returns early during the auto-optimization swap that starts right after the claim.

Only the in-memory store was affected. The Postgres and MySQL stores keep one row per leaf with an `is_missing_from_operators` flag that `add_leaves` upserts back to `FALSE`, so a single row can never be double-counted.

## Fix

Collapse `leaves` and `missing_operators_leaves` into a single id-keyed map, with a `missing_from_operators` flag on each entry, mirroring the SQL stores' model. One entry per id makes double-counting unrepresentable. Missing-from-operators leaves still count towards the balance but are excluded from the selectable pool, so they cannot back a payment or swap that would fail at FROST signing.

## Impact beyond CI

Any integrator polling `getInfo` shortly after a claim could observe double their real balance until the next refresh. This is the top offender behind the `Breez integration tests` job flakiness (`test_08_lnurl_send_all_with_fee_overpayment` and friends), where the doubled balance makes the test compute an impossible send amount and fail with `InsufficientFunds`.

## Observed in CI

The doubled-balance signature (`2 x (deposit - 99)`) appeared in these `Breez integration tests` runs:

- [run 29330756375](https://github.com/breez/spark-sdk/actions/runs/29330756375) (PR #991): `test_08` reported `99802` for a `50000` sat deposit.
- [run 29326871323](https://github.com/breez/spark-sdk/actions/runs/29326871323) (PR #992): same signature on `test_08` (`50000` -> `99802`) and `test_07` (`10000` -> `19802`). This branch never contained #991, confirming the bug predates it.

## Tests

Two new cases in the shared tree-store suite, run against every backend (in-memory, Postgres, MySQL):
- `test_add_leaves_clears_missing_from_operators`: a leaf set as missing-from-operators then re-added must not change the balance.
- `test_missing_from_operators_leaves_are_not_selectable`: such a leaf counts towards balance but cannot be reserved for payment.

The first fails on the pre-fix in-memory store (`available=49901 missing=49901 => balance=99802`) and passed on Postgres/MySQL, confirming the divergence this closes.

### DIFF
--- a/crates/spark/src/tree/store.rs
+++ b/crates/spark/src/tree/store.rs
@@ -67,7 +67,6 @@ struct ReservationEntry {
 #[derive(Default)]
 struct LeavesState {
     /// Every unreserved leaf, whether or not the operators all reported it.
-    /// Keyed by leaf id, so a leaf can never be counted twice.
     leaves: HashMap<TreeNodeId, StoredLeaf>,
     leaves_reservations: HashMap<LeavesReservationId, ReservationEntry>,
     /// Leaf IDs that have been finalized (spent) with their spent timestamp.
--- a/crates/spark/src/tree/store.rs
+++ b/crates/spark/src/tree/store.rs
@@ -26,6 +26,29 @@ pub const DEFAULT_RESERVATION_TIMEOUT: Duration = Duration::from_secs(60);
 struct StoredLeaf {
     node: TreeNode,
     added_at: SystemTime,
+    /// The coordinator returned this leaf but at least one operator did not.
+    /// It still counts towards the wallet balance, but cannot be selected to
+    /// back a payment or a swap until every operator reports it again.
+    /// Mirrors the `is_missing_from_operators` column of the SQL stores.
+    missing_from_operators: bool,
+}
+
+impl StoredLeaf {
+    fn available(node: &TreeNode, added_at: SystemTime) -> Self {
+        StoredLeaf {
+            node: node.clone(),
+            added_at,
+            missing_from_operators: false,
+        }
+    }
+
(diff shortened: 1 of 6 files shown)


crates/spark/src/tree/store.rs
```diff
@@ -43,8 +66,9 @@ struct ReservationEntry {
 
 #[derive(Default)]
 struct LeavesState {
+    /// Every unreserved leaf, whether or not the operators all reported it.
+    /// Keyed by leaf id, so a leaf can never be counted twice.
```
dangeross: nit: Its commenting on the fix, I think the first line is enough

dangeross (approved):
LGTM. Just I tiny nit

roeierez (approved):

PR #998, opened 2026-07-15
fix(deposit): retry static deposit refund on MISSING_ENTITY

danielgranhao:
## Problem

Refunding a static deposit right after it becomes visible can fail with
`NotFound` / `MISSING_ENTITY`. The coordinator already reveals the UTXO — it even
returns a claim quote — but an operator's chain watcher has not indexed the
transaction yet. Nothing on the refund path retried this transient window, so
`refund_deposit` surfaced a hard error to the caller.

This is the same class of "operators haven't converged yet" race the codebase
already handles for `LEAF_UNAVAILABLE` after a claim; the refund path just never
got the equivalent treatment.

It also surfaces as CI flakiness: `static_deposit_refund` waits for
`UnclaimedDeposits` and calls `refund_deposit` ~700ms later, landing squarely in
the indexing window
([failing run 29253662471](https://github.com/breez/spark-sdk/actions/runs/29253662471)):

```
test static_deposit_refund::case_1_seed ... FAILED
Error: ... status: NotFound, message: "failed to create utxo swap: ... utxo not found ..."
       ErrorInfo reason = MISSING_ENTITY
```

## Fix

- Add an `is_missing_entity_error` predicate alongside the existing
  `is_leaf_unavailable_error`, reading the structured gRPC `ErrorInfo` reason
  (not a message string).
- Back off and retry the refund on `MISSING_ENTITY` in
  `SparkWallet::refund_static_deposit`, using the same budget as the SDK's
  claim-transfer lookup (3 attempts, 500ms base, exponential ≈ 1.5s total). Every
  other error returns immediately.

### Why it's safe
- The retry wraps the whole service call, so each attempt re-signs with a **fresh
  nonce** (the failing operator call sits *after* nonce generation).
- `MISSING_ENTITY` means the swap was never created server-side, so retrying
  cannot double-spend.

### Scope
- Kept **out** of `is_backoff_retryable_error` so the leaf-selection retry paths
  are unaffected (a unit test guards this).
- No itest change needed — the SDK now absorbs the window, so `static_deposit_refund`
  passes and integrators calling `refund_deposit` in that window are fixed too.

### DIFF
--- a/crates/spark-wallet/src/wallet.rs
+++ b/crates/spark-wallet/src/wallet.rs
@@ -135,6 +135,22 @@ fn is_leaf_unavailable_error(error: &OperatorRpcError) -> bool {
     operator_error_reason(error).as_deref() == Some(LEAF_UNAVAILABLE_REASON)
 }
 
+/// `ErrorInfo` reason the operators set when the on-chain UTXO backing a static
+/// deposit refund is not indexed yet: their chain watcher lags the coordinator,
+/// which already reveals the UTXO. Transient, clears once the operators catch up.
+const MISSING_ENTITY_REASON: &str = "MISSING_ENTITY";
+
+/// Checks if the operators report the refund UTXO is not indexed yet. Surfaces
+/// right after a deposit becomes visible, so the caller backs off and retries.
+fn is_missing_entity_error(error: &OperatorRpcError) -> bool {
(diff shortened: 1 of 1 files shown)


dangeross (approved):
LGTM

roeierez (approved):

PR #1000, opened 2026-07-16
lnurl: harden public request handling and config

danielgranhao:
Bundle of low-risk LNURL server hardening:

- Enforce the advertised min/max sendable bounds on the invoice callback (also rejects amount 0), which were previously not checked.
- Guard the empty-domain-map fail-open so it never allows all hosts on mainnet (adds State.is_mainnet).
- Harden CRL fetching: add a request timeout, check the HTTP status before parsing the body, and trim/skip empty entries.
- Hard-fail on an invalid or wrong-length ssp_auth_seed instead of silently generating a random identity (unset still generates one).
- Fix config precedence to the documented CLI > ENV > TOML: only flags actually passed on the command line override env and the config file.
- Document that --scheme only sets the scheme of generated URLs; the server binds plain HTTP and expects TLS termination at a proxy.
- Fix README: 4,000,000,000 msat is 4,000,000 sats (was 40,000).


crates/breez-sdk/lnurl/README.md
```diff
@@ -97,7 +100,8 @@ address = "0.0.0.0:8080"
 auto_migrate = true
 log_level = "info"
 network = "mainnet"
-scheme = "https"
+scheme = "https"                     # Scheme for generated URLs only; the server
+                                    # binds plain HTTP, terminate TLS at a proxy
```
dangeross, line 104: nit: The indent is slightly off here

crates/breez-sdk/lnurl/src/routes.rs
```diff
@@ -1489,9 +1509,13 @@ async fn sanitize_domain<DB>(
...
 ) -> Result<String, (StatusCode, Json<Value>)> {
     let domain = domain.trim().to_lowercase();
-    // If domains list is empty allow all domains (for testing)
     let domains = state.domains.read().await;
-    if domains.is_empty() || domains.contains_key(&domain) {
+    if domains.contains_key(&domain) {
+        return Ok(domain);
+    }
+    // An empty allow-list falls open to any host, for local/test setups only.
+    // Never on mainnet: there it must be an explicit deny.
```
dangeross, line 1517: I wonder if this also needs a more explicit statement in the README?

dangeross (commented):
Looks good

crates/breez-sdk/lnurl/src/routes.rs
danielgranhao: Done

dangeross (approved):
LGTM

roeierez (approved):

PR #1002, opened 2026-07-16
lnurl: enqueue zap receipt on every invoice-paid, not just the first

danielgranhao:
When an invoice got paid, we stored the preimage and queued its Nostr zap receipt together. If the queueing step failed after the preimage was saved, a retry saw the preimage already there, skipped everything, and never re-queued the receipt, so it was lost for good.

This fixes it by always queueing the zap receipt when we process a paid invoice, not just the first time. Re-queuing is safe: duplicates are ignored, and the publisher skips any receipt that was already sent, so nothing gets published twice.


JssDWt (approved):
LGTM

dangeross (approved):
LGTM. I'm thinking the `/invoice-paid` / `/invoices-paid` endpoints could potentially be removed too

PR #1004, opened 2026-07-17
lnurl: harden the unregister endpoint

danielgranhao:
Harden the LNURL unregister endpoint:
- Only delete if the unregister message specifies the current username
- Add prefix to unregister message to avoid replay from a captured register for same username (old format without prefix still accepted to allow migration)


dangeross (approved):
LGTM

roeierez (approved):
LGTM

PR #1007, opened 2026-07-20
Cross-chain: Defer boltz client build until first use

### LINKED ISSUE #1005: lazy-initialize Boltz cross-chain provider
Boltz is initialized synchronously during `connect()`, even though the provider is only needed for cross-chain features. This adds a noticeable startup latency.

#### Proposed fix

Register the provider during `connect()`, but defer (or background) its network initialization until the first cross-chain operation. APIs that require the provider can await readiness when first used.

This removes non-essential work from the `connect()` path, reducing startup latency while preserving existing behavior.

dangeross:
This PR:
- Defers initialization of the boltz client out of the `connect()` call
- Monitors the pending boltz conversions and warms the client when needed
- Lazily initializes the boltz client on first use

This change also opens up including the cross chain feature by default in client-mode because provider configuration is now lightweight.

Fixes #1005

### DIFF
--- a/crates/breez-sdk/core/src/cross_chain/boltz.rs
+++ b/crates/breez-sdk/core/src/cross_chain/boltz.rs
@@ -28,13 +28,12 @@ use tracing::{debug, error, info, warn};
 use super::{
     CrossChainFeeMode, CrossChainPrepared, CrossChainProvider, CrossChainProviderContext,
     CrossChainRouteFilter, CrossChainRoutePair, CrossChainService, SourceAsset,
-    derive_btc_leg_transfer_id,
+    boltz_storage_adapter::PROVIDER_TAG_BOLTZ, derive_btc_leg_transfer_id,
 };
 use crate::{
     ConversionInfo, ConversionStatus, CrossChainAddressDetails, Network, PaymentMetadata,
     PaymentStatus, Storage,
     error::SdkError,
-    persist::{ConversionFilter, StorageListPaymentsRequest, StoragePaymentDetailsFilter},
     sdk::LightningSender,
     utils::{
         payments::resolve_and_insert_payment_metadata,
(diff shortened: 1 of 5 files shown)


roeierez (approved):
LGTM

crates/breez-sdk/core/src/cross_chain/boltz.rs
```diff
@@ -110,48 +98,111 @@ impl BoltzService {
...
+                    _ = shutdown_receiver.changed() => return,
+                    () = sleep(super::MONITOR_INTERVAL) => {}
+                }
+            }
+        });
+    }
+
+    async fn has_pending_conversions(storage: &Arc<dyn Storage>) -> Result<bool, SdkError> {
+        let payments = storage
+            .list_payments(StorageListPaymentsRequest {
```
danielgranhao: Did you consider using `storage.list_active_cross_chain_swaps("boltz".to_string())` instead of `list_payments`?

Besides being cheaper than listing payments, it would also catch cases where we started a payment but failed to write the payment metadata. If this happens, we wouldn't detect there's a pending conversion as it stands.

crates/breez-sdk/core/src/cross_chain/boltz.rs
```diff
@@ -110,48 +98,111 @@ impl BoltzService {
...
+                if service.client.initialized() {
+                    return;
+                }
+                match Self::has_pending_conversions(&service.storage).await {
+                    Ok(true) => match service.client().await {
+                        Ok(_) => return,
+                        Err(e) => warn!("Boltz resume monitor: warm failed, will retry: {e:?}"),
+                    },
+                    Ok(false) => {}
+                    Err(e) => debug!("Boltz resume monitor: pending check failed: {e:?}"),
```
danielgranhao: Maybe warn level instead?
```suggestion
                    Err(e) => warn!("Boltz resume monitor: pending check failed: {e:?}"),
```

erdemyerebasmaz (approved):

crates/breez-sdk/core/src/cross_chain/boltz.rs
```diff
@@ -1086,6 +1138,109 @@ mod tests {
...
+        fn create_temp_dir(name: &str) -> std::path::PathBuf {
+            let mut path = std::env::temp_dir();
+            path.push(format!(
+                "breez-test-boltz-{}-{}",
+                name,
+                uuid::Uuid::new_v4()
+            ));
+            std::fs::create_dir_all(&path).unwrap();
+            path
+        }
```
danielgranhao: nit: has a copy on line 1250. Could be moved to the parent mod tests

danielgranhao (commented):
Looks good - just 1 concern about how pending conversions are detected

crates/breez-sdk/core/src/cross_chain/boltz.rs
dangeross: Make sense, also when looking ahead to receive swaps

danielgranhao (approved):
LGTM

Issue #374, opened 2025-11-04
Add unilateral exit API


roeierez:
Currently it is only expose in the CLI.

kingonly:
https://github.com/breez/spark-sdk/pull/795

JssDWt:
#992 added the unilateral exit interface to the sdk

Next steps are:
- [x] Persisting ancestors and using persistent leaf storage by default, so we can unilaterally exit any time from local state when operators are offline/gone. #1010 
- [x] Export leaf state (for remote backup) #1029 
- [ ] Ensuring local data availability for _all_ leaf states, also intermediate states + available_missing_from_operators leaves #1028 (nice to have)

kingonly:
Closing this because the building blocks are complete. 
Opening #1068 for completeness.

PR #1010, opened 2026-07-21
Persistent ancestor store

JssDWt:
Store leaves and their ancestors locally, to ensure we can always exit from local state in case operators go down. 

Part of #374, successor of #992.

- Use sqlite leaf store instead of in-memory by default.
- Always download ancestors
- I tried to minimize the amount of times ancestors have to be read from storage or fetched from the operators.

Considerations:

- Ancestors are downloaded from a single operator 
- Failing to fetch ancestors is not intrusive to the point where we'd error
- Downloading the ancestors happens on refresh. That's the self-repair spot for any missing ancestors too.

Notes:

We'll note that with the 1/N trust model, you'd expect at least one operator to give you the ancestors. You can also trust that there was a valid ancestry chain at the moment of receiving with the same model. 

Then some philosiphical notes. There is not necessarily a good point at which to enforce having the ancestors. When you already own a leaf, you'd best spend it if it wouldn't have ancestors. If you have a leaf ready for you that you just have to claim, there is no reason not to claim it if it wouldn't have ancestors. It's better to claim and spend it than do nothing. Combined with the trust model notes the best-effort download seems like the best choice.


crates/spark-wallet/src/wallet.rs
```diff
@@ -1373,50 +1379,35 @@ impl SparkWallet {
...
+        }
         for leaf_id in leaf_ids {
-            let Some(leaf) = tree_nodes.get(leaf_id).cloned() else {
-                // Offline sourcing from local storage is deferred.
-                warn!("Requested leaf {leaf_id} not returned by the operators; cannot exit it");
-                continue;
-            };
-            let tree_service = self.tree_service.clone();
-            if let Err(e) =
-                spark::services::build_unilateral_exit_chain(leaf, &mut tree_nodes, move |ids| {
```
dangeross, line 1415: Possibly `build_unilateral_exit_chain` can removed from `spark`? I think it's only used for tests now

crates/spark/src/tree/service.rs
```diff
@@ -351,29 +402,29 @@ impl TreeService for SynchronousTreeService {
...
+            node_map.insert(leaf.id.clone(), leaf.clone());
+        }
+        let missing_operator_ids: Vec<TreeNodeId> = missing_operator_leaves
+            .iter()
+            .map(|l| l.id.clone())
+            .collect();
+        let new_leaf_ids: Vec<TreeNodeId> = new_leaves.iter().map(|l| l.id.clone()).collect();
+        let missing_operator_pedigrees = assemble_exit_chains(&node_map, &missing_operator_ids);
+        let refreshed_pedigrees = assemble_exit_chains(&node_map, &new_leaf_ids);
+        let pedigrees = self.check_renew_nodes(refreshed_pedigrees).await?;
```
dangeross: I'm concerned we're solely trusting the coordinator's pedigree set to update the ancestor store. We run `refresh_leaves` on every sync and now in `refresh_before_exit`, so a rogue coordinator could effectively wipe the leaf/ancestor store during normal operation, well before any exit. Our check with other operators is currently one-way (operator missing coordinator leaf), perhaps we should also be flagging leaves that are missing from the coordinator but not the operators and also include them for exit (I realize there is a potential lag effect here, but maybe more correct)?

crates/spark/src/tree/mod.rs
```diff
@@ -494,40 +523,45 @@ impl LeafLike for TreeNode {
...
+        Ok(nodes)
+    }
+
+    /// Reconstructs a leaf's exit chain, ordered root to leaf, by walking
+    /// `parent_node_id` up from `leaf_id`. A gap or cycle stops the walk and the
+    /// partial chain is returned rather than erroring: it may still be enough to
+    /// exit, and a caller checks completeness (the root has no parent) itself. A
+    /// leaf absent from the store returns an empty chain.
+    async fn get_exit_chain(&self, leaf_id: &TreeNodeId)
+    -> Result<Vec<TreeNode>, TreeServiceError>;
```
dangeross: Seems these three trait fns (`get_node`, `get_nodes`, `get_exit_chain`) have no caller outside tests.

crates/spark/src/tree/store.rs
dangeross: Should this also be put in the not_available bucket now (looking at the other store implementations)?

dangeross (commented):
Lookin good

crates/spark/src/tree/service.rs
JssDWt: I'm going to make a retained mode in the database. There is actually a failing integration test that fails on the same thing. Basically to add a leaf a single operator is enough. To remove a leaf all operators have to include proof.

crates/spark-wallet/src/wallet.rs
JssDWt: Yes, done

crates/spark/src/tree/mod.rs
JssDWt: Yes, removed, they came from the work in progress.

crates/spark/src/tree/store.rs
JssDWt: Yes, done.

crates/breez-sdk/core/src/persist/sqlite.rs
```diff
@@ -53,15 +54,25 @@ impl SqliteStorage {
...
         storage.migrate()?;
         Ok(storage)
     }
 
     pub(crate) fn get_connection(&self) -> Result<Connection, StorageError> {
-        Ok(Connection::open(self.get_db_path())?)
+        let conn = Connection::open(self.get_db_path())?;
+        // Wait for the lock instead of failing with SQLITE_BUSY when the shared-file
+        // tree store is mid-write.
+        conn.busy_timeout(Duration::from_secs(5))?;
```
roeierez: I think the default value is already 5 seconds.

crates/spark/src/tree/service.rs
JssDWt: Deferring this to the followup that addresses being exitable under all circumstances.

dangeross (approved):
LGTM

JssDWt:
- The ancestor store is now per-leaf. It is slightly less efficient space-wise, because shared ancestors are duplicated, but it saves in simplicity: no database recursion and no garbage collection.
- The ancestors are now fetched only after full operations, in the background. It is no longer an intrusive part of the operations.
- There is a config flag that determines whether to do the ancestor fetching.
- There is another flag that determines whether to fetch ancestors on sync_wallet.

crates/spark/src/tree/mod.rs
```diff
@@ -494,40 +525,38 @@ impl LeafLike for TreeNode {
...
+    /// existing id refreshes its mutable fields (status, transactions, parent);
+    /// its `value` and verifying key are fixed, and a change to either errors.
+    async fn add_leaves(&self, leaves: &[LeafPedigree]) -> Result<(), TreeServiceError>;
+
+    /// Stores the ancestors of `pedigrees`, leaving the leaf pool untouched.
+    ///
+    /// Unlike [`Self::add_leaves`], this neither inserts a leaf nor clears its spent
+    /// marker, so a leaf spent since its pedigree was resolved stays spent. Ancestors
+    /// of an id the pool no longer holds are written but unreachable, and the next
+    /// collection removes them.
```
dangeross: Seems ancestor rows only get cleaned up when their leaf is deleted, not a general sweep, so ancestors written after the leaf is already gone (a fresh leaf spent while its ancestors are still being fetched) could remain permanently. The transaction between selecting leaves and writing ancestors can't be atomic, so should the `store_ancestors` implementations check the leaf still exists as it writes?

crates/spark/src/services/timelock_manager.rs
```diff
@@ -75,116 +72,134 @@ impl TimelockManager {
...
+        &self,
+        leaf: &TreeNode,
+        nodes: &mut HashMap<TreeNodeId, TreeNode>,
+    ) -> Result<TreeNode, ServiceError> {
+        let parent_id = leaf
+            .parent_node_id
+            .clone()
+            .ok_or_else(|| ServiceError::Generic(format!("Node {} has no parent node", leaf.id)))?;
+        if !nodes.contains_key(&parent_id) {
+            for ancestor in self.fetch_leaf_ancestors(leaf).await? {
```
roeierez: Shouldn't we leave the fetch ancestor to the high level operation? The timelock manager is a part of such operation and here we are probably paying performance penalty on every renew?

crates/spark-wallet/src/wallet.rs
```diff
@@ -589,7 +613,7 @@ impl SparkWallet {
             }
         };
 
-        self.maybe_start_optimization().await;
+        self.on_leaves_changed().await;
```
roeierez, line 620: I like the fact the spark wallet layer exposes all the means to fetch/download/persist exit chain.
Intuitively I think the automatic download handling trigger should live at the breez sdk level per payment/high level operation rather than in the low level spark wallet.

roeierez (changes requested):
Looking really good and easy to follow! Had a quick pass and now going a bit deeper.

dangeross (commented):
Just this one question, otherwise looks good

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -1252,7 +1263,17 @@ pub struct TokenMetadata {
...
+pub struct SyncWalletRequest {
+    /// Whether this sync also resolves the exit chains of leaves still missing one,
+    /// so they can be spent unilaterally without the Spark operators.
+    ///
+    /// Costs one extra round trip per sync that finds missing chains. Leave unset to
+    /// let the background resolver handle it, which is the default and is what most
+    /// apps want. Set it when background services are disabled, since nothing else
+    /// resolves them there.
+    #[cfg_attr(feature = "uniffi", uniffi(default = None))]
+    pub include_exit_chains: Option<bool>,
```
roeierez: I am tending to decouple the unilateral exit flow from the regular sync. Price would be adding a dedicated endpoint but we will need it anyway for users to fetch the exit chain and backup.
Assuming we have a method let's say "fetch_exit_chain" then we can add there an optional flag on whether to also "sync" or only fetch locally.

crates/spark/src/tree/store.rs
```diff
@@ -523,11 +661,21 @@ impl InMemoryTreeStore {
...
+                    ensure_node_compatible(&stored.node, &fresh.node)?;
                     *stored = fresh;
                 }
             }
         }
 
+        // A pre-refresh leaf that reappeared (fresh, missing-operators, or
+        // preserved) keeps its ancestor chain, already upserted above; only an
+        // id still absent from the pool afterward is truly gone.
+        for id in &old_leaf_ids {
```
roeierez, line 642: I hope I read it right. On refresh we delete leaves that were not returned. Seems like if we are in a middle of an exit where some of the branches were confirmed on chain the associated leaves will not be returned by the operator (they are not available anymore?) which will cause us delete it from the store. Perhaps we are loosing access to the refund tx in that case?

crates/breez-sdk/breez-itest/tests/unilateral_exit.rs
```diff
@@ -643,6 +643,7 @@ async fn test_full_exit_and_sweep(#[case] backend: SignerBackend) -> Result<()>
 /// scan would re-drive the refund (with a fresh CPFP child) and rebuild the sweep.
 #[apply(each_backend)]
 #[test_log::test(tokio::test)]
+#[ignore = "resume needs the leaf in local storage, but a refresh deletes it once no operator reports it Available; fix by querying the operators for more statuses, or by not deleting leaves absent from a refresh"]
```
roeierez, line 646: I see now this comment that matches my comment on the refresh. Probably we should not delete leaves that were part of the exit process.

roeierez (changes requested):

crates/spark-postgres/src/tree_store.rs
```diff
@@ -1227,6 +1464,222 @@ impl PostgresTreeStore {
         Ok(())
     }
 
+    /// Errors if any incoming node conflicts with a stored node of the same id
+    /// on a field that must not change (`value`, verifying key). One query loads
+    /// the existing rows (leaf or ancestor) for the batch.
+    async fn check_nodes_compatible(
```
roeierez: This one is very expensive. It is invoked as part of set_leaves (refresh) and seems to query all leaves and ancestor from the db into memory. 
Same problem for node js store (I didn't go over the others).
What is the purpose of this function? Why do we need to check the tree node content vs the stored one?

roeierez (changes requested):

crates/spark/src/services/timelock_manager.rs
JssDWt: This was already there. You need the parents for renewal to ensure you're creating valid transactions. It was actually batching before though, I will put that back.

crates/breez-sdk/wasm/js/postgres-tree-store/index.cjs
```diff
@@ -212,6 +239,115 @@ class PostgresTreeStore {
...
+      // walked one chain at a time so a node id stored under another leaf can
+      // never cross-contaminate this one.
+      const ancestorsResult = await this.pool.query(
+        "SELECT leaf_id, data FROM brz_tree_ancestors WHERE user_id = $1 AND leaf_id = ANY($2)",
+        [this.identity, leafIds]
+      );
+      const leavesResult = await this.pool.query(
+        "SELECT data FROM brz_tree_leaves WHERE user_id = $1 AND id = ANY($2)",
+        [this.identity, leafIds]
+      );
```
danielgranhao: Leaves and their ancestors are being queried separately, so I think we could end up assembling a chain from 2 different snapshots. The doc above states "in one query" so probably unintentional? Same issue on mysql implementation

CLAUDE.md
```diff
@@ -113,7 +113,9 @@ JS implementations also have migration files (`migrations.cjs`) alongside their
...
+All SDK-owned Postgres and MySQL schema identifiers — tables, indexes, and named constraints (PKs, FKs) — are prefixed with `brz_` (e.g. `brz_payments`, `brz_idx_payments_user_timestamp`, `brz_fk_token_outputs_metadata_user`). This isolates the SDK's schema from customer-managed tables sharing the same database.
+
+**The SQLite tree store also uses the `brz_` prefix** (Rust `spark-sqlite` and the Node `better-sqlite3` store): it shares the wallet's main SQLite database file, so its tables must not collide with the main storage's, and it tracks its schema in a `brz_tree_schema_migrations` table rather than the file's `PRAGMA user_version`, which the main storage owns. The main SQLite storage's own tables and the browser **IndexedDB** stores are NOT prefixed — the main storage owns its file, and each IndexedDB database is per-origin isolated (the tree store there uses a separate `-tree` database).
```
danielgranhao: Nit: the `brz` prefix IMO isn't necessary for avoiding collision with the main sqlite storage. If we wanted a prefix for that purpose, something more specific like `tree_store` would be preferable, which we already have in a way (all tables are currently prefixed by `tree`). That said, I think it's fine to keep the `brz` prefix, but mainly for consistency with the non-sqlite implementations. What I would suggest (not blocking) is changing the explanation here and in in-code docs elsewhere

crates/spark-sqlite/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+workspace = true
+
+[dependencies]
+async-trait.workspace = true
+macros.workspace = true
+platform-utils.workspace = true
+rusqlite.workspace = true
+serde_json.workspace = true
+spark-wallet.workspace = true
+thiserror.workspace = true
```
danielgranhao: This can be removed

crates/spark-sqlite/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "spark-sqlite"
+version.workspace = true
+edition = "2024"
+
+[features]
+test-utils = ["spark-wallet/test-utils"]
```
danielgranhao: Also can be removed?

crates/spark-sqlite/src/lib.rs
```diff
@@ -0,0 +1,1516 @@
...
+        // after-refresh leaves are immune. Leaves present in the refresh below are
+        // re-inserted, and any candidate that does not come back is dropped below.
+        tx.execute(
+            "DELETE FROM brz_tree_leaves
+             WHERE reservation_id IS NULL AND added_at < ?1",
+            params![refresh_ms],
+        )
+        .map_err(|e| generic("delete old leaves", e))?;
+
+        Self::upsert_pedigree_ancestors(&tx, leaves)?;
```
danielgranhao: The leaf upsert below skips spent ids, but the ancestor upsert here doesn't. So it seems possible that we write ancestors that are orphaned from the get-go and thus are never cleaned up. We could skip spent too (same in other implementations)

crates/breez-sdk/wasm/src/tree_store/mod.rs
```diff
@@ -534,16 +586,19 @@ type LeafSelection =
...
+    leavesMissingExitChains: () => Promise<string[]>;
     getLeaves: () => Promise<Leaves>;
+    getExitChains: (leafIds: string[]) => Promise<LeafPedigree[]>;
     getAvailableBalance: () => Promise<bigint>;
     getVerifiedLeafKeys: () => Promise<[string, string, string][]>;
-    setLeaves: (leaves: TreeNode[], missingLeaves: TreeNode[], refreshStartedAtMs: number) => Promise<void>;
-    cancelReservation: (id: string, leavesToKeep: TreeNode[]) => Promise<void>;
-    finalizeReservation: (id: string, newLeaves: TreeNode[] | null) => Promise<void>;
+    setLeaves: (leaves: LeafPedigree[], missingLeaves: LeafPedigree[], refreshStartedAtMs: number) => Promise<void>;
+    cancelReservation: (id: string, leavesToKeep: LeafPedigree[]) => Promise<void>;
```
danielgranhao: cancel reservation still uses TreeNode, not LeafPedigree


```suggestion
    cancelReservation: (id: string, leavesToKeep: TreeNode[]) => Promise<void>;
```

crates/breez-sdk/wasm/src/tree_store/mod.rs
```diff
@@ -519,6 +564,13 @@ interface LeavesReservation {
     leaves: TreeNode[];
 }
 
+/** A leaf together with its ancestor chain (root to parent), so a stored leaf
```
danielgranhao: nit: the doc could be misread to state that the chain is ordered from root to parent when it's nearest-first

crates/spark-sqlite/src/lib.rs
```diff
@@ -0,0 +1,1516 @@
...
+        "CREATE TABLE IF NOT EXISTS brz_tree_schema_migrations (version INTEGER PRIMARY KEY);",
+    )
+    .map_err(|e| generic("create tree migrations table", e))?;
+    let current: i64 = conn
+        .query_row(
+            "SELECT COALESCE(MAX(version), 0) FROM brz_tree_schema_migrations",
+            [],
+            |row| row.get(0),
+        )
+        .map_err(|e| generic("read tree schema version", e))?;
```
danielgranhao: Shouldn't this be done inside the tx that starts below?

crates/breez-sdk/wasm/js/web-tree-store/index.js
```diff
@@ -0,0 +1,1495 @@
...
+        (res) => {
+          const rootLeafIds = new Set();
+          for (const row of res.ancestors) {
+            if (row.parent_node_id == null) rootLeafIds.add(row.leaf_id);
+          }
+          return res.leaves
+            .filter((l) => l.parent_node_id != null && !rootLeafIds.has(l.id))
+            .map((l) => l.id);
+        }
+      );
```
danielgranhao, line 424: This gets all leaves and ancestors after every leaf-changing operation. On SQL implementations we have an indexed NOT EXISTS. I realize a web backend is unlikely to get a large number of leaves, but one option to optimize this would be to add a e.g. `chain_complete` flag to the leaf row when writting ancestors.

crates/spark/src/tree/store.rs
```diff
@@ -1069,6 +1273,87 @@ mod tests {
...
+    #[async_test_all]
+    async fn test_get_exit_chains() {
+        shared_tests::test_get_exit_chains(&InMemoryTreeStore::new()).await;
+    }
+
+    #[async_test_all]
+    async fn test_get_exit_chain_missing_ancestor() {
+        shared_tests::test_get_exit_chain_missing_ancestor(&InMemoryTreeStore::new()).await;
+    }
+
```
danielgranhao, line 1257: I think we are missing `test_leaves_missing_exit_chains` here

crates/breez-sdk/breez-itest/tests/unilateral_exit.rs
JssDWt: Yes, but that has side-effects, that I will defer to the last followup PR on unilateral exit, where we no longer hard delete on absence.

crates/spark-postgres/src/tree_store.rs
JssDWt: I don't think we need it. Removed.

crates/breez-sdk/breez-itest/tests/unilateral_exit.rs
JssDWt: Addressed in #1028

crates/spark/src/tree/store.rs
JssDWt: Addressed in #1028

crates/spark-sqlite/src/lib.rs
```diff
@@ -0,0 +1,1496 @@
...
+
+    async fn leaves_missing_exit_chains(&self) -> Result<Vec<TreeNodeId>, TreeServiceError> {
+        let conn = self.get_connection()?;
+        let mut stmt = conn
+            .prepare(
+                "SELECT l.id FROM brz_tree_leaves l
+                 WHERE l.parent_node_id IS NOT NULL
+                   AND NOT EXISTS (
+                     SELECT 1 FROM brz_tree_ancestors a
+                     WHERE a.leaf_id = l.id AND a.parent_node_id IS NULL
```
dangeross: Does this handle a leaf whose parent changed since its chain was stored? It checks for any rooted ancestor row, not whether the chain actually connects from the leaf's current `parent_node_id`.

I'm thinking about the multi-device case, where a timelock renewal on device A re-parents a leaf onto a new split node (same id). Device B does a leaves-only refresh and this check still finds the old root and reports complete, meaning device B is missing the split node chain link between the current leaf and the previous parent leaving it unexitable (on device B).

crates/spark-sqlite/src/lib.rs
```diff
@@ -0,0 +1,1496 @@
...
+            for row in rows {
+                let (leaf_id, data) =
+                    row.map_err(|e| generic("read get_exit_chains ancestor row", e))?;
+                let leaf_id = TreeNodeId::from_str(&leaf_id).map_err(TreeServiceError::Generic)?;
+                ancestors_by_leaf
+                    .entry(leaf_id)
+                    .or_default()
+                    .push(Self::row_to_node(&data)?);
+            }
+        }
```
dangeross: It looks like the other implementations use one `UNION ALL` query making it atomic, should we do the same here?

crates/spark-sqlite/src/lib.rs
```diff
@@ -0,0 +1,1496 @@
...
+    // ---- row (de)serialization ----
+
+    fn row_to_node(data: &str) -> Result<TreeNode, TreeServiceError> {
+        serde_json::from_str(data).map_err(|e| generic("deserialize node", e))
+    }
+
+    // ---- ancestor persistence ----
+
+    /// Replaces the ancestor rows owned by each pedigree's leaf (see
+    /// `upsert_ancestors`).
```
danielgranhao:
```suggestion
```

crates/breez-sdk/wasm/js/node-tree-store/migrations.cjs
```diff
@@ -0,0 +1,175 @@
...
+                        value                     INTEGER NOT NULL DEFAULT 0,
+                        verifying_public_key      TEXT NOT NULL DEFAULT '',
+                        signing_public_key        TEXT NOT NULL DEFAULT '',
+                        data                      TEXT NOT NULL,
+                        is_missing_from_operators INTEGER NOT NULL DEFAULT 0,
+                        reservation_id            TEXT,
+                        added_at                  INTEGER
+                    )`,
+          `CREATE INDEX IF NOT EXISTS brz_idx_tree_leaves_parent ON brz_tree_leaves (parent_node_id)`,
+          `CREATE INDEX IF NOT EXISTS brz_idx_tree_leaves_reservation ON brz_tree_leaves (reservation_id)`,
```
dangeross, line 149: `spark-sqlite` backs the same slim-candidate selection query with this index, should we add it here too?
```
    CREATE INDEX brz_idx_tree_leaves_slim
        ON brz_tree_leaves (status, is_missing_from_operators, value)
        WHERE reservation_id IS NULL;
```

crates/breez-sdk/wasm/js/node-tree-store/index.cjs
```diff
@@ -0,0 +1,1187 @@
...
+      // another leaf can never cross-contaminate this one.
+      const placeholders = leafIds.map(() => "?").join(",");
+      const leafRows = this.db
+        .prepare(`SELECT data FROM brz_tree_leaves WHERE id IN (${placeholders})`)
+        .all(...leafIds);
+      const ancestorRows = this.db
+        .prepare(
+          `SELECT leaf_id, data FROM brz_tree_ancestors WHERE leaf_id IN (${placeholders})`
+        )
+        .all(...leafIds);
```
danielgranhao: Another instance of the 2 queries that should be 1

crates/spark-sqlite/src/lib.rs
```diff
@@ -0,0 +1,1496 @@
...
+}
+
+#[async_trait]
+impl TreeStore for SqliteTreeStore {
+    // ---- node/ancestor chain (durable, exit-critical) ----
+
+    async fn get_exit_chains(
+        &self,
+        leaf_ids: &[TreeNodeId],
+    ) -> Result<Vec<LeafPedigree>, TreeServiceError> {
```
danielgranhao, line 647: Here we also query leaves and their ancestors separately. Less important compared with pg/mysql, but as it stands each read could take their own WAL snapshot. Putting both on a tx would prevent it IIUC

crates/spark-postgres/src/tree_store.rs
```diff
@@ -1227,6 +1474,191 @@ impl PostgresTreeStore {
...
+    /// that must not change (`value`, verifying key): checked against a
+    /// stored leaf of the same id and this leaf's own previous copy, never
+    /// against another leaf's copy, since each leaf's exit chain uses only
+    /// its own rows and a table-wide check would cost every write the whole
+    /// table's size.
+    async fn check_ancestors_compatible(
+        &self,
+        tx: &tokio_postgres::Transaction<'_>,
+        rows: &[(String, &TreeNode)],
+    ) -> Result<(), TreeServiceError> {
```
danielgranhao: Shouldn't this have been removed too? Same for mysql

crates/spark-wallet/src/wallet.rs
```diff
@@ -1472,6 +1486,22 @@ impl SparkWallet {
...
+        if let Err(e) = self.tree_service.refresh_leaves().await {
+            warn!("unilateral exit: refresh failed, planning from local state: {e:?}");
+        }
+        // An exit is planned entirely from stored chains, so resolve the missing
+        // ones here rather than relying on whatever the caller has scheduled.
+        if let Err(e) = self.exit_chain_resolver.resolve_missing_chains().await {
+            warn!(
+                "unilateral exit: resolving exit chains failed, planning from local state: {e:?}"
+            );
+        }
```
danielgranhao: `resolve_missing_chains` may skip leaves due to backoff. Here, when refreshing before exit, I think it would be nice to ignore backoffs

crates/spark/src/tree/exit_chain_resolver.rs
```diff
@@ -0,0 +1,436 @@
...
+                    .iter()
+                    .filter(|pedigree| !pedigree.ancestors.is_empty())
+                    .cloned(),
+            )
+            .collect();
+        if !to_store.is_empty() {
+            self.tree_service.store_exit_chains(&to_store).await?;
+        }
+
+        self.record_outcomes(&resolved, &unresolved, now).await;
```
danielgranhao: This records outcomes for leaves returned by operators but leaves that are absent will still always result in a fetch. Maybe we can backoff `candidate` - `resolved` instead of `resolved`? Or, if the current behavior is acceptable, reflect it in the rustdocs above

danielgranhao (commented):
Final comments :)

danielgranhao (approved):
LGTM

dangeross (approved):
LGTM

crates/spark-postgres/src/tree_store.rs
```diff
@@ -262,17 +264,24 @@ impl TreeStore for PostgresTreeStore {
 
         let mut client = self.pool.get().await.map_err(map_err)?;
         let tx = client.transaction().await.map_err(map_err)?;
+        // Serialize with set_leaves/finalize_reservation: both mutate this
+        // tenant's leaf and ancestor rows, and running concurrently could
+        // interleave one transaction's delete with the other's insert for the
+        // same leaf.
+        self.acquire_write_lock(&tx).await?;
```
roeierez: This lock is a new addition here.
A claim writes no ancestor rows (`insert_leaves` passes ancestors: Vec::new()), so effectively adds a lock for nothing. Can we take it only when the batch actually carries ancestors: `if leaves.iter().any(|p| !p.ancestors.is_empty())?`

crates/spark-postgres/src/tree_store.rs
```diff
@@ -485,6 +606,29 @@ impl TreeStore for PostgresTreeStore {
...
+            .chain(missing_operators_leaves.iter())
+            .map(|p| p.leaf.id.to_string())
+            .collect();
+        tx.execute(
+            "DELETE FROM brz_tree_ancestors WHERE user_id = $1 AND leaf_id IN (
+                SELECT id FROM brz_tree_leaves
+                WHERE user_id = $1 AND reservation_id IS NULL AND added_at < $2
+                  AND id <> ALL($3)
+             )",
+            &[&self.identity, &refresh_timestamp, &fresh_ids],
```
roeierez: `id <> ALL($3)` on a bind parameter seems like a full scan of user leaves and since the ids passed are all leaves it performs probably at O(N^2). 
In steady state it deletes nothing. Perhaps instead add RETURNING id to the `brz_tree_leaves` delete right below and delete ancestors by that which is on most times empty?

crates/spark/src/tree/service.rs
```diff
@@ -351,29 +398,23 @@ impl TreeService for SynchronousTreeService {
...
-                {
-                    error!("Failed to set leaves: {e:?}");
-                }
-            })
-            .await?;
-
+        // The leaves carry no chains, so an already-stored chain is left alone
+        // rather than rewritten on every refresh. Collecting the chains of
+        // anything newly reported is the caller's to schedule.
+        let bare = |leaves: Vec<TreeNode>| -> Vec<LeafPedigree> {
```
roeierez: The chains are downloaded in the background and written through `store_ancestors`, so do the lifecycle methods need to take `LeafPedigree` at all? `bare()` here wraps leaves in empty pedigrees just to unwrap them in the stores, and `add_leaves`/`update_reservation` are the same (`insert_leaves` and `swap_leaves_internal` both pass ancestors: Vec::new()). Keeping them on &[TreeNode] and letting renewal route its parent through `store_ancestors` would remove the pedigree handling from all seven stores, removing advisory locks additions on 4 methods, and remove the need for the `id <> ALL($3)` expensive query in `set_leaves`.  WDYT?

roeierez (changes requested):

roeierez:
Have added 3 commits with the following changes:

* Removed the LeafPedigree from the tree store interface and back to the old interface. It turns out most of the methods passed empty slice for ancestor so that was a bit of a left over from the previous approach. Now we only store ancestor in one methods ( store_ancestors). That drops the added advisory locks and some complex queries.

* We don't persist partial chains, only complete. The allows us to simplify the query for missing ancestor by only lookup for the immediate parent rather than a path to the root.

* Batch the ancestor writes in the JS and SQLite stores

* Download exit chain trigger is now simplified. Spark wallet is responsible to emit "LeavesAdded" event which is the only trigger. It seems that wiring it to the high level operations makes it spread all over the place while this approach is focues on a single place. Also we keep the same preformance (even slightly better) since we only attempt to download if leaves were actually added (in contrary to payment where not always leaves are added).

crates/spark/src/tree/service.rs
```diff
@@ -791,17 +908,73 @@ impl SynchronousTreeService {
...
+}
+
+/// Whether `ancestors` is a chain that can back `leaf`'s unilateral exit. A leaf
+/// that is itself a root needs no ancestors.
+///
+/// Only the leaf's current parent is looked for. A stored chain always runs from
+/// some parent to a root, so containing the parent the leaf has now means it
+/// spans the whole path. Timelock renewal reparents a leaf onto a new split node
+/// under the same id, and this is what catches the chain it left behind.
+pub fn chain_backs_exit(leaf: &TreeNode, ancestors: &[TreeNode]) -> bool {
```
danielgranhao: Shouldn't this still check that the chain reaches a root? The rustdoc above states a stored chain always goes to a root, but this is also used on raw operator output (exit_chain_resolver.rs:25).

crates/spark/src/services/timelock_manager.rs
```diff
@@ -75,116 +72,168 @@ impl TimelockManager {
...
+        if let Some(split_node) = split_node {
+            nodes.insert(split_node.id.clone(), split_node);
         }
-
-        let renewed_nodes = futures::future::try_join_all(renew_futures).await?;
-        ready_nodes.extend(renewed_nodes);
-
-        Ok(ready_nodes)
+        nodes.insert(renewed_leaf.id.clone(), renewed_leaf.clone());
+        assemble_exit_chains(&nodes, std::slice::from_ref(&renewed_leaf.id))
```
danielgranhao, line 202: A zero-timelock renewal never fetches the leaf's ancestors, so when the leaf comes in without a chain (the usual case on a refresh, a claim or a swap) the chain rebuilt here is just the new split node. It still gets stored, and since that split node is the leaf's parent now, the leaf counts as having a complete chain and is never looked at again. It also overwrites whatever chain the leaf had before. So the leaf ends up un-exitable for good. Should zero-timelock leaves fetch their ancestors too?

If the comment on `chain_backs_exit` is addressed, I think this is caught later on and resolved.

crates/spark-mysql/src/tree_store.rs
```diff
@@ -526,7 +620,8 @@ impl TreeStore for MysqlTreeStore {
         id: &LeavesReservationId,
         leaves_to_keep: &[TreeNode],
     ) -> Result<(), TreeServiceError> {
-        // Scoped to a single `reservation_id`; row-level FK + MVCC suffice.
+        // No global write lock: this touches only the rows of one reservation
+        // and writes no ancestor rows.
```
danielgranhao: Don't we still need the lock given we may delete leaves? A chain write checks the leaf is still there and then writes its rows, holding the lock so nothing can delete that leaf in between. A cancel that skips the lock can land in that gap, leaving those rows behind with no leaf to own them, and nothing ever collects them.

 Same in `update_reservation`, and in both on Postgres.

crates/spark-sqlite/src/lib.rs
```diff
@@ -0,0 +1,1453 @@
...
+    // ---- row (de)serialization ----
+
+    fn row_to_node(data: &str) -> Result<TreeNode, TreeServiceError> {
+        serde_json::from_str(data).map_err(|e| generic("deserialize node", e))
+    }
+
+    // ---- ancestor persistence ----
+
+    /// Ids among `pedigrees` whose leaf the pool still holds, reserved or not.
+    fn existing_leaf_ids(
```
danielgranhao, line 219: This binds one parameter per leaf. We could end up hitting SQLite's limit per statement (32766). Same pattern in `delete_orphaned_ancestors` and `get_exit_chains`, the latter binding the ids twice so it gives out at half that.

Worth chunking?

For `delete_orphaned_ancestors` we don't need the ids. The leaf changes already ran in the same transaction, so `WHERE leaf_id NOT IN (SELECT id FROM brz_tree_leaves)` does the same job with no parameters at all.

crates/breez-sdk/wasm/js/node-tree-store/index.cjs
```diff
@@ -0,0 +1,1173 @@
...
+      this.db = null;
+    }
+  }
+
+  // ===== TreeStore Methods =====
+
+  /**
+   * Add leaves to the pool together with their ancestors. Receiving a leaf back
+   * clears any prior spent marker. Re-adding an id refreshes its mutable fields.
+   * @param {Array} leaves - Array of LeafPedigree { leaf, ancestors }
```
danielgranhao: These take plain leaves now rather than pedigrees, so the type is `TreeNode[]`, and the line above about storing ancestors is stale too. Same on `setLeaves`, `finalizeReservation` and `updateReservation`.

crates/spark-mysql/src/tree_store.rs
```diff
@@ -1560,6 +1843,96 @@ impl MysqlTreeStore {
...
+    ) -> Result<(), TreeServiceError> {
+        if rows.is_empty() {
+            return Ok(());
+        }
+
+        let mut sql = String::from(
+            "INSERT INTO brz_tree_ancestors \
+             (user_id, leaf_id, id, parent_node_id, status, data, value, verifying_public_key) VALUES ",
+        );
+        let mut params: Vec<Value> = Vec::with_capacity(rows.len() * 8);
```
danielgranhao: We might hit mysql's placeholder limit with just a few hundred leaves. Worth chunking IMO

crates/spark/src/tree/service.rs
roeierez: Done

crates/spark/src/services/timelock_manager.rs
roeierez: Done

crates/spark-sqlite/src/lib.rs
roeierez: Done

crates/spark-mysql/src/tree_store.rs
roeierez: Done

crates/spark-mysql/src/tree_store.rs
roeierez: Done

crates/breez-sdk/wasm/js/node-tree-store/index.cjs
roeierez: Done

crates/breez-sdk/wasm/js/node-tree-store/index.cjs
```diff
@@ -0,0 +1,1173 @@
...
+      // tagged by the owning leaf id, so a node id stored under another leaf
+      // can never cross-contaminate this one.
+      const placeholders = leafIds.map(() => "?").join(",");
+      const rows = this.db
+        .prepare(
+          `SELECT leaf_id, data FROM brz_tree_ancestors WHERE leaf_id IN (${placeholders})
+           UNION ALL
+           SELECT id AS leaf_id, data FROM brz_tree_leaves WHERE id IN (${placeholders})`
+        )
+        .all(...leafIds, ...leafIds);
```
dangeross: Should the wasm node implementation also chunk like the rust implementation? (thinking if `spark-mysql` should chunk too)

crates/spark/src/tree/exit_chain_resolver.rs
```diff
@@ -0,0 +1,567 @@
...
+        };
+        if candidates.is_empty() {
+            return Ok(());
+        }
+        trace!("Resolving exit chains for {} leaves", candidates.len());
+
+        let fetched = self
+            .tree_service
+            .fetch_pedigrees_from_operators(&candidates)
+            .await;
```
dangeross, line 114: I'm trying to figure out the memory load of handling a large set of leaves (10,000-100,000+ leaves) and ancestors the first time this is run with no stored chains yet. Seems to load them all into memory via paging and I'm wondering if its safer to do it in batches.

crates/spark/src/tree/service.rs
```diff
@@ -361,29 +419,24 @@ impl TreeService for SynchronousTreeService {
...
-                        .await
-                {
-                    error!("Failed to set leaves: {e:?}");
-                }
-            })
-            .await?;
-
+        // A refresh writes no chains, so an already-stored one is left alone
+        // rather than rewritten every minute. Collecting the chains of anything
+        // newly reported is the caller's to schedule.
```
dangeross: Reads like the caller has to do some action, but the service announces now via `notify_leaves_added` below.

crates/breez-sdk/wasm/js/node-tree-store/index.cjs
roeierez: Good point. Done.

crates/spark/src/tree/exit_chain_resolver.rs
roeierez: True, now loaded in chunks.

crates/spark/src/tree/service.rs
roeierez: fixed.

dangeross (approved):
LGTM

danielgranhao (approved):
LGTM

PR #1012, opened 2026-07-22
Add instant (0-conf) static deposit claims

### LINKED ISSUE #1006: Support zero conf deposits
https://docs.spark.money/wallets/0-conf#0-conf-deposits

dangeross:
The Spark Service Provider fronts a deposit's credit before it confirms, for a spread, so funds are usable immediately instead of waiting for confirmations. The spread is a percentage (~3%) + a flat sats amount (~100 sats) + on-chain fee rate component, so a single BPS cap admits large deposits and lets small ones (and high-fee periods) fall through to the normal wait-for-maturity claim rather than overpaying for speed.

- `Config.max_instant_deposit_claim_fee_bps`: opt-in ceiling on the SSP spread, in basis points of the deposit value. When set, the background sync instant-claims each non-mature deposit whose spread is within it.
- `ClaimDepositRequest.max_instant_fee_bps`: claim a specific not-yet-mature deposit instantly on demand.
- **Breaking** `ClaimDepositResponse.payment` is now optional: Unset for an instant claim (settles asynchronously), set for a standard claim.

Closes #1006


crates/breez-sdk/core/src/sdk/sync.rs
```diff
@@ -484,6 +599,64 @@ impl BreezSdk {
         );
         Ok(transfer_id)
     }
+
+    /// Attempts an instant 0-conf static deposit claim for `detailed_utxo`.
+    /// `Ok(Submitted)` on a submitted claim, `Ok(Declined)` for a terminal
+    /// outcome (no 0-conf plan, spread over the ceiling, or a failed claim
+    /// submission), and `Err` only for a failed quote fetch, which is transient
+    /// (the SSP may not have indexed the mempool tx yet) and should be retried.
+    pub(super) async fn instant_claim_utxo(
```
roeierez: NIT: perhaps this function should be in deposits.rs file?

crates/breez-sdk/core/src/sdk/sync.rs
```diff
@@ -501,3 +674,173 @@ impl BreezSdk {
...
+        .find(|p| p.confirmations == 0)
+    else {
+        return InstantClaimPlan::NoPlan;
+    };
+    // Price the spread off the selected plan's `amount` (the credit for the plan
+    // we chose), not the quote-level credit. They are equal today (the SSP
+    // returns one plan per quote, verified on deployed regtest), but keying on the
+    // plan stays correct if the SSP ever returns multiple plans with differing
+    // amounts.
+    let deposit_sats = quote_result.quote.deposit_amount.original_value;
```
roeierez: Should we trust the SSP for the amount? Seems better to take it from the utxo we already have?

crates/breez-sdk/core/src/sdk/deposits.rs
```diff
@@ -32,6 +32,51 @@ impl BreezSdk {
...
+                        )
+                        .await?;
+                    Ok(ClaimDepositResponse { payment: None })
+                }
+                Ok(InstantClaimOutcome::Declined(e)) => {
+                    // Terminal outcome (no 0-conf plan, spread over the ceiling,
+                    // or a failed submission whose outcome is unknown): mark, so
+                    // that if instant is also enabled in config the background
+                    // sync does not re-submit into the pre-swap window.
+                    error!("Instant claim declined: {e:?}");
```
roeierez: Do you see an option to reuse this block with the function `instant_claim_utxo_and_resolve_deposit` ?

crates/breez-sdk/core/src/sdk/deposits.rs
```diff
@@ -32,6 +32,51 @@ impl BreezSdk {
...
+                    self.storage
+                        .update_deposit(
```
roeierez: update deposit uses UPDATE statement and won't update the row if it doesn't exist yet. The deposit only added on sync so seems like it is possible to call claim_deposit and this update ends up with NOP.

crates/breez-sdk/core/src/sdk/deposits.rs
```diff
@@ -32,6 +32,51 @@ impl BreezSdk {
                 .fetch_detailed_utxo(&request.txid, request.vout)
                 .await?;
 
+        if let Some(max_instant_fee_bps) = request.max_instant_fee_bps {
+            return match self
+                .instant_claim_utxo(&detailed_utxo, max_instant_fee_bps)
```
roeierez: Should we check the `instant_claim_attempted` here before attempting to claim?
Perhaps instead of boolean `instant_claim_attempted` we should have a status?
e.g: `Declined | Submitted(claim_id)` ? So we can skip all zero-conf on Submitted?

crates/breez-sdk/core/src/sdk/sync.rs
```diff
@@ -310,35 +323,46 @@ impl BreezSdk {
...
-            .map(|(u, _)| u)
-            .collect();
-
         let mut claimed_deposits: Vec<DepositInfo> = Vec::new();
         let mut unclaimed_deposits: Vec<DepositInfo> = Vec::new();
-        for detailed_utxo in to_claim {
-            match self
-                .claim_utxo(&detailed_utxo, self.config.max_deposit_claim_fee.clone())
+        for (detailed_utxo, is_mature) in all_utxos {
+            let res = if is_mature {
```
roeierez, line 370: Also here I believe there is a race where a zero-conf already confirmed before the ssp swap lands and it looks like we are attempting to claim.
Perhaps the statuses I suggested in the previous comment can work here as well where we skip Submitted (or return the right result from the existing in flight row).

roeierez (changes requested):

roeierez (approved):
LGTM

crates/breez-sdk/breez-itest/tests/deposit_withdraw.rs
```diff
@@ -641,3 +644,150 @@ async fn test_deposits_to_multiple_addresses(
...
+/// first the SSP offers no 0-conf plan, and the test logs it and skips (nothing to
+/// assert). When the window is caught it asserts no synchronous payment, a
+/// credited balance net of the SSP spread, and the deposit leaving the unclaimed
+/// list. A rejected statement is neither a confirmation nor transient, so it fails
+/// fast. Requires faucet creds and the deployed regtest SSP to have 0-conf enabled.
+#[rstest]
+#[test_log::test(tokio::test)]
+async fn test_manual_instant_deposit_claim(
+    #[future] bob_strict_fee_sdk: Result<SdkInstance>,
+) -> Result<()> {
```
danielgranhao, line 666: Do you think it would be worth changing to instant deposits for all funding steps in itests? If it speeds up itest runtime it would be nice. We would still need the regular static deposit flow to be tested but it would become the odd one instead of being the norm.

crates/breez-sdk/core/src/sdk/sync.rs
```diff
@@ -277,13 +307,17 @@ impl BreezSdk {
...
             .iter()
             .map(|d| TxOutput {
                 txid: d.txid.clone(),
                 vout: d.vout,
             })
             .collect();
+        // Instant-claim status per deposit: the one-shot instant path is not
+        // retried once any status is set, and a mature deposit whose instant claim
+        // is still `Submitted` is not claimed while that claim is in flight.
+        let instant_status = instant_claim_status_map(&existing_deposits);
```
danielgranhao: This snapshot is taken before we sync chain deposits, which might be "slow" as it does a round-trip per UTXO. A manual instant claim could land during that time and would not be in the snapshot, potentially resulting in a normal claim being attempted on top. Moving this below `DepositChainSyncer::sync()` would prevent that

crates/breez-sdk/core/src/sdk/sync.rs
```diff
@@ -310,35 +344,53 @@ impl BreezSdk {
...
+                    &mut claimed_deposits,
+                    &mut unclaimed_deposits,
+                )
+                .await
+            } else {
+                // Not yet mature: attempt a one-shot 0-conf instant claim if enabled.
+                // Skip if an instant claim was already attempted (declined or submitted).
+                if instant_status.contains_key(&key) {
+                    continue;
                 }
```
danielgranhao: This skips on any status. A manual instant claim attempt that fails with `FeeExceeded` prevents later background attempts even if the config max fee is higher. For `NoPlan` we also won't try anymore. Are we sure the SSP can't return that temporarilly?

crates/spark/src/services/deposit.rs
```diff
@@ -1003,6 +1049,106 @@ impl DepositService {
...
+        }
+
+        // Raw bytes of the SSP quote signature (hex) go into the user statement.
+        let quote_signature_bytes = hex::decode(&quote.quote_signature)
+            .map_err(|e| ServiceError::Generic(format!("invalid quote signature hex: {e}")))?;
+
+        // The statement names the address the UTXO actually paid to, derived from
+        // the funding output. The wallet's current static address may have rotated
+        // since the deposit landed, so it cannot be regenerated here.
+        let output_index = quote.output_index as usize;
```
danielgranhao, line 1118: We could validate that the quote vout matches the vout we asked for

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -1104,12 +1126,20 @@ pub struct ClaimDepositRequest {
     pub vout: u32,
     #[cfg_attr(feature = "uniffi", uniffi(default=None))]
     pub max_fee: Option<MaxFee>,
+    /// Set to request an instant (0-conf) claim instead of waiting for the
+    /// deposit to mature, bounding the SSP spread at this many basis points of
+    /// the deposit value (100 bps = 1%).
```
danielgranhao: Nit: let's document that setting this causes `max_fee` to be ignored

danielgranhao (commented):
Looks good

crates/breez-sdk/breez-itest/tests/deposit_withdraw.rs
dangeross: I'd like to do this as a follow up PR, if that's ok

crates/breez-sdk/core/src/sdk/sync.rs
dangeross: I've added retrying `FeeExceeded` if the attempting bps is higher than the already tried one. I haven't seen `NoPlan` being transient, but we could add a reties field to the enum variant to cap the attempts. But I don't think it's needed. Wdyt?

danielgranhao: Sure, let's skip it for now.

danielgranhao (approved):
LGTM

crates/breez-sdk/breez-itest/tests/deposit_withdraw.rs
kingonly: I don't think 0-conf is going to be the default flow, so don't rush to change the tests. The flow ofc needs to be tested, but keep in mind it requires a direct business relationship with Lightspark (it's a paid service).

roeierez: True not by default but I like Daniel's idea to use it in our integration tests as it will save a lot of time to run these tests (Leave one test that test specific non zero conf deposit). Follow up PR is great Ross.

PR #1013, opened 2026-07-22
Batch token sending support

### LINKED ISSUE #999: Support sending tokens to multiple recipients at once
The public `BreezSdk` API can only transfer tokens to a single recipient per call (`sendPayment`), but the underlying layers already support sending to multiple recipients in one transaction.

`SparkWallet::transfer_tokens` accepts a `Vec<TransferTokenOutput>`, but the public path calls it with a hardcoded single-element vec (`send/spark_address.rs:170`, `payments/client_signing.rs:269`).

**Proposed:** Let callers send tokens to multiple recipients at once, backed by the existing `Vec` path. Exact API shape needs analysis.

danielgranhao:
Closes #999

Allows sending tokens to multiple recipients, including a mix of tokens and invoices/addresses in a single Spark token tx.

Changes:
-  Add `prepare_send_token_batch` / `send_token_batch`API to the SDK, returning one payment per recipient sharing a tx hash
- Support multiple token IDs and multiple fulfilled invoices in one token transfer, with one atomic output reservation across all stores
- Unify address and invoice token recipients into a single TokenRecipient type in spark-wallet
- Attribute each fulfilled invoice to exactly one payment, deterministically on both sender and receiver
- Support batch sends through the external (client) signing flow
- Name the short token in insufficient funds errors, surfaced through SdkError::InsufficientFunds


crates/breez-sdk/core/src/utils/token.rs
```diff
@@ -206,14 +219,60 @@ pub async fn token_transaction_to_payments(
...
+/// only the invoice string, and attachments are sorted before broadcast, so even
+/// their order says nothing about which output pays which. The match is therefore
+/// inferred from payee, token and amount, preferring an exact amount over an
+/// invoice that left the amount to the sender.
+///
+/// Two invoices from the same payee for the same token and amount are
+/// indistinguishable, so either assignment is equally correct. Which one is taken
+/// depends on the order of `unmatched`, so callers must put it in a canonical
+/// order first, or two wallets seeing the same transaction from different sources
+/// would attribute those invoices differently.
```
dangeross, line 246: Does this hold for address-vs-invoice too? An address output and an invoice output to the same payee, token and amount look identical to the match below, so the address recipient can consume the invoice.

crates/spark/src/token/mod.rs
```diff
@@ -256,29 +282,29 @@ pub trait TokenOutputStore: Send + Sync {
...
+        outputs_to_add: &TokenOutputs,
     ) -> Result<(), TokenOutputServiceError>;
 
+    /// Reserves outputs covering every target in one atomic step, so a transaction
+    /// spanning several tokens holds a single reservation rather than one per token.
     async fn reserve_token_outputs(
         &self,
-        token_identifier: &str,
-        target: ReservationTarget,
+        targets: &[(String, ReservationTarget)],
```
dangeross, line 297: Should we document that `targets` is a unique, order-significant list of token identifier / reservation target pairs? (Order carries into the final input ordering.) Same contract on `select_token_outputs`, and on both methods of `TokenOutputService` below.

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -1788,6 +1801,79 @@ pub struct SendPaymentRequest {
...
+/// A recipient after prepare has resolved any invoice destination.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct ResolvedTokenBatchRecipient {
+    pub destination: String,
+    /// Amount in the token's base units.
+    pub amount: u128,
+    pub token_identifier: String,
+    /// Present when `destination` was an invoice.
+    pub invoice_details: Option<SparkInvoiceDetails>,
```
roeierez: Perhaps better to have an enum here?  Sparinvoice/Address and omit the detination?

crates/breez-sdk/core/src/sdk/payments/mod.rs
```diff
@@ -91,6 +93,52 @@ impl BreezSdk {
...
+    /// Look for the batch with a `Token` payment details filter on the
+    /// transaction hash before sending it again.
+    #[instrument(level = "info", target = "breez_sdk_core::send_token_batch", skip_all)]
+    pub async fn send_token_batch(
+        &self,
+        request: SendTokenBatchRequest,
+    ) -> Result<SendTokenBatchResponse, SdkError> {
+        self.maybe_ensure_spark_private_mode_initialized().await?;
+        Box::pin(send::token_batch::send(self, request)).await
+    }
```
dangeross: I noticed in the JS spark wallet they a have a [fn to send transfers](https://github.com/buildonspark/spark/blob/e6bc4343ded7ea20f5aa5c79e65d075c3cecc846/sdks/js/packages/spark-sdk/src/spark-wallet/spark-wallet.ts#L3058-L3167) also to multiple recipients. Do you think its worth shaping (mainly naming) the API so we can incorporate this also in the future?

dangeross (commented):
Looking good

crates/breez-sdk/core/src/sdk/payments/send/mod.rs
```diff
@@ -26,6 +27,9 @@ pub(in crate::sdk) enum PublishOutcome {
     SwapCompleted,
     Sent(SendPaymentResponse),
     Replayed(SendPaymentResponse),
+    /// A batch package: one payment per recipient, in recipient order.
+    SentBatch(Vec<Payment>),
```
roeierez: This enum is internal. Does it make sense to use Vec<Payments> for all variants? And then we don't need the SentBatch variant?

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -1788,6 +1801,79 @@ pub struct SendPaymentRequest {
     pub idempotency_key: Option<String>,
 }
 
+/// A single payee in a batch token send.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct TokenBatchRecipient {
+    /// Spark address or Spark invoice identifying the payee.
+    pub destination: String,
```
roeierez: NIT: should it be "payment_request" like in the single token pay version?

roeierez (changes requested):
Looks good!

crates/breez-sdk/core/src/models/mod.rs

crates/breez-sdk/core/src/sdk/payments/send/mod.rs
danielgranhao: Done

crates/breez-sdk/core/src/models/mod.rs

crates/spark/src/token/mod.rs
danielgranhao: Done

crates/breez-sdk/core/src/utils/token.rs
danielgranhao: Yes, I've updated the rustdoc accordingly

crates/breez-sdk/core/src/sdk/payments/mod.rs
danielgranhao: Good point. Done

roeierez (approved):
LGTM

dangeross (approved):
LGTM. Just needs a rebase

PR #1014, opened 2026-07-23
Remove SQLite support from the LNURL server

danielgranhao:


roeierez (approved):

JssDWt (approved):
I think it would be good to remove sqlx as a dependency now. I think that change is not very large. Just have to do some magic on migrating the migrations. Once we do that, we can re-add the lnurl server to the workspace. That would be nice.

dangeross (approved):
LGTM, with a nit

There is a hanging comment reference to a `SQLite` implementation [here](https://github.com/breez/spark-sdk/blob/7e51003bce6ea5e1ffe2981a7311ebc059c4a727/crates/breez-sdk/lnurl/src/repository.rs#L188-L191).

PR #1017, opened 2026-07-27
feat: expose the Spark master identity public key user setting

danielgranhao:
This exposes Spark's `master_identity_public_key` wallet setting as a user setting on the Breez SDK API. Setting it allows the defined identity to have read access to the current instance identity even when it is in private mode, enabling watch-only clients that can show balance and payment history.

### DIFF
--- a/crates/breez-sdk/cli/src/command/mod.rs
+++ b/crates/breez-sdk/cli/src/command/mod.rs
@@ -17,8 +17,8 @@ use breez_sdk_spark::{
     PaymentRequest, PaymentStatus, PaymentType, PrepareLnurlPayRequest, PrepareSendPaymentRequest,
     ReceivePaymentMethod, ReceivePaymentRequest, RefundDepositRequest,
     RegisterLightningAddressRequest, SendPaymentMethod, SendPaymentOptions, SendPaymentRequest,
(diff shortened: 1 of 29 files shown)


crates/breez-sdk/cli/src/command/mod.rs
```diff
@@ -367,6 +367,15 @@ pub enum Command {
...
         spark_private_mode_enabled: Option<bool>,
+
+        /// Hex encoded public key to designate as the wallet's master identity,
+        /// allowing its holder to read the wallet under private mode.
+        #[clap(short = 'm', long = "master-key", conflicts_with = "clear_master_key")]
+        master_key: Option<String>,
+
+        /// Remove the wallet's designated master identity.
+        #[clap(long = "clear-master-key", action = clap::ArgAction::SetTrue)]
+        clear_master_key: bool,
```
dangeross: This is the only nit I could find 😅
```suggestion
        #[clap(short, long, conflicts_with = "clear_master_key")]
        master_key: Option<String>,

        /// Remove the wallet's designated master identity.
        #[clap(long, action = clap::ArgAction::SetTrue)]
        clear_master_key: bool,
```

dangeross (approved):
LGTM

roeierez (approved):
LGTM

PR #1018, opened 2026-07-28
Seed local operator clusters with pre-generated keyshares

JssDWt:
Local clusters ran DKG at startup, which stalls when several clusters share a CI runner and fails the fixture with `Timeout waiting for keyshares to be available` ([example](https://github.com/breez/spark-sdk/actions/runs/30302305965/job/90097946665)).

The keyshares are now pre-generated: a set captured from one real DKG round is committed under `crates/spark-itest/keyshares/` and copied into each operator database before boot, so no cluster runs DKG. Regenerate with `make capture-itest-keyshares`.

`local_wallet_tests`: 202s -> 21s.


roeierez (approved):

roeierez (approved):
Good one!

danielgranhao (approved):
Nice!

PR #1021, opened 2026-07-29
Drop sqlx and rejoin lnurl crate into workspace

danielgranhao:
Follow up from #1014 - https://github.com/breez/spark-sdk/pull/1014#pullrequestreview-4765511280


crates/breez-sdk/lnurl/src/postgresql/migrations.rs
```diff
@@ -0,0 +1,441 @@
...
+                )"
+            .to_string(),
+            "ALTER TABLE newly_paid ADD COLUMN claimed_at BIGINT".to_string(),
+        ],
+        vec![
+            "ALTER TABLE users DROP COLUMN lnurl_private_mode_enabled".to_string(),
+            "ALTER TABLE users DROP COLUMN nostr_pubkey".to_string(),
+            "ALTER TABLE users DROP COLUMN webhook_secret".to_string(),
+        ],
+        vec!["CREATE INDEX idx_users_pubkey ON users(pubkey)".to_string()],
```
JssDWt: This one has to go before the webhook_secret column, because the date in the filename / added date.

JssDWt (commented):
One thing I caught this morning. Haven't finished the review yet.

crates/breez-sdk/lnurl/src/postgresql/migrations.rs
danielgranhao: Thanks, fixed!

crates/breez-sdk/lnurl/Cargo.toml
```diff
@@ -1,53 +1,44 @@
 [package]
 name = "lnurl"
 edition = "2024"
-version = "0.1.0"
-
-[workspace]
+version.workspace = true
```
dangeross, line 4: Dropping `[workspace]` means the `Dockerfile` release build now inherits the workspace `[profile.release]` (`panic = "abort"`, `opt-level = "z"`, `lto = true`). As its own workspace it was getting cargo's defaults (`opt-level = 3`, `panic = "unwind"`). I'm mainly unsure of the behavior change when using `panic = "abort"` in a server environment.

Do you think we should add a separate profile for the server?

crates/breez-sdk/lnurl/src/postgresql/migrations.rs
```diff
@@ -0,0 +1,444 @@
...
+    }
+
+    let previous_tracker_exists: bool = tx
+        .query_one(table_exists_sql, &[&"_sqlx_migrations"])
+        .await
+        .map_err(map_db_error)?
+        .get(0);
+    if !previous_tracker_exists {
+        return Ok(());
+    }
```
dangeross, line 97: Is it risky keying this only on `_sqlx_migrations` existing? It's hardcoded in sqlx's postgres driver, so any sqlx app sharing the schema has exactly that table. A partner self hosting could point a new install at a database they already use and we'd adopt their count as ours.
- Below ours: we resume part way through the list and fail on the first statement touching a table that was never created. That rolls back, but the adopted rows are already committed, so it fails to start and keeps failing to start.
- At or past ours: everything is marked applied, nothing runs, and we start up clean against a schema with no lnurl tables in it.

crates/breez-sdk/lnurl/Cargo.toml
danielgranhao: Done

crates/breez-sdk/lnurl/src/postgresql/migrations.rs
danielgranhao: Good catch. Now keyed on the migration versions themselves rather than the row count, so a foreign or shared _sqlx_migrations can't be adopted as ours.

crates/breez-sdk/lnurl/src/postgresql/migrations.rs
```diff
@@ -0,0 +1,753 @@
...
+    20_260_314,         // multi_instance
+    20_260_323,         // drop_unused_user_columns
+    20_260_326_110_000, // rename_pending_zap_receipts
+    20_260_326_115_000, // invoice_domain_columns
+    20_260_326_120_000, // domain_webhooks
+    20_260_416_120_000, // webhook_signing
+    20_260_709_000_000, // domain_attribution
+    20_260_803_120_000, // used_signed_messages
+    20_260_814_120_000, // released_names
+    20_260_826_120_000, // address_registrations
```
dangeross, line 50: Looks like these three still exist as `.sql` files under `migrations/postgres/`, while the older ones are deleted. Looks like a rebase issue.

crates/breez-sdk/lnurl/README.md
```diff
@@ -46,7 +46,7 @@ brew install protobuf openssl pkg-config
 From the repository root:
 
 ```shell
-cargo build --release --manifest-path crates/breez-sdk/lnurl/Cargo.toml
+cargo build --release -p lnurl
```
dangeross: This still builds with the workspace `[profile.release]`, so it's the `panic = "abort"` binary the Dockerfile just moved away from. The path below and the `Native (Rust)` examples point at `target/release/lnurl` too.

Should these use `--profile release-lnurl`?

dangeross (approved):
LGTM

roeierez (approved):

PR #1022, opened 2026-07-29
fix(android): handle passkey indexing delay

erdemyerebasmaz:
Android can briefly fail to find a passkey immediately after creating it because Google Password Manager indexing is asynchronous.

Instead of waiting a fixed delay, the SDK now retries the lookup for up to 5 seconds. If the passkey still cannot be derived, the created passkey is returned so hosts can continue without creating a duplicate.

This makes the race condition harmless while avoiding unnecessary extra passkeys.

**Integrators**: `PasskeyError` gains a new variant, so exhaustive `when` / `switch` statements need an additional arm.

**Follow-up:** requesting the PRF value during passkey creation would remove this second step entirely on supported providers, but requires a provider interface change across all platforms. See #1024.

### DIFF
--- a/crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
+++ b/crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
@@ -82,12 +82,14 @@ private const val CORE_TAG = "PasskeyPrfCore"
 // =====================================================================
 
 /**
- * A newly-registered passkey is briefly not ready for the immediate
- * post-create assertion: Credential Manager can drop `prf.second` from a
- * dual-salt assertion (forcing a second single-salt prompt) or surface
- * the credential as not yet discoverable in the picker. [arm] opens a
- * window in which the next derive re-asks instead of trusting the first
- * miss.
+ * A newly-registered passkey is briefly not discoverable, so the
+ * immediate post-create assertion can miss a credential that exists.
+ * [arm] opens a window in which the next derive re-asks instead of
+ * trusting that first miss.
+ *
+ * Scoped to that case only: the retry catches `NoCredentialException`,
+ * so a dropped `prf.second` is not covered and still falls through to
+ * the single-salt recover in [CredentialManagerPrfCore.deriveSeeds].
  *
  * The window is a ceiling, not a wait: indexing takes as long as the
  * device takes, so a fixed sleep either returns before the credential is
--- a/crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
+++ b/crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
@@ -85,8 +85,14 @@ private const val CORE_TAG = "PasskeyPrfCore"
  * A newly-registered passkey is briefly not ready for the immediate
  * post-create assertion: Credential Manager can drop `prf.second` from a
  * dual-salt assertion (forcing a second single-salt prompt) or surface
- * the credential as not yet discoverable in the picker. Holding the next
- * derive up to [DEFAULT_DURATION_MS] lets the OS finish indexing.
+ * the credential as not yet discoverable in the picker. [arm] opens a
+ * window in which the next derive re-asks instead of trusting the first
+ * miss.
+ *
+ * The window is a ceiling, not a wait: indexing takes as long as the
+ * device takes, so a fixed sleep either returns before the credential is
+ * queryable or taxes every registration. Retrying costs nothing once the
+ * credential resolves.
  *
  * Mirrors iOS `PostCreateGraceTracker`; an instance lives inside
  * [CredentialManagerPrfCore] so every consumer that holds onto a single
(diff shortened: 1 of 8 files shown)


crates/breez-sdk/core/src/passkey/error.rs
```diff
@@ -118,6 +118,20 @@ pub enum PasskeyError {
...
+    /// behind, owning a wallet nothing points to.
+    ///
+    /// `kind` classifies the derive failure that followed the create, so
+    /// hosts keep the branching they would have had on the inner error.
+    #[error("Passkey created but derivation failed: {reason}")]
+    CreatedButNotDerived {
+        credential_id: Vec<u8>,
+        kind: ErrorKind,
+        reason: String,
+    },
```
dangeross: A cancel arrives two different ways now, depending where the user hit it: `Prf(UserCancelled)` from the sign-in probe, `CreatedButNotDerived { kind: Cancel }` from the register leg's derive prompt. Same button in `connect_with_passkey`, two shapes to unwrap.

Could we carry the inner error instead?

```suggestion
    #[error("Passkey created but derivation failed: {source}")]
    CreatedButNotDerived {
        credential_id: Vec<u8>,
        source: PrfProviderError,
    },
```

Wrap only `Prf(_)` and let `InvalidSalt` / `InvalidPrfOutput` / `MnemonicError` / `KeyDerivationError` through unchanged, then hosts unwrap once and keep the arms they already have.

The doc above needs a look either way. `kind` doesn't give hosts "the branching they would have had on the inner error": anything outside `Prf(_)` collapses to `Internal` via the `_` arm, so `InvalidSalt` and `MnemonicError` land as the same value and the real variant only survives in `reason`.  With `source` that sentence is true as written.

crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
```diff
@@ -85,8 +85,14 @@ private const val CORE_TAG = "PasskeyPrfCore"
...
- * the credential as not yet discoverable in the picker. Holding the next
- * derive up to [DEFAULT_DURATION_MS] lets the OS finish indexing.
+ * the credential as not yet discoverable in the picker. [arm] opens a
+ * window in which the next derive re-asks instead of trusting the first
+ * miss.
+ *
+ * The window is a ceiling, not a wait: indexing takes as long as the
+ * device takes, so a fixed sleep either returns before the credential is
+ * queryable or taxes every registration. Retrying costs nothing once the
+ * credential resolves.
```
dangeross: The retry only catches `NoCredentialException`, so the `prf.second` half of this sentence has nothing behind it now. That's the case the post-create derive actually hits too: `setup_wallet` sends `[account_master, label]`, so a dropped `second` falls through to the single-salt recover in `deriveSeeds` and costs the user a second prompt. It used to get up to 800 ms of settling first.

Did we ever see the wait helping there, or was it only ever the not-yet-discoverable case? If it's the latter, worth taking `prf.second` out of the sentence so the doc matches what the window does.

crates/breez-sdk/core/src/passkey/passkey_client.rs
```diff
@@ -212,7 +212,7 @@ impl PasskeyClient {
             .create_passkey(request.exclude_credentials.unwrap_or_default())
```
dangeross, line 218: `validate_label` doesn't run until `setup_wallet`, so an empty or over-long label creates a real passkey first and only then fails. Seems like a pre-existing issue, but that failure now comes back as `CreatedButNotDerived`, which signals to recover by signing in pinned to `credential_id`, which runs the same validation on the same label.

Can we validate before we create anything?

dangeross (commented):
Not in the diff, but passkey_onboarding.md needs looking at also. It still says every passkey failure normalizes to a PrfProviderError variant, and the table under it maps UserCancelled to a sticky retry, UserTimedOut to "do not auto-retry", PrfNotSupported to mnemonic fallback. None of those arms match on the register path any more. A when / switch with an else won't complain either, it'll just fall through to the generic branch.

crates/breez-sdk/core/src/passkey/error.rs
erdemyerebasmaz: Took the suggestion: it carries the PrfProviderError as `source` and wraps only `Prf(_)`, so the other variants pass through unwrapped. Addressed in 5d9496279.

crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
erdemyerebasmaz: Only ever the not-yet-discoverable case. Dropped `prf.second` from the sentence. Addressed in 5d9496279.

crates/breez-sdk/core/src/passkey/passkey_client.rs
erdemyerebasmaz: The label is now resolved and validated before the create. Addressed in 5d9496279.

docs/breez-sdk/src/guide/passkey_onboarding.md
```diff
@@ -89,6 +89,28 @@ Every passkey failure normalizes to a {{#name PrfProviderError}} variant. Match
 | {{#enum PrfProviderError::Configuration}} | Entitlement missing, AASA stale, or assetlinks malformed | Developer-facing error; surface the {{#enum PasskeyAvailability::NotAssociated}} reason. |
 | {{#enum PrfProviderError::Generic}} | Network or generic failure | Generic "try again later" UI. |
 
+Those rows cover {{#name PasskeyClient.sign_in}}, and
+{{#name PasskeyClient.register}} up to the point the credential is created.
+Once it exists, every failure arrives as the variant below instead, whatever
+caused it: a cancel, a timeout or an unsupported authenticator during the
+derive no longer surfaces as its own `PrfProviderError`.
```
dangeross: Contradicts the end of the next paragraph, which says mnemonic / key derivation / invalid PRF output keep their own variant. That one matches the code. The examples here are all `PrfProviderError`, so it's only the generalisation that's wrong.

```suggestion
Once it exists, a failure from the authenticator arrives as the variant below
instead: a cancel, a timeout or an unsupported authenticator during the
derive no longer surfaces as its own `PrfProviderError`.
```

crates/breez-sdk/core/src/passkey/mod.rs
```diff
@@ -220,12 +220,20 @@ impl Passkey {
     /// [`Self::store_label`] need no extra PRF prompts.
     /// `publish_label = false` skips the Nostr write: used by
     /// speculative cold-restore.
```
dangeross: This doc block need moving above `setup_wallet`

dangeross (commented):
Looks good, just some nits

docs/breez-sdk/src/guide/passkey_onboarding.md
erdemyerebasmaz: Took the suggestion. Addressed in a0036b379.

crates/breez-sdk/core/src/passkey/mod.rs
erdemyerebasmaz: Addressed in a0036b379.

dangeross (approved):
LGTM

PR #1024, opened 2026-07-29
feat: derive wallet during passkey creation

erdemyerebasmaz:
Passkey creation now requests the PRF value in the same ceremony, instead of requiring a follow-up assertion to derive from.

This removes the timing window where a newly created passkey exists but cannot yet be found, and costs one auth prompt instead of two where the authenticator evaluates PRF at create. An authenticator that returns nothing, or fewer outputs than salts, falls back to the existing assertion flow.

Supported on Android, iOS, web, React Native and Flutter.

#### Breaking changes:
- `PrfProvider.create_passkey` now takes salts and returns the credential plus optional seeds.
  - Rust implementations get a defaulted method; custom foreign implementations must match the new signature.
- The JS provider contract and the Dart `PrfProvider.createPasskey` change the same way.
  - Hosts using a built-in provider are unaffected on every platform.

#### Verified on device:
- Android and iOS both complete registration in a single prompt.
- Creating, logging out and signing back in resolves the same recovery phrase, so the value returned at create matches what an assertion returns for the same salts. A mismatch there would leave the wallet created at registration unreachable, and nothing in the code can detect it.

### DIFF
--- a/crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/PasskeyProvider.kt
+++ b/crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/PasskeyProvider.kt
@@ -1,6 +1,7 @@
 package technology.breez.spark.passkey
 
 import android.app.Activity
+import breez_sdk_spark.CreatePasskeyOutput
 import breez_sdk_spark.DeriveSeedsOutput
 import breez_sdk_spark.DeriveSeedsRequest
 import breez_sdk_spark.DomainAssociation
@@ -142,10 +143,31 @@ public class PasskeyProvider(
      * registered IDs in `excludeCredentials` so the platform refuses a
      * duplicate even after reinstall.
      */
-    override suspend fun createPasskey(excludeCredentials: List<ByteArray>): PasskeyCredential {
+    override suspend fun createPasskey(excludeCredentials: List<ByteArray>): PasskeyCredential =
+        createPasskeyWithPrf(excludeCredentials, emptyList()).credential
+
+    /**
+     * Registers with `prf.eval` requested, so an authenticator that
+     * evaluates PRF at create returns the seeds here and the caller needs
+     * no assertion. Credential Manager reports a credential it has not
+     * finished indexing as absent, so removing that assertion removes the
+     * only step that can trip over it.
+     *
+     * `seeds` stays null on authenticators that report PRF support without
+     * evaluating, which is every pre-eval-at-create implementation. The
+     * caller then derives as it always has.
+     */
+    override suspend fun createPasskeyWithPrf(
+        excludeCredentials: List<ByteArray>,
+        salts: List<String>,
+    ): CreatePasskeyOutput {
         try {
-            val c = core.register(excludeCredentials)
-            return PasskeyCredential(c.credentialId, c.userId, c.aaguid, c.backupEligible)
+            val registration = core.register(excludeCredentials, salts)
+            val c = registration.credential
+            return CreatePasskeyOutput(
+                PasskeyCredential(c.credentialId, c.userId, c.aaguid, c.backupEligible),
+                registration.seeds,
+            )
         } catch (e: CredentialManagerPrfCoreException) {
             throw e.toPrfProviderException()
         }
--- a/crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/PasskeyProvider.kt
+++ b/crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/PasskeyProvider.kt
@@ -136,17 +136,14 @@ public class PasskeyProvider(
     }
 
     /**
-     * Register a new passkey with PRF support. One ceremony, no derivation.
+     * Registers with `prf.eval` requested, so an authenticator that
+     * evaluates PRF at create returns the seeds here and the caller needs
+     * no assertion.
      *
      * `user.id` is never host-supplied: the core mints a fresh random 16-byte
      * handle and returns it as [PasskeyCredential.userId]. Pass already-
      * registered IDs in `excludeCredentials` so the platform refuses a
-     * duplicate even after reinstall.
-     */
-    /**
-     * Registers with `prf.eval` requested, so an authenticator that
-     * evaluates PRF at create returns the seeds here and the caller needs
-     * no assertion. Credential Manager reports a credential it has not
+     * duplicate even after reinstall. Credential Manager reports a credential it has not
      * finished indexing as absent, so removing that assertion removes the
      * only step that can trip over it.
      *
--- a/crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/PasskeyProvider.kt
+++ b/crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/PasskeyProvider.kt
@@ -162,7 +162,7 @@ public class PasskeyProvider(
         salts: List<String>,
     ): CreatePasskeyOutput {
         try {
-            val registration = core.register(excludeCredentials, salts)
+            val registration = core.registerWithPrf(excludeCredentials, salts)
             val c = registration.credential
             return CreatePasskeyOutput(
                 PasskeyCredential(c.credentialId, c.userId, c.aaguid, c.backupEligible),
--- a/crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/PasskeyProvider.kt
+++ b/crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/PasskeyProvider.kt
@@ -143,9 +143,6 @@ public class PasskeyProvider(
      * registered IDs in `excludeCredentials` so the platform refuses a
      * duplicate even after reinstall.
      */
-    override suspend fun createPasskey(excludeCredentials: List<ByteArray>): PasskeyCredential =
-        createPasskeyWithPrf(excludeCredentials, emptyList()).credential
-
     /**
      * Registers with `prf.eval` requested, so an authenticator that
      * evaluates PRF at create returns the seeds here and the caller needs
@@ -157,12 +154,12 @@ public class PasskeyProvider(
      * evaluating, which is every pre-eval-at-create implementation. The
      * caller then derives as it always has.
      */
-    override suspend fun createPasskeyWithPrf(
+    override suspend fun createPasskey(
         excludeCredentials: List<ByteArray>,
         salts: List<String>,
     ): CreatePasskeyOutput {
         try {
-            val registration = core.registerWithPrf(excludeCredentials, salts)
+            val registration = core.register(excludeCredentials, salts)
             val c = registration.credential
             return CreatePasskeyOutput(
                 PasskeyCredential(c.credentialId, c.userId, c.aaguid, c.backupEligible),
--- a/crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
+++ b/crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
@@ -475,6 +475,10 @@ public class CredentialManagerPrfCore(
             ?.optJSONObject("prf")
             ?.optJSONObject("results")
             ?: return null
+        // Never throw: the passkey exists by now, so a decode failure that
+        // escapes fails the whole registration, and the caller's natural
+        // recovery is to register again and strand this one. Null falls
+        // back to the assertion path, which derives from the same passkey.
         val first = results.optString("first").takeIf { it.isNotEmpty() } ?: return null
         val out = ArrayList<ByteArray>(2)
         out.add(decodeBase64UrlOrNull(first, "registration prf.first") ?: return null)
--- a/crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
+++ b/crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
@@ -461,13 +461,29 @@ public class CredentialManagerPrfCore(
     // ------------------------------------------------------------------
 
     /**
-     * Run one assertion ceremony for [salts] (1 or 2): build the WebAuthn
-     * request (cross-device hybrid suppressed unless
-     * [preferImmediatelyAvailableCredentials] is false), evaluate PRF, and
-     * return one 32-byte output per salt the authenticator evaluated plus
-     * the asserted credential ID. A dropped `results.second` yields a
-     * single-element list.
+     * PRF outputs carried on a registration response, or null when the
+     * authenticator reported support without evaluating (`prf.enabled`
+     * with no `results`), which is the pre-eval-at-create behavior.
      */
+    private fun readRegistrationPrfResults(responseJson: JSONObject): List<ByteArray>? {
+        // Never throw: the passkey exists by now, so a decode failure that
+        // escapes fails the whole registration, and the caller's natural
+        // recovery is to register again and strand this one. Null falls
+        // back to the assertion path, which derives from the same passkey.
+        val results = responseJson
+            .optJSONObject("clientExtensionResults")
+            ?.optJSONObject("prf")
+            ?.optJSONObject("results")
+            ?: return null
+        val first = results.optString("first").takeIf { it.isNotEmpty() } ?: return null
+        val out = ArrayList<ByteArray>(2)
+        out.add(decodeBase64UrlOrNull(first, "registration prf.first") ?: return null)
+        results.optString("second").takeIf { it.isNotEmpty() }?.let {
+            out.add(decodeBase64UrlOrNull(it, "registration prf.second") ?: return null)
+        }
+        return out
+    }
+
     /**
      * [assertPrf], re-asking until [retryUntilMs] while Credential Manager
      * still reports no credential. A passkey that exists but is not yet
@@ -499,6 +515,15 @@ public class CredentialManagerPrfCore(
             }
         }
     }
+
+    /**
+     * Run one assertion ceremony for [salts] (1 or 2): build the WebAuthn
+     * request (cross-device hybrid suppressed unless
+     * [preferImmediatelyAvailableCredentials] is false), evaluate PRF, and
+     * return one 32-byte output per salt the authenticator evaluated plus
+     * the asserted credential ID. A dropped `results.second` yields a
+     * single-element list.
+     */
 
     private suspend fun assertPrf(
         salts: List<String>,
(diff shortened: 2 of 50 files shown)


crates/breez-sdk/core/src/passkey/passkey_prf_provider.rs
```diff
@@ -95,6 +95,32 @@ pub trait PrfProvider: Send + Sync {
...
+    /// a create can outrun the platform's own credential bookkeeping: on
+    /// Android the new credential is briefly not resolvable, and the miss
+    /// is indistinguishable from a deleted passkey. Returning seeds here
+    /// removes that second ceremony, so the window cannot be hit.
+    ///
+    /// The default keeps the two-ceremony flow, so a provider that cannot
+    /// evaluate PRF at create needs no changes. Return `seeds: None` for
+    /// anything short of one output per salt (some authenticators drop
+    /// `prf.eval.second`): a partial result is not usable, and the caller
+    /// falls back to [`Self::derive_seeds`].
```
dangeross: This says when to return `None` but not what the values have to be. They have to equal what the same provider's `derive_seeds` returns for those salts, or register derives one wallet and every sign-in after it derives another, silently.

Worth spelling out here? The trait is foreign-implementable and the docs point custom providers (`YubiKey`, HSM, on-disk key material) straight at it.

```suggestion
    /// The default keeps the two-ceremony flow, so a provider that cannot
    /// evaluate PRF at create needs no changes. Return `seeds: None` for
    /// anything short of one output per salt (some authenticators drop
    /// `prf.eval.second`): a partial result is not usable, and the caller
    /// falls back to [`Self::derive_seeds`].
    ///
    /// Seeds returned here must equal what [`Self::derive_seeds`] would
    /// return for the same salts. The wallet is derived from them either
    /// way, so a mismatch means register and sign-in land on different
    /// wallets.
```

crates/breez-sdk/core/src/passkey/passkey_client.rs
```diff
@@ -695,6 +746,45 @@ mod tests {
...
+        let fallback_provider = Arc::new(MockProvider::new([7u8; 32]));
+        let fallback = test_client(fallback_provider.clone(), None)
+            .register(RegisterRequest {
+                label: Some("alice".to_string()),
+                ..Default::default()
+            })
+            .await
+            .unwrap();
+        assert!(!fallback_provider.derive_allow.lock().unwrap().is_empty());
+        assert_eq!(mnemonic_of(&with_prf.wallet), mnemonic_of(&fallback.wallet));
```
dangeross, line 808: `MockProvider` answers both paths from the same `output_for(salt)`, so the two mnemonics match by construction and this passes whatever a real authenticator does. It pins that the fast path derives from the same salts, not that create-time PRF equals assertion-time PRF.

Have we confirmed on a device that Credential Manager returns the same PRF output at create as at assertion? Register through the fast path then sign in through the assertion path is the check. If they ever differ, the user's funds are in a wallet nothing can reach again, and nothing here would catch it.

crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
```diff
@@ -657,10 +708,20 @@ public class CredentialManagerPrfCore(
...
-            // race the credential's PRF-readiness window (see grace tracker).
+            // Only usable as a complete set: a dropped `prf.eval.second`
+            // yields one output where the caller asked for two, and a
+            // partial derive is no derive at all.
+            val seeds = readRegistrationPrfResults(responseJson)
+                ?.takeIf { it.size == salts.size }
+            // Arm the post-create grace so a fallback derive doesn't race
+            // the credential's PRF-readiness window (see grace tracker).
+            // Unnecessary when seeds came back inline, but harmless: the
+            // window is consumed without waiting when nothing asserts.
```
dangeross: nit: nothing consumes it when nothing asserts. `consumeDeadline` only runs from `deriveSeeds`, so on the fast path the window stays armed until some later derive picks it up. Still harmless, just not for the reason given.

```suggestion
            // Unnecessary when seeds came back inline, but harmless: only
            // a derive reads the window, and the next one clears it.
```

crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
```diff
@@ -627,7 +660,25 @@ public class CredentialManagerPrfCore(
                     put("userVerification", "required")
                 })
                 put("extensions", JSONObject().apply {
-                    put("prf", JSONObject())
+                    // Ask the create ceremony to evaluate PRF as well as
+                    // report support. An authenticator that answers turns
+                    // registration into a single ceremony, so no assertion
+                    // has to race the credential becoming resolvable. One
+                    // that ignores it returns `enabled` only, and the
+                    // caller falls back to `deriveSeeds`.
```
dangeross, line 667: This assumes a provider that can't evaluate at create ignores the input. Does that hold outside GPM (1Password, Bitwarden, etc)? One that rejects an extension input it doesn't understand fails the registration outright, which is a worse trade than the window we're removing.

Might be a non-issue, just want to be sure it's been tried.

crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/PasskeyProvider.kt
```diff
@@ -142,10 +143,31 @@ public class PasskeyProvider(
...
+     * evaluates PRF at create returns the seeds here and the caller needs
+     * no assertion. Credential Manager reports a credential it has not
+     * finished indexing as absent, so removing that assertion removes the
+     * only step that can trip over it.
+     *
+     * `seeds` stays null on authenticators that report PRF support without
+     * evaluating, which is every pre-eval-at-create implementation. The
+     * caller then derives as it always has.
+     */
+    override suspend fun createPasskeyWithPrf(
```
dangeross: The RN provider needs this override too. `passkey-prf-provider.ts` hands the TS object to `SdkPasskeyClient` as a foreign `PrfProvider`, and foreign impls don't inherit the Rust default, so `createPasskeyWithPrf` has to exist there. Only `createPasskey` does.

Nothing flags it either: the `as unknown as PrfProvider` cast keeps tsc quiet, and `CLI / react-native` is green. `register` calls into it every time.

roeierez: Same issue exists for flutter. Need to add that implementation.

roeierez: Also can we add also the new implementation to the web passkey provider?

packages/flutter/android/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
```diff
@@ -466,6 +466,24 @@ public class CredentialManagerPrfCore(
...
+     */
+    private fun readRegistrationPrfResults(responseJson: JSONObject): List<ByteArray>? {
+        val results = responseJson
+            .optJSONObject("clientExtensionResults")
+            ?.optJSONObject("prf")
+            ?.optJSONObject("results")
+            ?: return null
+        val first = results.optString("first").takeIf { it.isNotEmpty() } ?: return null
+        val out = ArrayList<ByteArray>(2)
+        out.add(decodeBase64Url(first))
```
roeierez: Should we use decodeBase64UrlOrNull like in the assertion path? The passkey already created and this will throw an exception if non base64 is returned. Error here may lead to create another passkey later by the user.

roeierez (changes requested):
The approach looks robust and better than the previous two steps flow.

packages/flutter/android/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
erdemyerebasmaz: There will be a follow-up expanding this feature for Web, Flutter and RN.

Null-handling is addresed in 70b7125

crates/breez-sdk/core/src/passkey/passkey_prf_provider.rs
erdemyerebasmaz: Addressed in 01847c843.

crates/breez-sdk/core/src/passkey/passkey_client.rs
erdemyerebasmaz: Confirmed on device, both platforms: create, log out, sign back in resolves the same recovery phrase. The test still only pins the salts, as you say.

crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
erdemyerebasmaz: Dropped the arming on the fast path rather than rewording the comment. Addressed in 01847c843.

crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
erdemyerebasmaz: Verified on a device whose provider evaluated it, and WebAuthn specifies unrecognised extensions as ignorable. 1Password and Bitwarden as Android providers remain untested.

crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/PasskeyProvider.kt
erdemyerebasmaz: RN in 83e0d99e8, Flutter in 83e0d99e8, web in 58ec0fe28.

packages/flutter/android/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
erdemyerebasmaz: Addressed in a65cf7aa6.

crates/breez-sdk/bindings/langs/swift/Sources/BreezSdkSpark/PasskeyAssertionCore.swift
```diff
@@ -600,7 +628,7 @@ public final class PasskeyAssertionCore {
         // Arm the post-create grace so the immediate derive doesn't race
         // the credential's PRF-readiness window (see grace tracker).
         await graceTracker.arm(after: postCreateGraceTotal)
-        return credential
+        return registration
```
dangeross: Should we only arm the grace tracker if no seeds are resolved, like in Android? Otherwise the next call to `deriveSeeds` will wait it out for no reason.

```suggestion
        // Arm the post-create grace only when a derive still has to run:
        // the window exists for that assertion. Arming it after an inline
        // derive would leave it set for whatever ceremony came next.
        if registration.seeds == nil {
            await graceTracker.arm(after: postCreateGraceTotal)
        }
        return registration
```

crates/breez-sdk/wasm/js/passkey-prf-provider/index.js
```diff
@@ -507,7 +514,7 @@ export class PasskeyProvider {
      * @returns {Promise<{ credentialId: Uint8Array, userId: Uint8Array, aaguid: Uint8Array | null, backupEligible: boolean | null }>}
      * @private
      */
```
dangeross: This is the old flat shape, and `salts` has no `@param`.

```suggestion
     * @param {string[]} [salts=[]] - Salts to evaluate during the create
     *   ceremony; omit to register without deriving.
     * @returns {Promise<{ credential: { credentialId: Uint8Array, userId: Uint8Array, aaguid: Uint8Array | null, backupEligible: boolean | null }, seeds: Uint8Array[] | null }>}
     * @private
     */
```

packages/react-native/android/src/main/kotlin/com/breeztech/breezsdkspark/BreezSdkSparkPasskeyModule.kt
```diff
@@ -214,16 +215,22 @@ class BreezSdkSparkPasskeyModule(
...
                     userName = userName,
                     userDisplayName = userDisplayName,
                     activityProvider = { activity },
                     graceTracker = graceTracker,
-                ).register(excludeIds)
+                ).register(excludeIds, registerSalts)
+                val credential = registration.credential
                 // The core's `register` arms the shared grace tracker so the
                 // next `deriveSeeds` call holds out the credential's
                 // PRF-readiness window without the wrapper having to.
```
dangeross: nit: the Flutter plugin's copy of this got the extra clause, this one didn't.

```suggestion
                // The core's `register` arms the shared grace tracker so the
                // next `deriveSeeds` call holds out the credential's
                // PRF-readiness window without the wrapper having to, and
                // skips arming it when seeds came back inline.
```

packages/react-native/ios/BreezSdkSparkPasskey.swift
```diff
@@ -90,6 +90,7 @@ class BreezSdkSparkPasskey: NSObject {
         userName: String,
         userDisplayName: String,
         excludeCredentials: [String],
+        registerSalts: [String],
```
dangeross, line 93: `BreezSdkSparkPasskey.m` hasn't been updated, so the selector RN registers still ends `excludeCredentials:resolve:reject:` and no longer matches this method.

Needs `registerSalts:(NSArray *)registerSalts` between `excludeCredentials` and `resolve`.

dangeross (commented):
Looking good

crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
```diff
@@ -451,6 +451,30 @@ public class CredentialManagerPrfCore(
      * the asserted credential ID. A dropped `results.second` yields a
      * single-element list.
      */
+    /**
+     * PRF outputs carried on a registration response, or null when the
+     * authenticator reported support without evaluating (`prf.enabled`
+     * with no `results`), which is the pre-eval-at-create behavior.
+     */
+    private fun readRegistrationPrfResults(responseJson: JSONObject): List<ByteArray>? {
```
danielgranhao, line 468: `readRegistrationPrfResults` was places in between `assertPrf` from its KDoc. Same in Flutter and RN copies

crates/breez-sdk/bindings/langs/swift/Sources/BreezSdkSpark/PasskeyAssertionCore.swift
erdemyerebasmaz: Yes. Reached the same conclusion from a docs audit just before this landed. Addressed in 7366fb6c3.

crates/breez-sdk/wasm/js/passkey-prf-provider/index.js
erdemyerebasmaz: Addressed in 7366fb6c3.

packages/react-native/android/src/main/kotlin/com/breeztech/breezsdkspark/BreezSdkSparkPasskeyModule.kt
erdemyerebasmaz: Addressed in 7366fb6c3.

packages/react-native/ios/BreezSdkSparkPasskey.swift
erdemyerebasmaz: Good catch, that one is a runtime break neither the Swift nor the TypeScript compile would surface. Addressed in 7366fb6c3.

crates/breez-sdk/wasm/src/models/passkey_prf_provider.rs
danielgranhao: This needs to be updated, right?

crates/breez-sdk/core/src/passkey/passkey_client.rs
```diff
@@ -681,6 +722,47 @@ mod tests {
         }
     }
 
+    #[macros::async_test_all]
+    async fn register_skips_the_assertion_when_create_evaluates_prf() {
```
danielgranhao, line 781: Nice test! Wdyt about adding one for partial seeds too? The mock now never leads the guard to run with a mismatch of seeds len to salts len. A `with_partial_prf_at_create()` returning 1 seed for 2 salts, asserting we still fall back and still get the same mnemonic, would cover it.

crates/breez-sdk/bindings/langs/swift/Sources/BreezSdkSpark/PasskeyAssertionCore.swift
```diff
@@ -545,16 +559,24 @@ public final class PasskeyAssertionCore {
...
     }
 
-    /// Register a new passkey with PRF support (one platform prompt, no
-    /// seed derivation). `excludeCredentials` lists already-registered IDs
-    /// the platform must refuse as duplicates. The post-create grace
-    /// tracker is armed on success. `userId` is never host-supplied: the
+    /// Register a new passkey with PRF support. `excludeCredentials` lists
+    /// already-registered IDs the platform must refuse as duplicates.
+    /// Evaluating PRF for `salts` in the same ceremony returns the seeds
+    /// inline; the post-create grace tracker is armed only when it did
```
dangeross: nit: feels like these can be merged
> Evaluating PRF for `salts` in the same ceremony returns the seeds inline

> Asking the create ceremony to evaluate PRF for `salts` makes registration a single ceremony, so no assertion follows it

crates/breez-sdk/core/src/passkey/passkey_client.rs
danielgranhao: Stale doc?

dangeross (commented):
Minor nit (and merge conflicts), otherwise LGTM

danielgranhao (commented):
Looks good!

crates/breez-sdk/bindings/langs/shared/android-passkey/src/main/kotlin/technology/breez/spark/passkey/core/CredentialManagerPrfCore.kt
erdemyerebasmaz: Addressed in 2a50e2a36.

crates/breez-sdk/wasm/src/models/passkey_prf_provider.rs
erdemyerebasmaz: Addressed in 2a50e2a36.

crates/breez-sdk/core/src/passkey/passkey_client.rs
erdemyerebasmaz: Good call, the old mock could never trip that guard. Added with_partial_prf_at_create, one output for two salts, asserting the fallback runs and reaches the same mnemonic. Addressed in 2a50e2a36.

crates/breez-sdk/bindings/langs/swift/Sources/BreezSdkSpark/PasskeyAssertionCore.swift
erdemyerebasmaz: Addressed in 2a50e2a36.

crates/breez-sdk/core/src/passkey/passkey_client.rs
erdemyerebasmaz: Yes, and it also claimed the label is always published when the write is spawned best-effort. Addressed in 2a50e2a36.

dangeross (approved):
Good from my side

danielgranhao (approved):
LGTM

roeierez (approved):
LGTM

PR #1026, opened 2026-07-29
feat(ios): derive wallet during passkey creation

erdemyerebasmaz:
iOS can now request the PRF value during passkey creation instead of requiring a follow-up assertion.

This reduces one auth prompt during account creation where supported. If PRF values cannot be returned during creation, the existing assertion flow remains unchanged.

#### Relationship to #1024:
This builds on the `create_passkey_with_prf` interface added in #1024. Android and iOS now use the same PRF-at-create flow.


erdemyerebasmaz:
Folded into #1024, which is now the single platform-agnostic PR for deriving the wallet during passkey creation: Android, iOS, web, React Native and Flutter.

PR #1027, opened 2026-07-30
docs: correct two passkey comments that no longer match the code

erdemyerebasmaz:
Two comments left behind by #1022. Comments only, no behaviour change.

- `PasskeyError::kind()` says non-PRF variants map to `Internal`. `CreatedButNotDerived` delegates to the `PrfProviderError` it wraps, so it does not, and a caller branching on `kind()` alone cannot tell a post-creation failure from the same failure before the credential existed. The doc now says so and points at matching the variant.
- The Kotlin grace tracker says it mirrors iOS. iOS still waits the window out rather than re-asking, which is the fixed-sleep design the same comment rejects three lines above.

Found by an audit of the passkey docs against the code. Two further findings from that audit are already fixed in #1024, which touches the same lines, so they are left out here to avoid a conflict.

### DIFF
--- a/crates/breez-sdk/core/src/passkey/error.rs
+++ b/crates/breez-sdk/core/src/passkey/error.rs
@@ -138,9 +138,12 @@ pub enum PasskeyError {
 }
 
 impl PasskeyError {
-    /// Coarse classification of the failure. Non-PRF variants map to
-    /// `Internal`: they stem from SDK / network state, not the
-    /// authenticator, so surface a generic "try again later".
+    /// Coarse classification of the failure. `CreatedButNotDerived`
+    /// reports the classification of the `PrfProviderError` it wraps, so
+    /// it is indistinguishable here from the same failure before the
+    /// credential existed: match the variant itself to tell them apart.
+    /// The remaining variants map to `Internal`, since they stem from SDK
+    /// or network state rather than the authenticator.
     #[must_use]
     pub fn kind(&self) -> ErrorKind {
         match self {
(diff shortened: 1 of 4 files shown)


crates/breez-sdk/core/src/passkey/error.rs
```diff
@@ -138,9 +138,12 @@ pub enum PasskeyError {
...
 impl PasskeyError {
-    /// Coarse classification of the failure. Non-PRF variants map to
-    /// `Internal`: they stem from SDK / network state, not the
-    /// authenticator, so surface a generic "try again later".
+    /// Coarse classification of the failure. `CreatedButNotDerived`
+    /// reports the classification of the `PrfProviderError` it wraps, so
+    /// it is indistinguishable here from the same failure before the
+    /// credential existed: match the variant itself to tell them apart.
+    /// The remaining variants map to `Internal`, since they stem from SDK
+    /// or network state rather than the authenticator.
```
dangeross: nit: Bit clearer? as the Prf variant is not directly mentioned
```suggestion
    /// Coarse classification of the failure. `CreatedButNotDerived`
    /// reports the classification of the `PrfProviderError` it wraps, so
    /// it is indistinguishable here from the same failure before the
    /// credential existed: match the variant itself to tell them apart.
    /// The remaining non-Prf variants map to `Internal`, since they stem
    /// from SDK or network state rather than the authenticator.
```

dangeross (approved):
LGTM, with a nit

roeierez (approved):

crates/breez-sdk/core/src/passkey/error.rs
erdemyerebasmaz: Addressed in 106b3af45.

PR #1028, opened 2026-07-30
Exit any leaf that is ours, in any status

JssDWt:
Exit any leaf that is ours, whatever status the operators report it in, and
keep a leaf's chain until a spend is proven rather than deleting it on absence.

## API

`TreeStore` (Rust trait, all 5 backends, plus the JS `TreeStore` interface):

- `get_deleted_leaves()` / `getDeletedLeaves()` - the leaves kept only for their exit chain.
- `remove_leaves(leaf_ids)` / `removeLeaves(leafIds)` - drop a leaf and its chain for good.

`TreeService`:

- `list_leaves_kept_for_exit()` - the same leaves, for exit selection.
- `purge_proven_spent_leaves()` - retire the ones every operator proves have left us.

Storage gains an `is_deleted` column (SQLite, Postgres, MySQL, and the three
SQL-backed JS stores; IndexedDB needs no migration).

## Design

**Only exit leaves that are ours, but all of them.** `EXIT_CHAIN_STATUSES` is
gone. The chain walk is purely structural, and leaf selection consults no status
at all, so a leaf under a `RenewLocked` or `Investigation` ancestor is exitable
and an already-exited leaf stays selectable for a re-run to resolve as swept.

**Absence is not a spend.** A leaf no operator reports is marked rather than
deleted: it leaves the balance and payment selection but keeps its chain and
stays exitable. The operators cannot even report some of the statuses a leaf of
ours can reach, and `Lost` is exactly where they stop co-signing refunds.

**Every operator has an equal say.** The refresh queries all of them, not just
the coordinator. Adding a leaf takes one; dropping one takes unanimity.

**A deletion needs evidence, not testimony.** A leaf is only removed once every
operator returns a replacement refund that spends our stored node output, is
owned by somebody else, sits at a lower timelock, and carries a signature
verifying under our own verifying key. A leaf is 2-of-2, so that signature is
one we took part in producing and no operator set can forge it. Everything
weaker keeps the leaf: an unreachable operator holds the whole round.

The purge runs hourly in the background and before an exit is quoted.

## Note

The refresh cannot ask for `TREE_NODE_STATUS_PARENT_EXITED`: the operator
rejects the entire query with `unknown tree node status`, so one unrecognised
status costs every leaf rather than the ones in that status. Worth raising
separately; a leaf under an exited parent is meanwhile kept for its chain.

### DIFF
--- a/crates/spark/src/tree/service.rs
+++ b/crates/spark/src/tree/service.rs
@@ -505,28 +505,62 @@ impl TreeService for SynchronousTreeService {
             return Ok(0);
         }
         let ids: Vec<TreeNodeId> = deleted.iter().map(|leaf| leaf.id.clone()).collect();
-        // By id rather than by owner. An owner query is status-filtered, and the
-        // operators drop Investigation, Lost and Reimbursed from it whatever we
-        // ask for, so it cannot tell a spent leaf from one it will not talk
-        // about. A lookup by id answers for a node in any status.
-        let theirs: HashMap<TreeNodeId, TreeNode> = self
-            .fetch_nodes(&ids, false)
-            .await?
-            .into_iter()
-            .map(|node| (node.id.clone(), node))
+        let source = Source::NodeIds(TreeNodeIds {
+            node_ids: ids.iter().map(ToString::to_string).collect(),
+        });
+
+        // Ask every operator, by id. By id because an owner query is
+        // status-filtered, and the operators drop Investigation, Lost and
+        // Reimbursed from it whatever statuses we ask for, so it cannot tell a
+        // spent leaf from one it will not talk about; a lookup by id answers for
+        // a node in any status. Every operator because this ends in a deletion,
+        // and no single one of them, coordinator included, gets to decide that on
+        // its own.
+        let operators: Vec<_> = self
+            .operator_pool
+            .get_all_operators()
+            .map(|op| (op.id, op.client.clone()))
             .collect();
-        let proven: Vec<TreeNodeId> = deleted
-            .iter()
-            .filter(|leaf| {
-                theirs
+        let operator_count = operators.len();
+        let responses = join_all(operators.iter().map(|(id, client)| {
+            let source = source.clone();
+            async move {
+                (
+                    *id,
+                    self.query_nodes(client, false, Some(source), vec![]).await,
+                )
+            }
+        }))
+        .await;
+
+        // An operator we could not reach has not vouched for anything, so its
+        // silence holds the leaves rather than counting towards a deletion.
+        let mut agreeing: HashMap<TreeNodeId, usize> = HashMap::new();
+        for (id, res) in responses {
+            let Ok(nodes) = res else {
+                debug!("purge: operator {id} unreachable, keeping every leaf this round");
+                return Ok(0);
+            };
+            let theirs: HashMap<TreeNodeId, TreeNode> =
+                nodes.into_iter().map(|n| (n.id.clone(), n)).collect();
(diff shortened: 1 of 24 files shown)


crates/spark/src/tree/service.rs
```diff
@@ -372,7 +390,7 @@ impl TreeService for SynchronousTreeService {
         } else {
             HashMap::new()
         };
-        let unverified_leaves: Vec<&TreeNode> = available_leaves
+        let unverified_leaves: Vec<&TreeNode> = reported_leaves
```
dangeross, line 402: It seems that now we widen what's verified there's a refresh cost to an external signer, once a wallet has exited leaves. `verified_leaf_keys_from_leaves` excludes `not_available` leaves, so they're never in `already_verified`, and `Exited` is terminal so they don't age out.

Turnkey memoizes leaf pubkeys itself so there it's just cold start. An integrator's `ExternalSparkSigner` might have no cache above it.

I'm trying to look at "is this optimal", but maybe if we're exiting we don't need to be optimal any more?

crates/spark/src/tree/service.rs
dangeross: Change to "every reported leaf"?

crates/spark/src/tree/service.rs
```diff
@@ -267,97 +270,112 @@ impl TreeService for SynchronousTreeService {
...
-                        );
-                        missing_operator_leaves_map.insert(leaf.id.clone(), leaf.clone());
-                    }
-                }
+        // A leaf every operator reported identically is clean. One that some of them
+        // dropped, or described differently, is still ours and still counts towards
+        // the balance, but it is held back from payments until they agree again. A
+        // leaf no operator reports is simply absent from here, and that absence is
+        // what the store reads as a deletion: it now takes every operator to drop a
+        // leaf, where before the coordinator alone decided.
```
dangeross, line 351: Should a leaf only one operator still reports count towards the balance? The coordinator dropping it used to be enough to remove it, so `is_missing_from_operators` now covers both "everyone has it but describes it differently" and "only one still has it, the coordinator dropped it", and the balance counts both the same.

It's non-selectable either way, so we'd show a balance that can't back a payment until that operator catches up.

dangeross (commented):
Posting review progress...

crates/spark/src/tree/service.rs
JssDWt: Do you mean that on a turnkey backend you have to get_public_key_for_leaf for each of those leaves, and that might do CREATE_WALLET_ACCOUNTS? I think this is fair tbh.

crates/spark/src/tree/service.rs
JssDWt: That's a good question. I think it's quite the same scenario as when the coordinator reported a leaf, but one of the other operators didn't? In that case we decided it should count as balance, so it should count as balance vice versa as well?

crates/spark/src/tree/service.rs
dangeross: I think we can drop this, I was thinking of the query cost going through the RPS/pacer per cold start but it's really minor

crates/spark/src/tree/service.rs
dangeross: I have a feeling a finalized send followed shortly after by a refresh could bounce the balance back up? The leaf goes locally at finalize, so if any operator is still reporting it while the coordinator has moved on, we'd union it back in and count it. Am I overthinking it?

JssDWt:
Replaced by #1098

PR #1029, opened 2026-07-30
Back up and restore unilateral exit state

JssDWt:
Backs up the data a unilateral exit is built from, so it survives losing the device's storage while the operators are unreachable.

## API

- `SdkEvent::UnilateralExitStateChanged`: the exit data changed, so a previously exported copy is out of date.
- `exportUnilateralExitState()` returns `exitState`, the whole exit state as one opaque value.
- `importUnilateralExitState({ exitState })` returns `importedLeaves`, `skippedLeaves` and `skippedChains`.

Exported data is JSON. It has a version.


crates/spark-wallet/src/wallet.rs
```diff
@@ -1502,6 +1524,88 @@ impl SparkWallet {
...
+                .collect();
+
+            let to_restore: Vec<LeafPedigree> = owned
+                .into_iter()
+                .map(|mut pedigree| {
+                    let stored_ancestors = stored.get(&pedigree.leaf.id);
+                    let stored_len = stored_ancestors.map_or(0, Vec::len);
+                    if reaches_root(&pedigree) && pedigree.ancestors.len() >= stored_len {
+                        return pedigree;
+                    }
```
danielgranhao: This protects the ancestors but don't we also need some protection to prevent importing an older leaf that has had its timelock renewed?

crates/spark-wallet/src/wallet.rs
```diff
@@ -174,6 +177,15 @@ fn is_backoff_retryable_error(error: &OperatorRpcError) -> bool {
...
 }
 
+/// Whether a chain reaches a root, which is what makes the leaf exitable
+/// without the operators. A leaf that is itself a root needs no ancestors.
+fn reaches_root(pedigree: &LeafPedigree) -> bool {
+    match pedigree.ancestors.last() {
+        Some(topmost) => topmost.parent_node_id.is_none(),
+        None => pedigree.leaf.parent_node_id.is_none(),
+    }
+}
```
danielgranhao: Should we also validate that the chain is actually linked? Edge case I know but a corrupted backup with a parentless node anywhere would pass and replace a good stored chain.

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -2674,3 +2674,35 @@ pub struct UnilateralExitResponse {
...
+}
+
+/// Result of `import_unilateral_exit_state`.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct ImportUnilateralExitStateResponse {
+    /// Leaves merged into the wallet's exit state.
+    pub imported_leaves: u32,
+    /// Leaves left out because the exit state does not record this wallet as
+    /// their owner.
```
danielgranhao, line 2704: nit: IIUC, this also counts leaves where the wallet had no stored copy. 
```suggestion
    /// Leaves left out because the incoming chain wasn't usable
```

docs/breez-sdk/src/guide/unilateral_exit.md
```diff
@@ -145,6 +145,24 @@ To re-broadcast the same leaves at a higher fee, quote again with the same {{#en
...
 
+## Back up the exit data
+
+The transactions an exit is built from are held in the SDK's local storage. While the operators are reachable they can be fetched again, so a wallet restored from its seed rebuilds them on its own. When that storage is gone and the operators are unreachable, they cannot be recovered from anywhere, and the leaves they cover cannot be exited.
+
+{{#name export_unilateral_exit_state}} returns that data as a single opaque value, covering every leaf the wallet holds together with the transactions that spend it. It reflects what is present when it is called: a leaf whose data has not been collected yet is exported without it. The value grows with the number of leaves and can reach several megabytes.
```
danielgranhao, line 152: Worth explaining that the data exposes the wallet's shape? A recommendation to encrypt it could make sense

danielgranhao (commented):
Looks good

roeierez:
@danielgranhao @dangeross I rebased and simplified a bit the event triggers.
We trigger event of ExitStateChanged whenever leaf ancestors are fetched and added to the store.
This can happen on two flows:
1. Refresh timelock within the tree service
2. exit_chain_resolver downloaded and added ancestors to the store.

crates/spark/src/tree/service.rs
```diff
@@ -803,26 +896,22 @@ impl SynchronousTreeService {
...
-        if let Err(e) = self.state.store_ancestors(&renewed).await {
+        if !renewed.is_empty()
+            && let Err(e) = self.state.store_ancestors(&renewed).await
+        {
             warn!("Failed to store renewed exit chains: {e:?}");
         }
+        // Announced even with nothing written: the ordinary renewal leaves a leaf
+        // where it is and re-signs it there, so the leaf a backup holds is stale
+        // though no chain moved.
+        self.signal_exit_state_changed();
```
dangeross, line 919: If a renewal is due and fails, is `UnilateralExitStateChanged` still emitted? `any_renewal_due` is read off the leaves going in, and `check_renew_nodes` swallows the failure and returns the leaf unchanged, so it's still due next pass and we'd announce again on every `refresh_leaves`. Am I reading that right?

dangeross (commented):
Looking good

crates/spark-wallet/src/wallet.rs
```diff
@@ -1503,6 +1591,141 @@ impl SparkWallet {
...
+                            chains_only.push(incoming);
+                        } else {
+                            to_restore.push(incoming);
+                        }
+                    }
+                    None => {
+                        if parts == ImportableParts::Nothing {
+                            skipped_chains += 1;
+                            continue;
+                        }
```
danielgranhao, line 1696: Here we drop the leaf, not just the chain, but we only increment `skipped_chains`, which implies a leaf was kept, only the chain skipped. Should something tell the caller the leaf didn't make it?

danielgranhao (commented):
Looks good

crates/spark/src/tree/service.rs
roeierez: Good catch. I changed that to compare the leaf after the renewal (attempt) and only if differs trigger the event.

dangeross (approved):
Looks good from my side

danielgranhao (approved):

PR #1034, opened 2026-08-06
Chunk leaf upserts in the SQL tree stores

danielgranhao:
`set_leaves` writes the entire leaf set in a single `INSERT`. At ~100k leaves:

- **Postgres**: one statement materializing 100k JSON blobs, holding a pooled connection and the advisory lock throughout. A partner deployment saw pool exhaustion, then `terminating connection because of crash of another server process` and cluster recovery.
- **MySQL**: 6 placeholders per leaf against a 65535 cap, so past ~10,922 leaves the write fails outright with `ERROR 1390: Prepared statement contains too many placeholders`.

Chunk at 1000 in all four backends that build such a statement (Rust and JS, both engines). Chunks run inside the caller's transaction, so the set still lands atomically.

Adds `test_set_leaves_across_chunk_boundary` to the shared suite, wired into all five harnesses. Sized at 12,345 leaves so it exceeds the MySQL cap and fails without the fix.

### DIFF
--- a/crates/breez-sdk/wasm/js/mysql-tree-store/index.cjs
+++ b/crates/breez-sdk/wasm/js/mysql-tree-store/index.cjs
@@ -44,6 +44,16 @@ const RESERVATION_TIMEOUT_SECS = 300;
 const SPENT_MARKER_CLEANUP_THRESHOLD_MS = 5 * 60 * 1000;
 
 /**
+ * Leaves per INSERT when upserting a refreshed leaf set.
+ *
+ * Required for correctness, not just memory: this statement binds 6
+ * placeholders per leaf, and the MySQL protocol caps a prepared statement at
+ * 65535. A wallet holding six figures of leaves would otherwise exceed both
+ * that cap and max_allowed_packet.
+ */
+const LEAF_UPSERT_CHUNK_SIZE = 1000;
+
+/**
  * Slim projection: only (id, value) for leaves the selection might use.
  * Includes all leaves with value <= the max target (covers exact-match + the
  * small-leaf accumulators for the minimum-amount path) plus the single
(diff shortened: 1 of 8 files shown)


crates/spark/src/tree/tests.rs
```diff
@@ -221,6 +221,40 @@ pub async fn test_set_leaves(store: &dyn TreeStore) {
...
+    assert_eq!(
+        stored.iter().map(|l| l.value).sum::<u64>(),
+        expected_total,
+        "every leaf must round-trip with its own value"
+    );
+    assert_eq!(store.get_available_balance().await.unwrap(), expected_total);
+
+    // Re-running must upsert in place rather than duplicate across chunks.
+    let refresh_start = future_refresh_start(store).await;
+    store.set_leaves(&leaves, &[], refresh_start).await.unwrap();
```
roeierez: The re-run never reaches the upsert path. future_refresh_start is 10s ahead, so the DELETE clears all 12,345 rows and this insert goes into an empty table. ON CONFLICT / ON DUPLICATE KEY never fires. I checked on a real Postgres: re-running with a 10-leaf subset leaves exactly 10 rows. Use past_refresh_start here so the rows stay and every chunk collides.

crates/spark-postgres/src/tree_store.rs
```diff
@@ -1189,41 +1197,52 @@ impl PostgresTreeStore {
...
+                missing_flags.push(is_missing_from_operators);
+                data_values.push(Self::serialize_node(leaf)?);
+            }
 
-        for leaf in filtered {
-            ids.push(leaf.id.to_string());
-            statuses.push(leaf.status.to_string());
-            missing_flags.push(is_missing_from_operators);
-            data_values.push(Self::serialize_node(leaf)?);
+            tx.execute(
```
roeierez, line 1242: Passing the SQL as a &str makes tokio-postgres prepare it again on every chunk (there is no cache). 100 chunks becomes 200 round trips instead of 101, all while holding the advisory lock the payment path waits on. Hoist let stmt = tx.prepare(...) above the loop and pass &stmt.

roeierez (changes requested):
Two comments, otherwise looks good.

crates/spark/src/tree/tests.rs
danielgranhao: Done

crates/spark-postgres/src/tree_store.rs
danielgranhao: Done

crates/breez-sdk/wasm/js/mysql-tree-store/index.cjs
```diff
@@ -43,6 +43,16 @@ const WRITE_LOCK_TIMEOUT_SECS = 30;
 const RESERVATION_TIMEOUT_SECS = 300;
 const SPENT_MARKER_CLEANUP_THRESHOLD_MS = 5 * 60 * 1000;
 
+/**
+ * Leaves per INSERT when upserting a refreshed leaf set.
+ *
+ * Required for correctness, not just memory: this statement binds 6
+ * placeholders per leaf, and the MySQL protocol caps a prepared statement at
+ * 65535. A wallet holding six figures of leaves would otherwise exceed both
+ * that cap and max_allowed_packet.
```
dangeross, line 52: I don't think this applies to `conn.query()`, only `conn.execute()` uses prepared statements. But its still good to chunk it.

roeierez (approved):
LGTM

PR #1035, opened 2026-08-06
Retry operator auth failures and add session expiry margin

danielgranhao:
A session handshake failure inside `build_interceptor` escapes `call_with_auth_retry` through `?`, bypassing its retry arm, so one failed re-authentication fails the whole calling operation (e.g. an entire leaf refresh).

- Retry a failed interceptor build once with a forced session refresh
- Treat sessions within 30s of expiry as invalid, so tokens are refreshed up front instead of expiring mid-call
- Add unit tests for both


roeierez (approved):
LGTM

PR #1038, opened 2026-08-07
Fix Expo config plugin resolution on Expo SDK 56

### LINKED ISSUE #1037: Expo config plugin fails to load on Expo SDK 56: `Cannot find module '@expo/config-plugins'`
**Version:** `@breeztech/breez-sdk-spark-react-native@0.19.2`, Expo SDK 56 (RN 0.85.3)

`plugin/build/index.js` does a bare `require("@expo/config-plugins")`, but the package is only listed in `devDependencies`. It resolved on Expo SDK 55 because `@expo/config-plugins` was hoisted to the root `node_modules`. SDK 56 nests it under `expo/`, `@expo/cli/` and `expo-splash-screen/`, so the require fails.

Breaks `expo config`, `expo prebuild` and `expo-doctor` — the project cannot be built at all.

```
PluginError: Cannot find module '@expo/config-plugins'
Require stack:
- node_modules/@breeztech/breez-sdk-spark-react-native/plugin/build/index.js
- node_modules/@breeztech/breez-sdk-spark-react-native/app.plugin.js
```

**Fix:** import from `expo/config-plugins` (the sub-export of `expo`) instead of `@expo/config-plugins`. This is what `expo-doctor` recommends:

> The package "@expo/config-plugins" should not be installed directly in your project. You should instead use "expo/config-plugins" which is a sub-export of the expo package. If you installed "@expo/config-plugins" to fulfill a peer dependency for a config plugin, the plugin's maintainer should switch to the "expo/config-plugins" import.

**Workaround:** install `@expo/config-plugins` matching the SDK (`~56.0.14`) at the project root.

dangeross:
The compiled config plugin did a bare `require('@expo/config-plugins')`, a package we do not depend on. It only ever resolved because Expo SDK <= 55 hoisted it to the app's root `node_modules`. SDK 56 nests it under `expo/`, `@expo/cli/` and `expo-splash-screen/`, so the require throws and `expo config`, `expo prebuild` and `expo-doctor` all fail, leaving the project unbuildable.

The plugin now imports from `expo/config-plugins`, the sub-export of `expo`, which resolves regardless of hoisting. This is what `expo-doctor` recommends.

Fixes #1037

### Changes
- `plugin/src/*.ts`: four import specifiers switched to `expo/config-plugins`.
- `package.json`: `@expo/config-plugins` devDependency replaced with `expo`. It only supplies build-time types, so no runtime or peer dependency is added.
- `ci.yaml`: new `rn-expo-plugin` job. The plugin was previously compiled only at publish time (`plugin/src` is excluded from both root tsconfigs), which is how this shipped. The job builds it and fails on a bare `@expo/config-plugins` require.

### Compatibility
`expo/config-plugins` is an unchanged re-export shim in every release from SDK 47 through 57, and no `expo` version declares an `exports` map that could block the subpath. Verified against throwaway apps on SDK 52 to 56, in both hoisted and nested `node_modules` layouts. `expo config --type introspect` confirms the gradle property and iOS entitlement mods still apply. Reverting the specifier reproduces the reported error, with the require stack pointing at `app.plugin.js`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -70,6 +70,37 @@ jobs:
         working-directory: packages/react-native
         run: node ./scripts/post-ubrn.js --check
 
+  rn-expo-plugin:
+    name: RN Expo plugin
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: packages/react-native
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: packages/react-native/.nvmrc
+
+      # Skips the postinstall hook, which downloads release artifacts keyed off
+      # a `checksums` field that only exists in the published package.json.
+      - name: Install dependencies
+        run: yarn --mode=skip-build
+
+      - name: Build the Expo config plugin
+        run: yarn plugin:build
+
+      # A bare `@expo/config-plugins` require fails from Expo SDK 56 onward,
+      # where the package is nested under `expo/` instead of hoisted.
+      - name: Verify config-plugins is resolved through expo
+        run: |
+          if grep -rnE "(require\(|from )['\"]@expo/config-plugins['\"]" plugin/build; then
+            echo "::error::plugin/build must resolve config-plugins via 'expo/config-plugins'"
+            exit 1
+          fi
+
   itest:
     name: Integration test (${{ matrix.backend.name }})
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -96,7 +96,7 @@ jobs:
       # where the package is nested under `expo/` instead of hoisted.
       - name: Verify config-plugins is resolved through expo
         run: |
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yaml
```diff
@@ -70,6 +70,37 @@ jobs:
...
+        run: yarn --mode=skip-build
+
+      - name: Build the Expo config plugin
+        run: yarn plugin:build
+
+      # A bare `@expo/config-plugins` require fails from Expo SDK 56 onward,
+      # where the package is nested under `expo/` instead of hoisted.
+      - name: Verify config-plugins is resolved through expo
+        run: |
+          if grep -rnE "(require\(|from )['\"]@expo/config-plugins['\"]" plugin/build; then
```
danielgranhao: Nit: `@expo/config-plugins/build/...` slips past this (quote required right after the name) and fails on SDK 56 the same way.

```suggestion
          if grep -rnE "(require\(|from )['\"]@expo/config-plugins(/|['\"])" plugin/build; then
```

danielgranhao (approved):
Left a nit but isn't blocking. Looks good

roeierez (approved):

PR #1039, opened 2026-08-07
Raise the operator gRPC decode limit to 20 MiB

danielgranhao:
Transfer pages embed full leaf data, so a page of leaf-heavy transfers can exceed tonic's 4 MiB decode default. The failure repeats on the same page every sync, so the payment sync offset stops advancing. Observed in production.

Raises the decode limit on the operator Spark and token service clients to 20 MiB, matching the Spark JS SDK.


dangeross (approved):
LGTM

roeierez (approved):

PR #1040, opened 2026-08-07
Log leaf-bearing RPC requests at trace instead of debug

danielgranhao:
Requests carrying per-leaf signing jobs render unbounded in leaf count: a 64-leaf swap produces a ~700KB log line, past the per-entry limit of backends like GCP Cloud Logging. `DEFAULT_FILTER` enables `spark=debug`, so these shipped on by default.

Covers the requests embedding `StartTransferRequest` or `ClaimPackage`. Filter-only and single-leaf requests are unchanged.

### DIFF
--- a/crates/spark/src/operator/rpc/spark_rpc_client.rs
+++ b/crates/spark/src/operator/rpc/spark_rpc_client.rs
@@ -23,7 +23,7 @@ use tonic::metadata::Ascii;
 use tonic::metadata::MetadataValue;
 use tonic::service::Interceptor;
 use tonic::service::interceptor::InterceptedService;
-use tracing::{debug, instrument};
+use tracing::{debug, instrument, trace};
 
 #[derive(Clone, Default)]
 pub struct QueryNodesPaginatedRequest {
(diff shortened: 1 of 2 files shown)


crates/spark/src/operator/rpc/spark_rpc_client.rs
```diff
@@ -118,7 +123,7 @@ impl SparkRpcClient {
         &self,
         req: StartTransferRequest,
     ) -> Result<StartTransferResponse> {
-        debug!("Calling start_transfer with request: {:?}", req);
+        trace!("Calling start_transfer with request: {:?}", req);
```
dangeross, line 126: `transfer.rs:366` already does `trace!("About to send start_transfer_request: {:?}", ...)` on the same struct right before this call, so at `spark=trace` we now render the ~700KB payload twice. Should we drop the other one?

dangeross (approved):
LGTM, with a nit

crates/spark/src/operator/rpc/spark_rpc_client.rs
danielgranhao: Good catch, dropped the caller-side log in 7719b47cc and kept this one since it covers all callers.

PR #1042, opened 2026-08-07
Switch default mainnet chain service to mempool.space

danielgranhao:
A user reported an on-chain deposit not showing up because chain requests were failing with rate-limit errors. Measurements against both providers (real \`/tx/:txid\` lookups, ramping request rate from a single IP) showed:

| | blockstream.info | mempool.space |
|---|---|---|
| First 429 | at 10 req/s | at 40 req/s |
| Recovery after backing off | ~1 hour (IP penalty box) | ~5 seconds |
| Median latency | ~495 ms | ~139 ms |

Blockstream's penalty box outlasts the SDK's ~8s retry budget by orders of magnitude, so a throttled device sees its deposit sync fail for the whole window. mempool.space's throttle clears within the existing retry budget.

The API surface is identical for every endpoint the SDK uses (verified against live responses, including spent/unspent outspends and broadcast). Mainnet now also uses the dedicated \`/v1/fees/recommended\` endpoint, as regtest already does.

### DIFF
--- a/crates/breez-sdk/core/src/sdk_builder.rs
+++ b/crates/breez-sdk/core/src/sdk_builder.rs
@@ -804,7 +804,7 @@ async fn resolve_storage(
 }
 
 /// Resolves the chain service: caller-supplied override → REST config → network
-/// default (Esplora on mainnet, mempool.space on regtest).
+/// default (mempool.space on mainnet, a hosted mempool instance on regtest).
 fn resolve_chain_service(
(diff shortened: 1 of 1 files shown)


danielgranhao:
Related: https://github.com/breez/spark-sdk/issues/1043

roeierez (approved):
Nice!

dangeross (approved):

PR #1046, opened 2026-08-10
Fix iOS CocoaPods builds for the Flutter and React Native packages

### LINKED ISSUE #1045: Flutter issues building iOS
We're integrating `breez_sdk_spark_flutter` into a production Flutter app and have hit two blocking issues in the iOS CocoaPods build path. Both are in packaging, not runtime behaviour, and both affect every CocoaPods consumer.

- **Package:** `breez_sdk_spark_flutter` v0.19.2 and v0.20.0-dev1 (both affected)

- **Build command:** `flutter build ios --release --no-codesign`

- **Podfile:** `use_frameworks!`, deployment target iOS 13.0

- **Flutter note:** The build log also warns `The following plugins do not support Swift Package Manager for ios: breez_sdk_spark_flutter`, which confirms the CocoaPods path is the only viable one for Flutter integrators today.

We're currently working around issue 1 with a `post_install` Podfile hook that injects a `module.modulemap`, and we have not yet shipped because of issue 2. A proper upstream fix would unblock every CocoaPods-based Flutter integrator.

---

### 1. `import PasskeyPRFHelperObjC` cannot resolve under CocoaPods

**File:** `packages/flutter/ios/Classes/PasskeyAssertionCore.swift` 

**Line:** `import PasskeyPRFHelperObjC`

**Problem:** `PasskeyPRFHelperObjC` is a Swift Package Manager target in `bindings/langs/swift/Sources/PasskeyPRFHelperObjC/`. The Flutter plugin's CocoaPods `podspec` compiles `PasskeyPRFHelper.h/.m` directly into the `breez_sdk_spark_flutter` module itself, so no separate `PasskeyPRFHelperObjC` module exists. The build fails with:

```

Swift Compiler Error: Unable to resolve module dependency: 'PasskeyPRFHelperObjC'

```

---

### 2. `Data?` unwrap missing in `BreezSdkSparkPasskeyPlugin.swift`

**File:** `packages/flutter/ios/Classes/BreezSdkSparkPasskeyPlugin.swift` 

**Function:** `handleCreatePasskey(_:result:)` 

**Line:** `"userId": registered.userId.base64EncodedString()`

**Problem:** `IosPasskeyCredential.userId` is declared as `Data?` (optional) in `PasskeyAssertionCore.swift`:

```swift

public struct IosPasskeyCredential {

    public let credentialId: Data

    public let userId: Data?   // ← optional

    ...

}

```

Calling `.base64EncodedString()` on `Data?` without unwrapping causes a Swift compiler error. The `aaguid` line just below it already handles this correctly:

```swift

"aaguid": registered.aaguid?.base64EncodedString() as Any?,

```

dangeross:
Two iOS CocoaPods build failures reported against `breez_sdk_spark_flutter`. Both are present in React Native too, since the packages share the same passkey sources.

Fixes #1045

### 1. `import PasskeyPRFHelperObjC` does not resolve

`PasskeyPRFHelperObjC` is a module only under SPM. Both podspecs compile `PasskeyPRFHelper.h/.m` into the same pod as the Swift, where the helper arrives through the pod's umbrella header instead. The import is now conditional, so the shared file works on both paths.

### 2. `userId` not unwrapped

`IosPasskeyCredential.userId` is `Data?` (nil on sign-in, where an assertion carries no attestation) and was passed to `base64EncodedString()` directly. Now matches the `aaguid` line beside it. The Dart and TS decode sites are relaxed to accept null; both public models already typed it nullable.

### 3. C++ headers in the React Native pod umbrella

Found while validating RN. The podspec declared no `public_header_files`, so every C++ and ObjC++ header became public, including one carrying `#error This file must be compiled as Obj-C++`. That was inert until the passkey feature added the first Swift to the pod, which forces CocoaPods to build an ObjC module from the umbrella. `public_header_files` is now narrowed to the one header the Swift needs. The podspec is generated, so the edit is also re-applied by `scripts/post-ubrn.js`.

### Prevention

No CI job compiled these sources: the Flutter job runs on `ubuntu-latest` and does `dart analyze` only. Adds an `ios-cocoapods-compile` job that typechecks both packages' iOS Swift against an ObjC-only umbrella header, the shape CocoaPods produces for a mixed ObjC/Swift pod. That job checks the sources compile; keeping the podspecs' umbrella ObjC-only is enforced separately by `rn-post-ubrn-check`.

### Verification

Built for device through real `pod install` + Xcode, from deleted `Pods/`:

| | Flutter | React Native |
|---|---|---|
| Device build (arm64 Release) | `Runner.app` built | `BUILD SUCCEEDED` |
| Passkey symbols in shipped binary | yes | yes |
| Each fix reverted individually | fails with the reported error | fails with the reported error |

### DIFF
--- a/crates/breez-sdk/bindings/langs/swift/Sources/BreezSdkSpark/PasskeyProvider.swift
+++ b/crates/breez-sdk/bindings/langs/swift/Sources/BreezSdkSpark/PasskeyProvider.swift
@@ -1,8 +1,6 @@
 import AuthenticationServices
 import Foundation
-#if canImport(PasskeyPRFHelperObjC)
 import PasskeyPRFHelperObjC
-#endif
 import Security
 
 /// iOS / macOS platform-specific options for [`PasskeyProvider`].
--- a/crates/breez-sdk/bindings/langs/swift/Sources/BreezSdkSpark/PasskeyProvider.swift
+++ b/crates/breez-sdk/bindings/langs/swift/Sources/BreezSdkSpark/PasskeyProvider.swift
@@ -1,6 +1,8 @@
 import AuthenticationServices
 import Foundation
+#if canImport(PasskeyPRFHelperObjC)
 import PasskeyPRFHelperObjC
+#endif
 import Security
 
 /// iOS / macOS platform-specific options for [`PasskeyProvider`].
--- a/packages/react-native/scripts/post-ubrn.js
+++ b/packages/react-native/scripts/post-ubrn.js
@@ -196,6 +196,32 @@ patchFile(
   }
 );
 
+// ---------------------------------------------------------------------------
+// 3. BreezSdkSparkReactNative.podspec
+// ---------------------------------------------------------------------------
+
+patchFile(
+  'BreezSdkSparkReactNative.podspec',
+  'public_header_files (keep C++ out of the ObjC umbrella)',
+  (content, label, relPath) => {
+    if (content.includes('s.public_header_files')) {
+      return content;
+    }
+    const anchor = '  s.vendored_frameworks = "build/RnBreezSdkSpark.xcframework"';
+    requireAnchor(content, anchor, label, relPath);
+    const injected = [
+      '  # The pod contains Swift, so CocoaPods builds an ObjC module from the umbrella',
+      '  # header, which is parsed as ObjC and not ObjC++. Every other header here is',
+      '  # C++ or reaches it transitively (BreezSdkSparkReactNative.h imports the',
+      '  # ObjC++ codegen spec under RCT_NEW_ARCH_ENABLED), so only the helper the',
+      '  # Swift sources need is public.',
+      '  s.public_header_files = "ios/PasskeyPRFHelper.h"',
+      anchor,
+    ].join('\n');
+    return content.replace(anchor, injected);
+  }
+);
+
 if (errors.length > 0) {
   console.error('');
   console.error(`[post-ubrn] ${errors.length} patch(es) failed:`);
--- a/packages/react-native/scripts/post-ubrn.js
+++ b/packages/react-native/scripts/post-ubrn.js
@@ -20,6 +20,12 @@
  *        - register BreezSdkSparkPasskeyModule alongside the generated
  *          UniFFI TurboModule so React Native can find it at runtime
  *
+ *   3. BreezSdkSparkReactNative.podspec
+ *        - narrow `public_header_files` to the ObjC PRF helper. The pod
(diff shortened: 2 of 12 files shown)


crates/breez-sdk/bindings/langs/swift/Sources/BreezSdkSpark/PasskeyProvider.swift
```diff
@@ -1,6 +1,8 @@
 import AuthenticationServices
 import Foundation
+#if canImport(PasskeyPRFHelperObjC)
 import PasskeyPRFHelperObjC
+#endif
```
erdemyerebasmaz: This guard does nothing here. The file only ships in the SPM package, where the module always exists. The pods do not get a copy of it.

The three `PasskeyAssertionCore.swift` copies do need it, and they say why in a comment. This one has no comment, so it looks necessary.

Remove it, or add the same comment.

scripts/check-ios-cocoapods.sh
```diff
@@ -0,0 +1,134 @@
...
+  exit 0
+fi
+
+IOS_SDK="$(xcrun --sdk iphoneos --show-sdk-path)"
+
+# Flutter.framework ships inside the Flutter SDK. subosito/flutter-action sets
+# FLUTTER_ROOT; locally, derive it from whichever flutter is on PATH.
+if [[ -z "${FLUTTER_ROOT:-}" ]]; then
+  if command -v flutter >/dev/null 2>&1; then
+    FLUTTER_ROOT="$(cd "$(dirname "$(readlink -f "$(command -v flutter)" 2>/dev/null || command -v flutter)")/.." && pwd)"
```
erdemyerebasmaz: Locally this derives the root from `flutter` on PATH. Version managers (fenv, fvm, asdf, mise) put a wrapper script there, so the root is wrong and the script reports a missing `Flutter.xcframework`.

Ask Flutter for its root instead:

```suggestion
    FLUTTER_ROOT="$(flutter --version --machine | sed -n 's/.*"flutterRoot": *"\([^"]*\)".*/\1/p')"
```

which works for all install types.

packages/react-native/scripts/post-ubrn.js
erdemyerebasmaz, line 22: Docblock is now incomplete, patch 3 is not listed.

erdemyerebasmaz (changes requested):
These findings looked important.

erdemyerebasmaz (approved):
LGTM

danielgranhao (approved):

PR #1047, opened 2026-08-10
Skip signature verification on unsigned direct refund transactions

dangeross:
The three `spark-itest` deposit tests have failed on every `main` run since 2026-08-07, across all storage backends (in-memory, postgres, mysql):

- `deployed_deposit_tests::test_non_static_deposit_with_faucet`
- `deployed_deposit_tests::test_non_static_deposit_then_coop_withdraw`
- `deployed_wallet_settings_tests::test_master_identity_grants_read_access`

All fail in `claim_deposit` with:
```
Service error: bitcoin error: invalid transaction: signed tx input has empty witness
```

### Cause

Spark operator commit [`64272c0d1`](https://github.com/buildonspark/spark/commit/64272c0d12755fc805c010c11cfb29adf9a3e6bf), "Strip direct transaction witnesses from public API (#7898)", added a gRPC interceptor that strips witnesses from every nested `TreeNode` direct transaction on `SparkService`, to stop callers broadcasting paths meant to remain operator-controlled. The interceptor re-encodes the field with `SerializeTxNoWitness`, so it still decodes but carries no signature.

`FinalizeDepositTreeCreation` is a `SparkService` RPC, so its returned `direct_from_cpfp_refund_tx` now arrives unsigned, and `create_tree_root` verified it unconditionally. The CPFP `node_tx` and `refund_tx` are untouched by the interceptor and still verify.

### Change

Verify `direct_from_cpfp_refund_tx` only when it carries a witness. `node_tx` and `refund_tx` stay unconditional.

Conditional rather than removed so the check still runs against operators predating [`64272c0d1`](https://github.com/buildonspark/spark/commit/64272c0d12755fc805c010c11cfb29adf9a3e6bf), which is where the local itests exercise it. No security is lost: current operators strip that witness for every caller, so no client can validate the field either way. A malformed non-empty transaction is still rejected by the `TreeNode` conversion.

### DIFF
--- a/crates/breez-sdk/breez-itest/tests/unilateral_exit.rs
+++ b/crates/breez-sdk/breez-itest/tests/unilateral_exit.rs
@@ -40,7 +40,7 @@ use spark_itest::helpers::{
     fund_p2tr_utxo_with_key, fund_p2wpkh_utxo, fund_p2wpkh_utxo_with_key, sign_cpfp_psbt_p2tr,
     submit_package_with_csv_retry,
 };
-use spark_wallet::{RefundOutput, is_ephemeral_anchor_output};
+use spark_wallet::is_ephemeral_anchor_output;
 
 // The exit fee rate the tests build at, in sat/vByte (the public API unit). 1
 // sat/vByte is the mainnet min-relay floor.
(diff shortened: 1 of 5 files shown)


danielgranhao:
>Conditional rather than removed so the check still runs against operators predating [64272c0d1](https://github.com/buildonspark/spark/commit/64272c0d12755fc805c010c11cfb29adf9a3e6bf), which is where the local itests exercise it. No security is lost: current operators strip that witness for every caller, so no client can validate the field either way. A malformed non-empty transaction is still rejected by the TreeNode conversion.

@dangeross Did you consider bumping the SO images in local itests and drop the check?

roeierez (approved):
LGTM.

danielgranhao (approved):

PR #1054, opened 2026-08-13
Cross-chain: reject token contracts as send recipients

danielgranhao:
A transfer to a token's own contract credits the contract's balance, and these contracts expose no way to spend it, so the funds are unrecoverable. Providers screen for this server-side but coverage varies by chain, and by then the Spark leg has already settled.

`prepare` now rejects a recipient matching any token contract the route listing surfaces for the destination chain, before any funds move.


roeierez (approved):
LGTM

roeierez (approved):

PR #1057, opened 2026-08-13
Validate early requested amount equals invoice amount if both exists

roeierez:


crates/breez-sdk/core/src/sdk/payments/prepare/bolt11.rs
```diff
@@ -59,6 +59,17 @@ fn validate_request(
         ));
     }
 
+    if let Some(amount_msat) = invoice_details.amount_msat
+        && let Some(amount) = request.amount
+    {
+        let invoice_amount_sats = u128::from(amount_msat.div_ceil(1000));
```
danielgranhao, line 65: Just a note that this rounds up but later in line 152 we round down, so a 1500 msat invoice prepares as 1 sat if you omit `amount` and rejects `amount` 1 while requiring 2. This is resolved when taken together with #1056, which also changes that line to round up.

danielgranhao (approved):
LGTM

PR #1059, opened 2026-08-17
Prepare Payment Link using Cash App

### LINKED ISSUE #907: Pay from cash app to stables
One of the most powerful use cases is fiat to stables via cash app. Currently, users can top up using cash app, but we need to make it easier for them to pay external bolt11 invoices using cash app, but more importantly once we support USDT/USDC to make it as easy as possible to pay to USDT/USDC using cash app.

dangeross:
Adds a `prepare_payment_link` API to quote a cross-chain order and return a `cash.app` deep link an external payer opens to send USDC/USDT and have it delivered to a recipient on an external chain (EVM/Solana/Tron). The SDK is orchestrator only, so no wallet funds move and nothing is written to payment history.

Closes #907


kingonly:
MoonPay doesn't make sense because it's slow

crates/breez-sdk/core/src/sdk/api.rs
```diff
@@ -436,6 +446,196 @@ impl BreezSdk {
...
+            self.config
+                .cross_chain_config
+                .as_ref()
+                .and_then(|c| c.default_slippage_bps),
+        )?;
+
+        // The `amount` is a 6-decimal USD figure. Convert it to the sats the
+        // provider's `prepare` expects (for both fee modes) via the live rate.
+        let btc_usd = fetch_btc_usd_rate(self.cross_chain_context.fiat_service().as_ref()).await?;
+        let source_sats = convert_destination_amount_to_sats(amount, btc_usd, 6)?;
```
danielgranhao: Worth validating that right after this that `source_sats` is above zero? A sub-sat USD amount can result in 0 sats which will then fail with a orchestra service error instead of a clear "amount too small".

danielgranhao: nit: Pass route.decimals instead of literal `6`?

crates/breez-sdk/core/src/sdk/api.rs
```diff
@@ -436,6 +446,196 @@ impl BreezSdk {
...
+        // The provider returns the BOLT11 the external payer funds in its
+        // context. We build the `cash.app` deep link ourselves and never call
+        // the send stage: the external payer funds the provider's deposit
+        // directly, so there is no Spark-side transfer to make and the provider
+        // delivers autonomously once the payer settles.
+        let deposit_target = deposit_target(&prepared.provider_context);
+
+        // Verify the provider's target is a BOLT11 before sending the payer to it.
+        let parsed_target = self.parse(&deposit_target).await;
+        check_deposit_target(&parsed_target)?;
```
danielgranhao: Should we also check that the invoice has the expected amount?

danielgranhao (commented):
Looks good

roeierez (approved):
LGTM

danielgranhao (approved):
LGTM

PR #1061, opened 2026-08-18
Base the refund fee floor on the real transaction size

### LINKED ISSUE #1049: Refund fee floor of 194 sats rejects valid 1 sat/vB refunds
`MIN_REFUND_FEE_SATS = 194` (`crates/spark/src/services/deposit.rs:45`) is checked against a fee computed from the real signed transaction size (`deposit.rs:368-375`). That size is 111 vB for a P2TR refund address, 99 for P2WPKH, 102 for P2PKH, so `Fee::Rate { sat_per_vbyte: 1 }` yields 111 sats and is rejected. The floor is effectively 1.75 to 1.96 sat/vB. The constant's own comment describes it as a "194 vbyte estimate for 1-in/1-out tx at 1 sat/vB", so 1 sat/vB was the intent, but 194 is nearly twice the real size.

Impact: the floor costs 83 to 95 sats over the true 1 sat/vB fee, raising the minimum refundable deposit by the same amount, and any `recommended_fees` rate below ~2 sat/vB is unusable. A 500 sat deposit refunded to taproot leaves 306 sats, under the 330 sat dust limit, and fails; at 1 sat/vB it would leave 389 and succeed. #574 lowered the constant from 300 to 194 without catching this.

danielgranhao:
Fixes #1049.

Now 1 sat/vB over the real signed size, matching the deposit claim path. The floor peaks at 111 sats, so no previously accepted fee is rejected.

### DIFF
--- a/crates/spark/src/services/deposit.rs
+++ b/crates/spark/src/services/deposit.rs
@@ -54,9 +54,10 @@ const INSTANT_UTXO_SWAP_REQUEST_TYPE: u64 = 3;
 /// changes the operators' share, not the user's.
 const STATIC_DEPOSIT_KEY_INDEX: u32 = 0;
 
-// Conservative minimum fee threshold for refund transactions
-// Based on 194 vbyte estimate for 1-in/1-out tx at 1 sat/vB minimum relay fee.
-const MIN_REFUND_FEE_SATS: u64 = 194;
+/// Bitcoin Core's default minimum relay fee rate. The refund fee floor is this
+/// rate applied to the real signed size of the refund transaction, which varies
+/// with the refund address type.
+const MIN_RELAY_FEE_SAT_PER_VBYTE: u64 = 1;
 
 /// Witness vbytes for a single Schnorr signature: ceil(66 witness bytes / 4)
 /// Witness structure: 1 (stack items) + 1 (sig length varint) + 64 (signature) = 66 bytes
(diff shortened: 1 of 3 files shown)


crates/spark/src/services/deposit.rs
```diff
@@ -1543,4 +1544,33 @@ mod tests {
...
+            txid: Txid::from_slice(&[7u8; 32]).unwrap(),
+            vout: 0,
+        };
+        let key = test_key(1);
+        let refund_addresses = [
+            p2tr(&key),
+            Address::p2wpkh(
+                &CompressedPublicKey(bitcoin::PublicKey::new(key).inner),
+                NETWORK,
+            ),
```
dangeross: could it be simplified to this?
```suggestion
            Address::p2wpkh(&CompressedPublicKey(key), NETWORK),
```

dangeross (approved):
LGTM, with a (maybe) nit

crates/spark/src/services/deposit.rs
dangeross: no need 👍

danielgranhao: Done in 54df11365 `rejects_non_taproot_output_for_the_verifying_key` now uses the same short form, and it fits on one line.

roeierez (approved):
LGTM

PR #1063, opened 2026-08-20
fix(react-native): resolve currentActivity on RN 0.80+

erdemyerebasmaz:
## Problem

The Android passkey module does not compile on React Native 0.85: `currentActivity` is an unresolved reference, and the `Any` vs `Activity` errors follow from it.

RN 0.80 converted `ReactContextBaseJavaModule` to Kotlin, so `getCurrentActivity()` is now a Kotlin function rather than a Java getter. Kotlin synthetic property access does not apply, so the bare reference no longer resolves.

## Change

Use `reactContext.currentActivity` at the 3 call sites. `ReactContext` is still Java, so the property form resolves on RN 0.78 (the current devDependency) and on 0.85. This is the replacement RN's own deprecation notice gives.


dangeross (approved):
LGTM

PR #1066, opened 2026-08-20
Remove the use of axum_extra::extract::Host

yaslama:
It's deprecated and we can always assume that Host is the right host

### DIFF
--- a/crates/breez-sdk/lnurl/src/routes.rs
+++ b/crates/breez-sdk/lnurl/src/routes.rs
@@ -1493,7 +1493,7 @@ where
             return Ok(Host(host.to_string()));
         }
 
-        if let Some(authority) = parts.uri.authority().map(|a| a.as_str()) {
+        if let Some(authority) = parts.uri.authority() {
             return Ok(Host(authority.to_string()));
         }
 
--- a/crates/breez-sdk/lnurl/src/routes.rs
+++ b/crates/breez-sdk/lnurl/src/routes.rs
@@ -1,11 +1,10 @@
 use axum::{
     Extension, Json,
     body::Bytes,
-    extract::{Path, Query},
-    http::{HeaderMap, StatusCode},
+    extract::{FromRequestParts, Path, Query},
+    http::{HeaderMap, StatusCode, request::Parts},
     response::IntoResponse,
 };
-use axum_extra::extract::Host;
 use bitcoin::{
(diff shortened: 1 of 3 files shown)


crates/breez-sdk/lnurl/src/routes.rs
```diff
@@ -1481,6 +1480,30 @@ async fn sanitize_domain<DB>(
...
+    S: Send + Sync,
+{
+    type Rejection = (StatusCode, Json<Value>);
+
+    async fn from_request_parts(parts: &mut Parts, _state: &S) -> Result<Self, Self::Rejection> {
+        let host = parts
+            .headers
+            .get("host")
+            .and_then(|h| h.to_str().ok())
+            .and_then(|h| h.split(':').next())
```
danielgranhao: This drops the port, but the allow-list keys keep it. For for example `Host: localhost:8080` now looks up `localhost` and misses.  localhost:8080 is the documented dev setup (default --domains)

Also turns `[::1]:8080` into `[`

crates/breez-sdk/lnurl/src/routes.rs
```diff
@@ -1481,6 +1480,30 @@ async fn sanitize_domain<DB>(
...
+            .ok_or_else(|| {
```
danielgranhao: At least if we keep the split above, we may get `Host("")` if the header is `Host:`. Worth checking for empty here.

crates/breez-sdk/lnurl/src/routes.rs
```diff
@@ -1481,6 +1480,30 @@ async fn sanitize_domain<DB>(
...
+impl<S> FromRequestParts<S> for Host
+where
+    S: Send + Sync,
+{
+    type Rejection = (StatusCode, Json<Value>);
+
+    async fn from_request_parts(parts: &mut Parts, _state: &S) -> Result<Self, Self::Rejection> {
+        let host = parts
+            .headers
+            .get("host")
```
danielgranhao: HTTP/2 puts the host in `:authority`, not a header, and `axum::serve(..).into_make_service()` accepts h2c. Those requests get a 400 now. Should we keep the `parts.uri.authority()` fallback the old extractor had?

danielgranhao (approved):

roeierez (approved):

PR #1067, opened 2026-08-20
Quote deposit claims to inform user selection

dangeross:
This PR adds an API that prices both ways of claiming so an app can show the trade and let the user pick, and unifies the early and mature claim paths under one fee ceiling.

`fetch_claim_deposit_quote` returns the deposit's current depth with both options priced:

```
{
  amount_sats: 20000, 
  confirmations: 1,
  instant: { confirmations_required: 1, fee_sats: 347, credit_amount_sats: 19653, is_estimate: false },
  mature:  { confirmations_required: 3, fee_sats: 198, credit_amount_sats: 19802, is_estimate: true }
}
```

`confirmations_required` is the depth an option becomes claimable at, so the wait is that minus `confirmations`. `is_estimate` flags the mature fee before the provider will quote it. `instant` is absent when there is no early option worth showing. Acting on the user's choice means passing a `max_fee` of at least the chosen `fee_sats`.

The SDK claims a deposit early whenever the provider's spread fits within the configured `max_deposit_claim_fee`.

Glow PR https://github.com/breez/glow-web/pull/388


crates/breez-sdk/core/src/sdk/deposits.rs
```diff
@@ -39,15 +49,31 @@ impl BreezSdk {
...
+                u32::MAX
+            });
+        // Immature deposits take the early path, bounded by the same ceiling.
+        if !self
+            .is_deposit_mature_at(&detailed_utxo, confirmations)
+            .await?
+        {
             return self
-                .instant_claim_deposit(&detailed_utxo, max_instant_fee_bps)
+                .instant_claim_deposit(&detailed_utxo, max_fee, confirmations)
```
roeierez, line 91: Should we only protect in flight claims that are are immature? Or this guard belongs also to matured claims?

roeierez (commented):
Looks good

crates/breez-sdk/core/src/sdk/deposits.rs
dangeross: I considered it. The sync flow is guarded on both branches, but thought its maybe better to leave the manual mature branch unguarded against the in-flight claim. If the immature claim is stuck somehow but seen submitted, there is no way to force a mature claim. 

Wdyt @roeierez, better to guard it here or let the claim go through let the SSP decide?

crates/breez-sdk/core/src/sdk/deposits.rs
```diff
@@ -290,112 +550,164 @@ impl BreezSdk {
...
+                        max_fee_sats: None,
                     }),
                 }
             }
             InstantClaimPlan::NoPlan => Ok(InstantClaimOutcome::Declined {
-                error: SdkError::Generic("No instant (0-conf) claim plan available".to_string()),
-                reason: InstantClaimDeclineReason::NoPlan,
+                error: SdkError::Generic("No instant claim plan available".to_string()),
+                max_fee_sats: None,
             }),
```
danielgranhao, line 559: Could an initial empty plan simply mean the SSP hasn't indexed the tx yet? If so, it would make sense not to store it as a decline.

danielgranhao (commented):
Just a question about something I'm not sure about

crates/breez-sdk/core/src/sdk/deposits.rs
dangeross: NoPlan here means we rejected the plan the SSP offered, because it needs more confirmations than the deposit has. If the SSP genuinely can't find the tx it throws instead, which we treat as transient and retry. Otherwise it does return a plan, just a 1-conf one for an unconfirmed deposit rather than a 0-conf one.

roeierez (approved):
LGTM

danielgranhao (approved):
LGTM

PR #1069, opened 2026-08-20
Cross-chain send fixes

dangeross:
An Orchestra cross-chain send polls its own outbound Spark transfer to learn when it settles. That poll dispatched on transfer status alone, so an outgoing transfer sitting at `SenderKeyTweaked` was routed into the claim path: the
wallet tried to decrypt leaf keys it had just encrypted for the recipient, which cannot succeed. Every send logged
`Failed to claim transfer: signer error: generic error: failed to decrypt: invalid message` at least once, burned an operator round trip, and sat in claim backoff before returning. No funds were ever at risk, because an outgoing transfer has no leaf for us to claim.

The same sends also returned a `Payment` with `conversion_info` and `conversion_details` unset, so callers could not read the Orchestra order id that lets them query an order's progress.

### Unused code

`utils/conversions.rs` had a `#![allow(dead_code)]` hiding five unreferenced functions (`get_payment_with_conversion_details`, `needs_enrichment`, `enrich_payment`, `fold_conversion_status`, `build_conversions`), four of which landed in `utils/payments.rs`. The status fold was dropped, since the stored status is used as-is. All removed, along with the allow.


crates/breez-sdk/core/src/cross_chain/orchestra.rs
danielgranhao: Not from this PR, but should this be `true`? payment_id is the Orchestra output's vout (resolved with tx_inputs_are_ours = true), and false only yields outputs owned by us, so the lookup never matches and every token cross-chain send waits out the full 30s before falling back to the Err branch, which passes true anyway.

danielgranhao (commented):
Looks good

crates/breez-sdk/core/src/cross_chain/orchestra.rs
dangeross: Nice catch

danielgranhao (approved):

PR #1073, opened 2026-08-21
docs: Publish per-language LLM markdown editions

dangeross:
The `.md` files we publish for LLMs were expanded HTML in a markdown wrapper: 274 `<custom-tabs>` widgets, 5,112 nine-way `<span>` copies of identifiers, and 67 section headings that were raw `<h2>` so no heading existed at all. 

This PR makes the preprocessor renderer-aware, so the website keeps its tabs and language spans while every other
renderer gets real markdown, and adds a per-language edition of the whole book. It also documents what each `SdkEvent` actually carries, which is what prompted the work.

### Published structure
Additive. `/guide/*.md` and `/llms-full.txt` keep their current paths and shape.

| Path | Contents |
| --- | --- |
| `/llms.txt` | Index. Language editions listed first, each tagged with the dependency it matches |
| `/guide/events.md` | Canonical page, examples in all nine languages (unchanged) |
| `/llms-full.txt` | Whole book, all languages |
| `/llms-rust-full.txt` | Whole book, one language. Nine of these, around 330 KB against 1.01 MB |
| `/llms/rust/llms.txt` | Per-edition index |
| `/llms/rust/guide/events.md` | Per-edition page, 69 per edition |

An edition carries one language throughout, with identifiers in that language's naming convention: `get_info` in Rust, `GetInfo` in Go, `getInfo` in Kotlin.

### Before and after

`guide/get_info.md` opened with a raw `<h1>`, so the page had no title, and its first sentence cost 473 bytes.

Before:

    <h1 id="getting-the-sdk-info">
        <a class="header" href="#getting-the-sdk-info">Getting the SDK info</a>
        <a class="tag" href="...#method.get_info">API docs</a>
    </h1>

    Once connected, you can retrieve the current state of the SDK at any time using <code
    class="lang-fn"><span class="fn-rust">get_info</span><span class="fn-python">get_info
    </span><span class="fn-swift">getInfo</span> ...six more spans... </code>. This returns:

After, 106 bytes:

    # Getting the SDK info

    API docs: https://breez.github.io/spark-sdk/...#method.get_info

    Once connected, you can retrieve the current state of the SDK at any time using
    `get_info`. This returns:

Snippet blocks were bold paragraphs, invisible to any outline or chunker. They are now headings nested under the enclosing section.

Before:

    <h2 id="add-event-listener">
        <a class="header" href="#add-event-listener">Add event listener</a>
        <a class="tag" target="_blank" href="https://breez.github.io/spark-sdk/...#method.add_event_listener">API docs</a>
    </h2>

    <custom-tabs category="lang">
    <div slot="title">Rust</div>
    <section>

    ```rust,ignore

After:

    ## Add event listener

    API docs: https://breez.github.io/spark-sdk/...#method.add_event_listener

    ### Rust

    ```rust

### Events

`events.md` gains a payload reference: all ten events with their payload type and what to do, plus the `DepositInfo` fields that drive the decisions. Six `SdkEvent` variants had no doc comment at all and `DepositInfo` fields were undocumented. Snippet comments across all nine languages now name the payload and the field that matters, so `NewDeposits` explains what `is_mature` means rather than restating the variant name.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -29,7 +29,6 @@ jobs:
           cargo install --path ./snippets-processor
           cargo install mdbook-variables --vers "^0.2" --locked
           cargo install mdbook-pagetoc --vers "^0.2" --locked
-          cargo install mdbook-llms-txt-tools
 
       - name: Get latest tag
         uses: actions-ecosystem/action-get-latest-tag@v1
--- a/crates/breez-sdk/core/src/events.rs
+++ b/crates/breez-sdk/core/src/events.rs
@@ -19,38 +19,40 @@ use crate::{DepositInfo, LightningAddressInfo, Payment, sdk::RuntimeEvent};
 pub enum SdkEvent {
     /// Emitted when the wallet has been synchronized with the network
     Synced,
-    /// Emitted when the SDK was unable to claim deposits
+    /// Emitted when the SDK was unable to claim deposits. Each deposit carries a
+    /// `claim_error` with the reason.
     UnclaimedDeposits {
         unclaimed_deposits: Vec<DepositInfo>,
     },
-    ClaimedDeposits {
-        claimed_deposits: Vec<DepositInfo>,
-    },
-    PaymentSucceeded {
-        payment: Payment,
-    },
-    PaymentPending {
-        payment: Payment,
-    },
-    PaymentFailed {
-        payment: Payment,
-    },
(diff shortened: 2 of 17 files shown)


dsbreez:
Awesome, built it locally and went through. Dependency tags awesome, so it can correctly route now.

dsbreez (approved):

PR #1074, opened 2026-08-22
Fix Postgres verify-ca and verify-full pool initialization

aljazceru:
The pool selects extended SSL modes before constructing the Rustls connector, but tokio-postgres only accepts disable/prefer/require while parsing its Config. Normalize verify-ca and verify-full to require for the driver Config while retaining the original extended mode for the verifying Rustls connector. Adds regression coverage for both extended modes.


danielgranhao (approved):
LGTM

roeierez (approved):
LGTM

PR #1076, opened 2026-08-24
Add SDK-wide SOCKS5 proxy support

### LINKED ISSUE #1048: Add SDK-wide SOCKS5 proxy support
Breez SDK Spark creates its own HTTP and gRPC connections but does not expose a global proxy configuration.

  Could the SDK provide an optional SDK-wide SOCKS5 proxy setting that:

  - Applies to every HTTP and gRPC connection.
  - Resolves hostnames through the proxy to prevent DNS leaks.
  - Fails closed instead of falling back to direct connections.

  This would enable integrators to reliably route all SDK traffic through Tor or another SOCKS5 proxy without maintaining a custom SDK fork.

danielgranhao:
Closes #1048. Native only, fail-closed, no global state.

- `proxy` on `Config` and `SdkContextConfig`, cross-checked in `resolve_context`.
- HTTP via reqwest `socks5h://` (remote DNS); gRPC via a tower connector, TLS untouched.
- BIP353 switches to DoH when proxied: plain DNS is UDP and would escape the tunnel.
- Rejected at `connect`: WASM, `connections_per_operator > 1`, half-specified credentials.
- Consolidated the ad-hoc clients so everything routes through one place. Bulk of the diff.
- `clippy.toml` bans direct `reqwest::Client` construction, so a new ad-hoc client fails the build.
- Passkey rejects a credentialed proxy at construction: relays can't authenticate to a proxy,
  and the label it would fail to publish is a PRF salt the wallet seed derives from.
- Boltz stays unproxied and is not rejected; Orchestra is on the shared client, so cross-chain works.
- CLI gains `--proxy`, `--proxy-user`, `--proxy-password`.
- Itests run against a real SOCKS5 container, including a dead-proxy case asserting no direct fallback.

API breaks: `get_spark_status` takes a request record, `PasskeyClient::new` / `from_config`
are fallible, `parse_input` removed.

### DIFF
--- a/crates/breez-sdk/core/src/models/mod.rs
+++ b/crates/breez-sdk/core/src/models/mod.rs
@@ -619,6 +619,23 @@ impl From<&ProxyConfig> for platform_utils::ProxyConfig {
 }
 
 impl ProxyConfig {
+    /// A validated HTTP client honouring `proxy`, for the components built
+    /// outside the SDK's shared context. Everything inside it takes the
+    /// context's pooled client instead and never sees a proxy.
+    pub(crate) fn http_client(
+        proxy: Option<&Self>,
+        user_agent: Option<&str>,
+    ) -> Result<std::sync::Arc<dyn platform_utils::HttpClient>, SdkError> {
+        if let Some(proxy) = proxy {
+            proxy.validate()?;
+        }
+        platform_utils::create_http_client_with_proxy(
+            user_agent,
+            proxy.map(platform_utils::ProxyConfig::from).as_ref(),
+        )
+        .map_err(|e| SdkError::InvalidInput(format!("Failed to build proxied HTTP client: {e}")))
+    }
+
     /// Rejects a proxy the SDK cannot honour end to end. Accepting one it can
     /// only partly apply would leave some traffic going direct, which is worse
     /// than refusing outright.
--- a/crates/breez-sdk/core/src/models/mod.rs
+++ b/crates/breez-sdk/core/src/models/mod.rs
@@ -582,6 +582,74 @@ impl FromStr for Network {
     }
 }
 
+/// A SOCKS5 proxy carrying the connections the SDK opens.
+///
+/// Hostnames are resolved by the proxy rather than locally, so a DNS query
+/// never discloses which host is being reached. A connection that cannot be
+/// established through the proxy fails: the SDK never falls back to a direct
+/// one.
+///
+/// Not supported on WASM, where the browser owns connection setup and exposes
+/// no proxy control. In Node, route the SDK by installing a proxy dispatcher
+/// on the global `fetch` instead.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct ProxyConfig {
+    /// Proxy host. An IP address, or a name resolvable locally: reaching the
+    /// proxy is the one lookup that cannot itself go through the proxy.
+    pub host: String,
+    pub port: u16,
+    /// Username for SOCKS5 username/password authentication. Authentication is
+    /// only offered when both this and `password` are set.
+    #[cfg_attr(feature = "uniffi", uniffi(default = None))]
+    pub username: Option<String>,
+    #[cfg_attr(feature = "uniffi", uniffi(default = None))]
+    pub password: Option<String>,
+}
+
+impl From<&ProxyConfig> for platform_utils::ProxyConfig {
+    fn from(config: &ProxyConfig) -> Self {
+        Self {
+            host: config.host.clone(),
+            port: config.port,
+            username: config.username.clone(),
+            password: config.password.clone(),
+        }
+    }
+}
+
+impl ProxyConfig {
+    /// Rejects a proxy the SDK cannot honour end to end. Accepting one it can
+    /// only partly apply would leave some traffic going direct, which is worse
+    /// than refusing outright.
+    pub(crate) fn validate(&self) -> Result<(), SdkError> {
+        if cfg!(all(target_family = "wasm", target_os = "unknown")) {
+            return Err(SdkError::InvalidInput(
+                "A SOCKS5 proxy is not supported on WASM: the browser owns connection setup and \
+                 exposes no proxy control. In Node, install a proxy dispatcher on the global \
+                 fetch instead."
+                    .to_string(),
+            ));
+        }
+        if self.host.is_empty() {
+            return Err(SdkError::InvalidInput(
+                "Proxy host must not be empty".to_string(),
+            ));
+        }
+        if self.port == 0 {
+            return Err(SdkError::InvalidInput(
+                "Proxy port must not be 0".to_string(),
+            ));
+        }
+        if self.username.is_some() != self.password.is_some() {
+            return Err(SdkError::InvalidInput(
+                "Proxy username and password must be set together".to_string(),
+            ));
+        }
+        Ok(())
+    }
+}
+
 #[derive(Debug, Clone)]
 #[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
 #[allow(clippy::struct_excessive_bools)]
@@ -705,6 +773,16 @@ pub struct Config {
     /// must be `None`, and `optimization_config.auto_enabled` must be `false`.
     /// `default_server_config` already sets these compatible values.
     pub background_tasks_enabled: bool,
+
+    /// Routes the connections the SDK opens through a SOCKS5 proxy.
+    ///
+    /// Covers HTTP and gRPC alike, and resolves hostnames at the proxy so no
+    /// DNS query leaks the destination. `None` (default) connects directly.
+    ///
+    /// When an [`SdkContext`](crate::SdkContext) is supplied to the builder,
+    /// its proxy must match this one: the context owns the shared clients, so
+    /// a disagreement would mean part of the traffic bypassed the proxy.
+    pub proxy: Option<ProxyConfig>,
 
     /// Configuration for cross-chain sends via Orchestra and Boltz.
     ///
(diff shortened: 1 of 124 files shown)


crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -582,6 +582,74 @@ impl FromStr for Network {
...
+    pub(crate) fn validate(&self) -> Result<(), SdkError> {
+        if cfg!(all(target_family = "wasm", target_os = "unknown")) {
+            return Err(SdkError::InvalidInput(
+                "A SOCKS5 proxy is not supported on WASM: the browser owns connection setup and \
+                 exposes no proxy control. In Node, install a proxy dispatcher on the global \
+                 fetch instead."
+                    .to_string(),
+            ));
+        }
+        if self.host.is_empty() {
```
dangeross: `"10.0.0.1 "` gets through this and then panics building the HTTP client, since `reqwest_url()` interpolates the host into a URL without checking it. Should we validate it more strictly here?

crates/platform-utils/src/http/client.rs
```diff
@@ -38,26 +39,57 @@ impl ReqwestHttpClient {
...
             }
+            if let Some(proxy) = proxy {
+                match reqwest::Proxy::all(proxy.reqwest_url()) {
+                    Ok(proxy) => builder = builder.proxy(proxy),
+                    Err(e) => {
+                        // Continuing would build a client that connects
+                        // directly, which is the one outcome a proxied client
+                        // must never produce.
+                        tracing::error!("Failed to build SOCKS5 proxy: {e}");
+                        panic!("Failed to build SOCKS5 proxy: {e}");
```
dangeross: Can we return an error here rather than panic? `doh::build_client` makes the same `Proxy::all` call and propagates with `?`.

crates/platform-utils/src/proxy/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+        }
+    }
+
+    /// The `socks5h` URL reqwest needs. The `h` suffix is what makes reqwest
+    /// hand the hostname to the proxy instead of resolving it locally.
+    ///
+    /// Credentials are carried in the URL userinfo, percent-encoded so that a
+    /// password containing `:`, `@` or `/` cannot break out of its component.
+    #[must_use]
+    pub fn reqwest_url(&self) -> String {
```
dangeross: Currently the host validation allows an `@` in the host, say someone pastes `breez@proxy.internal`. With user `u` and password `p` that gives us:

```
socks5h://u:p@breez@proxy.internal:9050
```

The last `@` wins when that gets parsed, so we connect to `proxy.internal` and send `p@breez` as the password. `Proxy::all(...).basic_auth(user, pass)` would set the credentials on the parsed URL instead.

Not asking for changes if the validation tightens up.

crates/platform-utils/src/proxy/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+            (Some(user), Some(pass)) => Some((user.as_str(), pass.as_str())),
+            _ => None,
+        }
+    }
+
+    /// `host:port`, the form both the reqwest proxy URL and the SOCKS5
+    /// connector dial.
+    ///
+    /// A bare IPv6 literal is bracketed: without brackets its own colons read
+    /// as the port separator, and neither a URL nor a socket address parses.
```
dangeross: Currently the host validation allows a trailing `/`, say someone pastes `proxy.internal/` with port 9050. That gives us:

```
HTTP  ->  proxy.internal:1080
gRPC  ->  proxy.internal/:9050
```

The `/` ends the host, so the URL carries no port and reqwest falls back to the `socks5h` default of 1080. The connector dials the literal string and fails DNS. So `host:port` only holds here for a well formed host.

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -1076,6 +1137,7 @@ fn build_cross_chain_context(
                 Arc::clone(spark_wallet),
                 Arc::clone(storage),
                 Arc::clone(&cached_fiat),
+                Arc::clone(http_client),
```
dangeross, line 1140: Should we consider skipping registering the Boltz provider when a proxy is set? The only downside is a pre-existing swap would stop being claimable once a proxy is added.

crates/breez-sdk/core/src/passkey/passkey_client.rs
```diff
@@ -396,7 +400,7 @@ impl PasskeyClient {
         prf_provider: Arc<dyn PrfProvider>,
         sdk_config: &crate::Config,
         passkey_config: Option<PasskeyConfig>,
-    ) -> Self {
+    ) -> Result<Self, PasskeyError> {
         Self::new(prf_provider, sdk_config.api_key.clone(), passkey_config)
```
dangeross: nit: we forward `api_key` here but not `proxy`, so the label relays could connect direct even if a config proxy is set. Should we default it from `sdk_config.proxy` when unset?

docs/breez-sdk/src/guide/proxy.md
```diff
@@ -0,0 +1,82 @@
...
+
+A configuration the SDK cannot honour is rejected where it is supplied, rather
+than partly applied: at {{#name connect}}, or at the constructor of a component
+built outside the SDK.
+
+| Rejected combination | Why |
+|---|---|
+| A proxy on WASM | The browser owns connection setup and exposes no proxy control. |
+| {{#name proxy}} with {{#name connections_per_operator}} above 1 | Balanced operator connections build their own connectors and cannot be routed. |
+| A proxy carrying credentials on {{#name PasskeyConfig}} | Nostr relay connections cannot authenticate to a proxy. Wallet labels live on those relays and are one of the salts a wallet seed derives from, so a label that cannot be published cannot be recovered: this fails before a wallet exists rather than after one is funded. |
```
dangeross: Missing a row for cross-chain. Boltz connects direct when a proxy is set.

docs/breez-sdk/src/guide/proxy.md
```diff
@@ -0,0 +1,82 @@
...
+{{#tabs config:config-proxy}}
+
+Set {{#name username}} and {{#name password}} together for proxies that require
+RFC 1929 authentication. Setting only one of them is rejected.
+
+## What the proxy covers
+
+Both HTTP and gRPC traffic: the Spark operators and service provider, the Breez
+backend, real-time sync, the chain service, LNURL and Lightning address lookups,
+and external input parsers.
```
dangeross: nit: Flashnet isn't in this list. Orchestra proxies, but Boltz connects direct when a proxy is set.

docs/breez-sdk/src/guide/proxy.md
dangeross, line 1: I think the paragraphs should be on single lines to allow page resizing vs fixed columns

dangeross (commented):
Lookn good

roeierez (commented):
I wonder if we can do that in a way that connections are initiated (with or without proxy) in a central place.
For example Orchestra just gets an injected Arc<dyn HttpClient> and never needs to know a proxy exists. could the chain service, Turnkey and get_spark_status take the context (or its client) the same way, instead of adding a proxy field to each of them and every one rebuild its own client? That would drop the four duplicated validate + create_http_client_with_proxy blocks and enforce the proxy usage at a higher layer.
I didn't go all the way with it, just intuition.

crates/breez-sdk/core/src/sdk_builder.rs
danielgranhao: Ended up adding proxy support to boltz client

crates/breez-sdk/core/src/passkey/passkey_client.rs
danielgranhao: Done

docs/breez-sdk/src/guide/proxy.md
danielgranhao: Went more generic instead of listing all components

docs/breez-sdk/src/guide/proxy.md
danielgranhao: Fixed

danielgranhao:
> I wonder if we can do that in a way that connections are initiated (with or without proxy) in a central place. For example, Orchestra just gets an injected Arc and never needs to know a proxy exists. could the chain service, Turnkey and get_spark_status take the context (or its client) the same way, instead of adding a proxy field to each of them and every one rebuild its own client? That would drop the four duplicated validate + create_http_client_with_proxy blocks and enforce the proxy usage at a higher layer. I didn't go all the way with it, just intuition.

Extracted the repeated validation and http client builds to `ProxyConfig::http_client`. 

The injection pattern is what's already being done inside the SDK. What still carries a proxy is what's built before an SDK exists, and some of it can't take a client at all (the passkey relays need a socket address).

roeierez (approved):
LGTM

crates/flashnet/src/orchestra/client.rs
```diff
@@ -43,11 +43,12 @@ impl OrchestraClient {
     pub fn new(
         config_resolver: Arc<dyn OrchestraConfigResolver>,
         spark_wallet: Arc<SparkWallet>,
+        http_client: Arc<dyn HttpClient>,
     ) -> Self {
         Self {
             config_resolver,
             config: OnceCell::new(),
-            http_client: platform_utils::DefaultHttpClient::default(),
+            http_client,
```
dangeross, line 51: This is bugging me why it's not using the shared client. Maybe user agent issues with flashnet? Can you test it?

danielgranhao: I think it was probably unintentional. Tested using `mainnet_cross_chain::test_cross_chain_orchestra_fees_excluded_evm`

dangeross (approved):
LGTM

PR #1078, opened 2026-08-25
Fix resuming a unilateral exit

roeierez:
A caller can drive a unilateral exit in one shot: call it once, take the signed transactions, and broadcast them in order. The SDK also offers to track on-chain confirmations for you, so you don't have to. To use that, the caller re-calls the unilateral exit periodically and gets back whatever still needs broadcasting. That resume path was broken in two ways.

**The tree lost its exited leaves.** Every call refreshes leaves from the operators, and the query omitted the EXITED and ONCHAIN statuses. Leaves in those statuses were deleted from the store, so the next call no longer had the whole tree and the exit failed. The query now includes both statuses, and a leaf with an exit under way survives a refresh.

**A resumed exit asked for the full funding again.** Planning ran before any chain observation, so it priced a CPFP child for every node and every refund, including branches already confirmed. The caller was asked to fund the whole job a second time instead of just the work left. Chain observation now runs before the planner and is passed in as input, so the planner charges only for the children it still has to build. A leaf whose steps have all settled takes no funding at all.

### DIFF
--- a/crates/breez-sdk/core/src/sdk/unilateral_exit.rs
+++ b/crates/breez-sdk/core/src/sdk/unilateral_exit.rs
@@ -11,8 +11,8 @@ use bitcoin::{
 
 use spark_wallet::{
     AddressUtxo, ChainQuery, ChainResult, CpfpInput, ExitTxKind, ExitTxStatus, Observation,
-    PreparedUnilateralExit, SpendInfo, TreeNodeId, UnilateralExitBuild, build_unilateral_exit,
-    is_ephemeral_anchor_output, next_chain_queries,
+    PreparedUnilateralExit, SettledExitSteps, SpendInfo, TreeNodeId, UnilateralExitBuild,
+    build_unilateral_exit, is_ephemeral_anchor_output, next_chain_queries, scan_settlement,
 };
 
 use tracing::{debug, trace, warn};
@@ -197,15 +197,21 @@ impl BreezSdk {
         }
 
         let fee_rate_sat_per_kw = sat_per_kw_from_vbyte(prepared.fee_rate_sat_per_vbyte);
-        let prepared_exit = self
+        // Ask the chain what the exit has already done before sizing its funding:
+        // a step that is settled takes no CPFP child, and pricing one anyway can
+        // reject a nearly-finished exit at any amount its remaining work justifies.
+        let context = self
             .spark_wallet
-            .prepare_unilateral_exit_plan(
-                fee_rate_sat_per_kw,
-                spark_wallet::ExitLeafSelection::Specific(leaf_ids),
-                funding_inputs,
-                dest_script_len,
-            )
+            .load_exit_context(spark_wallet::ExitLeafSelection::Specific(leaf_ids))
             .await?;
+        let (settled, settlement_observations) = resolve_settled_steps(chain, &context).await?;
+        let prepared_exit = self.spark_wallet.plan_exit(
+            &context,
+            fee_rate_sat_per_kw,
+            funding_inputs,
+            dest_script_len,
+            Some(&settled),
+        )?;
         if prepared_exit.plan.selected_leaves.is_empty() {
             debug!("unilateral_exit: plan selected no leaves, returning empty result");
             return Ok(empty_exit_response());
@@ -227,7 +233,8 @@ impl BreezSdk {
             })
             .collect();
 
-        let observed = resolve_exit_observations(chain, &prepared_exit).await?;
+        let observed =
+            resolve_exit_observations(chain, &prepared_exit, settlement_observations).await?;
         let build = build_unilateral_exit(&prepared_exit, &observed, fee_rate_sat_per_kw)?;
         let recoverable_value_sat = build.recoverable_value_sat;
         let build_fee_sat = build.total_fee_sat;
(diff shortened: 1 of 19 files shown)


crates/breez-sdk/core/src/sdk/unilateral_exit.rs
```diff
@@ -197,15 +197,21 @@ impl BreezSdk {
...
             .spark_wallet
-            .prepare_unilateral_exit_plan(
-                fee_rate_sat_per_kw,
-                spark_wallet::ExitLeafSelection::Specific(leaf_ids),
-                funding_inputs,
-                dest_script_len,
-            )
+            .load_exit_context(spark_wallet::ExitLeafSelection::Specific(leaf_ids))
             .await?;
+        let (settled, settlement_observations) = resolve_settled_steps(chain, &context).await?;
```
danielgranhao, line 207: We don't use this in `prepare_unilateral_exit` above, and there we still quote the whole job. Shouldn't we do the same there?

crates/spark-wallet/src/unilateral_exit.rs
```diff
@@ -1252,6 +1326,24 @@ fn refund_output_value(
 type BranchFunding = Vec<(TreeNodeId, Vec<CpfpInput>)>;
 
 /// Resolves the fan-out step and the per-branch funding it feeds. A confirmed
```
danielgranhao: This should be moved down. `remaining_cpfp_cost` was inserted in the middle of previously existing rustdocs of `resolve_fan_out_funding`

crates/spark/src/tree/service.rs
danielgranhao: Now that we can also have Exited or Onchain leaves won't we try to make them go through `check_renew_nodes`? If such a leaf has its timelock almost gone, we will always try to renew it anyway. I think we should skip these leaves in `check_renew_nodes`.

Benign but easily avoidable wasted work.

crates/spark/src/services/unilateral_exit.rs
```diff
@@ -89,13 +78,41 @@ pub struct UnilateralExitPlan {
 /// Selects which leaves to exit and maps funding inputs to branches. Never
 /// fetches: works offline as long as `tree_nodes` holds each selected leaf's
 /// full ancestor chain.
```
danielgranhao: rustdocs should be moved down to `plan_unilateral_exit`

crates/spark/src/tree/service.rs
```diff
@@ -423,7 +422,7 @@ impl TreeService for SynchronousTreeService {
         } else {
             HashMap::new()
         };
```
danielgranhao, line 438: This doesn't include unavailable leaves (not returned by `get_verified_leaf_keys`), so we'll fetch each leaf's public key on every refresh (we want to avoid this, especially with remote signers).

crates/spark/src/services/unilateral_exit.rs
```diff
@@ -146,14 +183,14 @@ pub fn plan_unilateral_exit(
         let cpfp_cost = if inputs.len() > 1 {
             first_child_cpfp_floor(
                 &tree_nodes,
-                &selected[0].id,
+                &driving[0].id,
```
danielgranhao, line 192: When there's more than one funding inpute this re-costs the leaf through `first_child_cpfp_floor`, pricing a fresh exit and ignoring the discount. I think `first_child_cpfp_floor` could take `settled` and discount the leaf it recomputes. 

Same thing on line 771 causes a multi-leaf path to do an unnecessary fan-out.

crates/spark-wallet/src/unilateral_exit.rs
danielgranhao, line 770: `branch_leaf_ids` is every branch, but the fan-out only has an output per driving leaf now, so with one settled leaf and two driving, we end up zipping 2 outputs against 3 ids and hit the FundingUtxoConflict below on every resume. Could we build `branch_leaf_ids` from the branches that actually got funding?

crates/spark/src/services/unilateral_exit.rs
```diff
@@ -131,7 +155,20 @@ pub fn plan_unilateral_exit(
...
+    // never reaches the sweep.
+    let driving: Vec<UnilateralExitSelectedLeaf> = selected
+        .iter()
+        .filter(|leaf| leaf.cpfp_cost > 0)
+        .cloned()
+        .collect();
+
+    let (assigned, fan_out_psbt) = if driving.is_empty() {
+        (Vec::new(), None)
+    } else if driving.len() == 1 {
```
danielgranhao, line 177: A leaf settling can flip this from the fan-out arm to the single-leaf arm, and then the confirmed fan-out is never adopted. The last branch funds from the outpoint it has already spent. Keep the fan-out arm once one has gone out?

crates/spark/src/services/unilateral_exit.rs
```diff
@@ -121,7 +138,14 @@ pub fn plan_unilateral_exit(
         fee_rate_sat_per_kw,
     };
 
-    let selected = evaluate_unilateral_exit_leaf_costs(&tree_nodes, leaf_ids, &params, filter)?;
+    let mut selected = evaluate_unilateral_exit_leaf_costs(&tree_nodes, leaf_ids, &params, filter)?;
```
danielgranhao: Filter drops unprofitable leaves inside this call, before the discount below runs, so a half-exited leaf is priced as a fresh exit and Auto drops it for good when it could be profitable given the discount.

crates/spark-wallet/src/wallet.rs
```diff
@@ -1787,43 +1800,73 @@ impl SparkWallet {
     /// Prepares a unilateral exit of the selected leaves: loads the exit tree,
     /// plans it, and derives each leaf's P2TR refund address. `Auto` keeps only
     /// profitable leaves; `Specific` exits every requested one.
```
dangeross: Looks like this should be moved down to `plan_unilateral_exit`

JssDWt:
Replaced by #1095

PR #1079, opened 2026-08-25
Fix the failing integration test and LNURL CI jobs

danielgranhao:
Fixes the jobs failing on `main`. Five unrelated causes, one commit each.

**1. `generate_deposit_address` is disabled on the operators** — breaks all three Integration test jobs, which share the `spark-itest` suite. Since Aug 24 it returns `Unavailable`/`METHOD_DISABLED`, their knob-gated kill switch, on mainnet and regtest alike. Not a deprecation (no `deprecated` in the proto, handler live, JS SDK still exposes `getSingleUseDepositAddress()`); asked Lightspark why. The session and master-identity tests only used a non-static address incidentally, so they move to static via a new `fund_wallet_via_static_deposit` helper — a static claim charges a fee, so balance assertions follow the credited amount. The two tests covering the non-static flow itself are `#[ignore]`d.

**2. Lightning sends settle before `send_payment` returns**, so `test_06`'s `Pending` assertion always fails. No timeout value fixes it — `Some(0)` still returns `Completed`. It now covers the `completion_timeout_secs == 0` branch, accepting either status and asserting the send succeeds and credits exactly; `test_10` still pins the status down for non-zero timeouts.

**3. LNURL webhook tests raced their own writes** — deliveries are spawned, not awaited, so reading the row after a fixed sleep is a race. Ten sites now poll.

**4. `generatetoaddress` outlived the RPC timeout.** Refund timelocks start at 2000 blocks and were mined in one call, exceeding 60s on a slow host. Now batched at 100. Predates this PR (reproduces on `main`); `unilateral_exit` goes 26/3 → 29/0.

**5. A deposit claim could not retry.** `wait_for_claimed_event` treated any `UnclaimedDeposits` as fatal, though the SDK retries on the next sync. Only terminal reasons abort now.

Reviewer notes: item 2 is now a tolerant assertion — happy to delete `test_06` as superseded by `test_10` instead. Item 4 is pre-existing and could be split out if you'd rather keep this scoped.

### DIFF
--- a/crates/breez-sdk/breez-itest/src/helpers/mod.rs
+++ b/crates/breez-sdk/breez-itest/src/helpers/mod.rs
@@ -379,17 +379,24 @@ pub async fn count_unilateral_exit_state_changed_events(
 
 /// Wait for a deposit claim to succeed by listening to SDK events
 ///
+/// An `UnclaimedDeposits` event is not by itself a failure. The SDK emits one
+/// on every sync that finds a deposit it could not claim yet, including the
+/// transient case where the operators have not all seen the funding tx deep
+/// enough, and it retries on the next sync. Only a terminal reason aborts;
+/// anything else keeps waiting and is reported if the timeout wins.
+///
 /// # Arguments
 /// * `event_rx` - Event receiver channel from build_sdk
 /// * `timeout_secs` - Maximum time to wait in seconds
 ///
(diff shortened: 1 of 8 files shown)


roeierez (approved):

dangeross (approved):
LGTM

PR #1082, opened 2026-08-26
lnurl: limit address registrations per pubkey per domain

danielgranhao:
Adds an app-level rate limit to the LNURL server's register endpoint: each pubkey may perform at most 5 successful registrations per domain in a rolling 24h window (`--max-registrations-per-day`, 0 disables).

- Only registrations that change the held name count: description updates are free, reclaiming a reserved name counts, transfers don't.
- Enforced in Postgres inside the registration transaction, so it holds across instances and concurrent requests, and a refused registration (429) leaves no trace.
- Log rows are pruned hourly once they age out of the window.

### DIFF
--- a/crates/breez-sdk/lnurl/src/routes.rs
+++ b/crates/breez-sdk/lnurl/src/routes.rs
@@ -220,7 +220,7 @@ where
             description,
         };
 
-        if let Err(e) = state.db.upsert_user(&user).await {
+        if let Err(e) = state.db.upsert_user(&user, state.registration_limit).await {
             match e {
                 LnurlRepositoryError::NameTaken => {
                     trace!("name already taken: {}", user.name);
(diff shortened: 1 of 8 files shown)


crates/breez-sdk/lnurl/src/routes.rs
```diff
@@ -361,10 +368,14 @@ where
                 }
                 // A held name has no owner to transfer it, so the source check
                 // inside the transfer answers first and a transfer never
-                // reports a name as reserved. Matched rather than folded into a
-                // catch-all, so a transfer that starts checking holds has to
-                // answer for the status code here.
-                LnurlRepositoryError::NameReserved | LnurlRepositoryError::General(_) => {
+                // reports a name as reserved. The registration limit binds the
+                // register route only, so a transfer never reports it either.
```
roeierez, line 372: Are we sure we don't want to limit transfers as well? One can burn an address with every transfer.

roeierez (changes requested):

roeierez (approved):
LGTM

crates/breez-sdk/lnurl/src/routes.rs
danielgranhao: Transfers doesn't consume new addresses

dangeross (approved):
LGTM

PR #1087, opened 2026-08-28
Fix on-chain sends from a zero balance stable-balance

### LINKED ISSUE #1086: On-chain send from a stable balance wallet fails at prepare with zero sats, and at send with an expired fee quote
## Summary

Two failures stop a Bitcoin address send funded by the USDB balance.

Prepare fetches the coop exit fee quote, and that call reserves leaves even when the amount is `None`, so a wallet holding no sats fails before the conversion is reached.

With a small sats reserve seeded, prepare succeeds, but the quote it captured is consumed only after the AMM conversion has run. When prepare to withdraw outlives the quote, the SSP rejects the withdrawal, after the conversion has already executed.

## Environment

`@breeztech/breez-sdk-spark-react-native` 0.22.0, mainnet, on device, stable balance active with USDB as the active label. Measurements are from device runs. Source references are at tag 0.22.0 and `main` at `1ed3ee0`, and anything read rather than run is marked as such.

## Expected

The stable balance guide states, verbatim, "When you send a bitcoin payment and your sats balance is insufficient, the SDK converts USDB back to bitcoin to cover the payment."

## Reproduce and actual result

**1. Zero sats.** A wallet with 46.71 USDB and 0 sats. `prepareSendPayment` for a Bitcoin address fails with `SdkError.SparkError`, inner `TreeServiceError(NonReservableLeaves)`. Three request shapes fail identically, so the client cannot route around it: a plain send, an explicit `ToBitcoin` in `conversionOptions`, and the token denominated shape (amount in USDB base units, plus `tokenIdentifier`, plus `ToBitcoin`).

**2. Control.** A wallet holding 38,604 sats prepares the same address and amount without error. The failure tracks the empty leaf set, not the address and not the rail.

**3. Reserve seeded.** Prepare succeeds and returns a `conversionEstimate`. `sendPayment` then fails with:

```
Service error: service provider error: graphql error: The coop exit fee quote has expired, please request a new quote.
```

The conversion had already executed. The sats landed in the wallet, the withdrawal never happened, and the stable balance sweeper converted them back to USDB. Nothing was unrecoverable, but the user's dollars round tripped for a payment that did not go out.

Counts are small. One 3,000 sat send completed, one 4,000 sat send failed, and two runs of identical code gave one success and one failure we have not explained. Please read the amounts as context rather than as a distribution.

`SendOnchainFeeQuote.expiresAt` read immediately after prepare came back 59 seconds ahead of the device clock. That is one sample against a client clock, not a server side TTL.

## Root cause

**The quote is captured before the conversion and consumed after it.**

Both Bitcoin address prepare paths call `fetch_coop_exit_fee_quote` and store the result in the prepare response (`prepare/bitcoin_address.rs:130-134` for the sats denominated path, `:209-215` for the token denominated one). Neither path avoids it, since the switch at `:68` requires amount, token identifier and `ToBitcoin` together. On the sats denominated path an active stable balance sets `fee_quote_amount = None` (`:116-129`).

`send/mod.rs:333-350` routes into the conversion pipeline whenever the prepare response carries a `conversion_estimate`. That pipeline executes the AMM leg first, then `send_internal` pulls the stored quote back out (`send/mod.rs:413-414`) and hands it to `bitcoin_address::send`, where `fee_sats` and `fee_quote_id` come from the prepare-time quote with no refetch (`spark-wallet/src/wallet.rs:1448-1466`).

`SendOnchainFeeQuote` carries `expires_at`, and we found no client side check of it on either path. The expiry checks we did find are on the cross-chain quotes (`cross_chain/orchestra.rs`, `cross_chain/boltz.rs`), so the SSP is the first thing to notice.

**A `None` amount still selects leaves.**

Both paths carry a comment describing the opposite of what happens (`prepare/bitcoin_address.rs:209-210`):

```rust
// Pass None for amount — the sats don't exist yet (still tokens),
// so leaf selection would fail. Get a generic fee estimate instead.
```

`None` reaches `select_leaves_with_retry(None)` (`wallet.rs:783-788`), which retrieves all available leaves, and the resulting empty vector is rejected with `TreeServiceError::NonReservableLeaves` (`store.rs:951-953`).

That error never reaches the caller. It is logged (`wallet.rs:2517-2525`), and the insufficient funds early return is gated on `if let Some(target_amounts)`, so it is skipped for `None` (`wallet.rs:2551-2557`). Three attempts run, and the caller gets `Generic("Failed to select leaves after all retries")` (`wallet.rs:2565-2568`).

Read from source rather than measured, the trigger is an empty selectable pool, not a token only wallet specifically. `store.rs:92-96` excludes leaves that are reserved, not `Available`, or missing from operators, so a wallet whose leaves are momentarily all reserved would land in the same place. We have not reproduced that.

Lightning is not affected the same way because `fetch_lightning_send_fee_estimate` performs no leaf selection (`wallet.rs:800-809`). That Lightning prepare succeeds at zero sats is measured.

## Versions

`prepare/bitcoin_address.rs` is the identical blob at 0.22.0, 0.22.1, 0.22.2, 0.22.3, 0.23.0 and `main` at `1ed3ee0`, and `fetch_coop_exit_fee_quote` and `select_leaves_with_retry` are byte identical between 0.22.0 and `main`. 0.23.0 swapped the native tree store to `spark-sqlite`, which keeps the same behavior on this path (all leaves for a `None` target at `spark-sqlite/src/lib.rs:628`, `NonReservableLeaves` on an empty selection at `:1002` and `:1010-1012`). Our device runs were on 0.22.0, and the newer versions were compared by reading source.

## Workarounds attempted

- Re-requesting the quote immediately before `sendPayment`. The conversion still runs after the new quote, so a slow conversion outlives it just the same. This is the run in which the 4,000 sat send failed.
- Buying the sats up front with a separate `ToBitcoin` conversion, then sending a plain sats payment. With stable balance active the SDK re-injects a conversion, because the freshly converted sats have not settled when prepare reads the balance. One attempt produced three conversions.
- Keeping a small permanent sats reserve. This gets past the prepare failure and is how we observed the expiry, but it does not affect the timing, and it forces a wallet configured for dollars to hold sats.

## Possible directions

- Fetching the coop exit quote after the conversion has settled and before `withdraw`, so the clock starts on the sats that will actually be spent.
- Validating `expires_at` on the coop exit quote the way the cross-chain quotes already are, so expiry surfaces as a typed error rather than an SSP GraphQL string.
- Deciding what a quote request with no leaves should mean. The comments at `:116-118` and `:209-210` describe the intended behavior as a generic estimate that skips leaf selection. We did not test whether the SSP accepts an empty `leaf_external_ids` list, so we are not claiming that dropping the local reservation is sufficient.

## Related

#697 covers quote invalidation when the leaf set changes between the quote and the send, and its thread notes that Lightspark may remove the leaf count validation. We are reporting wall clock expiry during a token conversion, on the strength of the SSP message itself, but a conversion also changes the leaf set, and we have not run the control that would separate the two (waiting past the quote lifetime with no conversion, then withdrawing). #1052 touches the same stable balance surface without covering this case.

Happy to test a fix against a branch. We can rerun the same measurements.

dangeross:
A on-chain send funded by USDB and zero sats balance failed at prepare with `NonReservableLeaves`, and once past that, could fail again at send with an expired fee quote after the conversion had already run.

The provider will not quote a withdrawal with no leaves. So prepare now estimates the fee locally when the balance is zero, and quotes for real whenever any leaves exist.

Fixes #1086

### Changes
- `SendOnchainFeeQuote.is_estimate`: New field, set when the wallet holds no bitcoin and a conversion will fund the send. Callers should present those fees as an estimate. 
- `Quote is refreshed before withdrawing`: Quotes live 60 seconds and the prepare-time one was previously carried across the conversion and spent unchecked. `send` now refetches near expiry and never exceeds the fee prepare quoted for the chosen speed. If the rate rose past it, the send steps down to the fastest speed still within that fee rather than overspending, and fails only if even the slowest is out of reach.

### DIFF
--- a/crates/breez-sdk/core/src/sdk/payments/prepare/bitcoin_address.rs
+++ b/crates/breez-sdk/core/src/sdk/payments/prepare/bitcoin_address.rs
@@ -1,6 +1,6 @@
 use crate::{
     BitcoinAddressDetails, ConversionOptions, ConversionType, FeePolicy, SendOnchainFeeQuote,
-    SendPaymentMethod,
+    SendOnchainSpeedFeeQuote, SendPaymentMethod,
     error::SdkError,
     models::{PrepareSendPaymentRequest, PrepareSendPaymentResponse},
     sdk::BreezSdk,
@@ -8,6 +8,23 @@ use crate::{
     token_conversion::ConversionAmount,
     utils::bitcoin_dust::get_dust_limit_sats,
 };
+
+// A wallet with nothing to price against cannot be quoted by the provider, which
+// is the state a stable-balance wallet is in until its token conversion runs.
+// These constants estimate the fee locally for that case.
+
+/// How far the provider's per-speed rates sit above a high-priority rate,
+/// indexed slow, medium, fast. Measured between 1 and 5 sat/vB, where the
+/// network's block targets converge. Under congestion the provider may price
+/// the speeds off targets further apart than this.
+const ESTIMATE_SPEED_PREMIUM_SAT_PER_VBYTE: [u64; 3] = [1, 2, 3];
+/// Headroom on top, for the rate moving while the conversion runs.
+const ESTIMATE_DRIFT_HEADROOM_SAT_PER_VBYTE: u64 = 1;
+/// The provider prices every withdrawal at this size whatever the input count:
+/// the wallet's funds are spent on the Spark side, not as on-chain inputs.
+const ESTIMATE_VSIZE_VBYTES: u64 = 240;
+/// The provider's flat service fee, the same for every speed and amount.
+const ESTIMATE_USER_FEE_SAT: u64 = 750;
 
 /// Validates a Bitcoin address request and returns the validated amount.
 fn validate_request(request: &PrepareSendPaymentRequest) -> Result<u128, SdkError> {
@@ -113,25 +130,37 @@ async fn prepare_sats_denominated(
         )));
     }
 
-    // When a token→sats conversion will run (either auto-filled by an active
-    // stable balance, or explicitly requested via `ToBitcoin` options), the
-    // destination sats don't exist yet — pass None to skip leaf selection.
+    // A wallet short of the amount funds the send by converting tokens, and the
+    // SSP will not price an exit before that conversion has produced the sats,
+    // so bound the fee locally and quote at send instead.
     let stable_balance_active = match &sdk.stable_balance {
         Some(sb) => sb.get_active_label().await.is_some(),
         None => false,
     };
-    let sats_from_conversion =
-        stable_balance_active || conversion::is_to_bitcoin(request.conversion_options.as_ref());
-    let fee_quote_amount = if sats_from_conversion {
-        None
+    let balance_sats = sdk.spark_wallet.get_balance().await?;
+    let conversion_funds_the_send = sats_from_conversion(
+        stable_balance_active,
+        request.conversion_options.as_ref(),
+        balance_sats,
+        amount_u64,
+    );
+    let fee_quote: SendOnchainFeeQuote = if conversion_funds_the_send && balance_sats == 0 {
+        estimate_coop_exit_fee_quote(sdk).await?
     } else {
-        Some(amount.try_into()?)
+        // Short of the amount but holding something, the wallet still has funds
+        // to price against, and the fee does not vary with what is selected.
+        // Asking for the amount when it cannot cover it is what surfaces
+        // insufficient funds on a send with no conversion behind it.
+        let target_sats = if conversion_funds_the_send {
+            None
+        } else {
+            Some(amount_u64)
+        };
+        sdk.spark_wallet
+            .fetch_coop_exit_fee_quote(&withdrawal_address.address, target_sats)
+            .await?
+            .into()
     };
-    let fee_quote: SendOnchainFeeQuote = sdk
-        .spark_wallet
-        .fetch_coop_exit_fee_quote(&withdrawal_address.address, fee_quote_amount)
-        .await?
-        .into();
 
     // For FeesIncluded, validate the output after fees using the best case
     // (slow/lowest fee). Only reject if even the cheapest option results in dust.
@@ -206,17 +235,20 @@ async fn prepare_token_denominated(
         )));
     }
 
-    // Pass None for amount — the sats don't exist yet (still tokens),
-    // so leaf selection would fail. Get a generic fee estimate instead.
-    let fee_quote: SendOnchainFeeQuote = sdk
-        .spark_wallet
-        .fetch_coop_exit_fee_quote(&withdrawal_address.address, None)
-        .await?
-        .into();
+    // Estimate only when the wallet holds nothing to price against. Any balance
+    // at all, as a send-all wallet holding both tokens and sats has, gives a
+    // real quote: the fee does not vary with what is selected.
+    let fee_quote: SendOnchainFeeQuote = if sdk.spark_wallet.get_balance().await? == 0 {
+        estimate_coop_exit_fee_quote(sdk).await?
+    } else {
+        sdk.spark_wallet
+            .fetch_coop_exit_fee_quote(&withdrawal_address.address, None)
+            .await?
+            .into()
+    };
 
     // Token-denominated converts the input into sats; fees come out of the
-    // converted output, which is the FeesIncluded shape — use the slow tier
-    // as the best case for the post-fee dust check.
+    // converted output, which is the FeesIncluded shape.
     validate_dust(
         total_u64,
         dust_limit_sats,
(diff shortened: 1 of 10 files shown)


crates/breez-sdk/core/src/sdk/payments/prepare/bitcoin_address.rs
```diff
@@ -113,25 +130,37 @@ async fn prepare_sats_denominated(
...
-    let fee_quote_amount = if sats_from_conversion {
-        None
+    let balance_sats = sdk.spark_wallet.get_balance().await?;
+    let conversion_funds_the_send = sats_from_conversion(
+        stable_balance_active,
+        request.conversion_options.as_ref(),
+        balance_sats,
+        amount_u64,
+    );
+    let fee_quote: SendOnchainFeeQuote = if conversion_funds_the_send && balance_sats == 0 {
```
roeierez, line 148: Is it only when balance_sats = 0? 
What happen if the user has 1 sat balance for example? Will we be able to get a real quote?

roeierez (commented):
Looks good

crates/breez-sdk/core/src/sdk/payments/prepare/bitcoin_address.rs
dangeross: Yes, still works if you have a 1 sat leaf. Quotes don't vary with available leaves/balance

roeierez (approved):

crates/breez-sdk/core/src/sdk/payments/prepare/bitcoin_address.rs
```diff
@@ -113,25 +130,37 @@ async fn prepare_sats_denominated(
...
+        estimate_coop_exit_fee_quote(sdk).await?
     } else {
-        Some(amount.try_into()?)
+        // Short of the amount but holding something, the wallet still has funds
+        // to price against, and the fee does not vary with what is selected.
+        // Asking for the amount when it cannot cover it is what surfaces
+        // insufficient funds on a send with no conversion behind it.
```
JssDWt: "the fee does not vary with what is selected" - are we sure about that? If that is the case, then why do we have to send leaf ids in order to fetch a coop exit fee quote?

crates/breez-sdk/core/src/sdk/payments/send/bitcoin_address.rs
```diff
@@ -149,6 +162,102 @@ pub(super) async fn send_signed(
...
+            OnchainConfirmationSpeed::Slow,
+        ],
+        OnchainConfirmationSpeed::Medium => &[
+            OnchainConfirmationSpeed::Medium,
+            OnchainConfirmationSpeed::Slow,
+        ],
+        OnchainConfirmationSpeed::Slow => &[OnchainConfirmationSpeed::Slow],
+    };
+
+    for (stepped_down, candidate) in candidates.iter().enumerate() {
```
JssDWt: This is a nit, but stepped_down reads like a boolean, while it's actually the index of the slice. As a style rule I'd say let's be explicit about the fact it's an index first, then initialize a boolean based on the index later.

docs/breez-sdk/src/guide/send_payment.md
```diff
@@ -36,6 +36,10 @@ If the invoice also contains a Spark address, the payment can be sent directly v
 
 For Bitcoin addresses, the amount must be set in the request. The prepare response includes fee quotes for three payment speeds: Slow, Medium, and Fast.
 
+The quote's {{#name is_estimate}} flag is set when your bitcoin balance is zero and a token conversion will fund the payment, which happens automatically when [Stable Balance](./stable_balance.md) is active. With no bitcoin to price the withdrawal against, the fees shown are derived from current on-chain rates and the real ones may differ. They are an upper bound: the SDK fetches a real quote once the conversion completes, and fails rather than spending more than it estimated. Present this as an estimate if you show it.
```
JssDWt: "With no bitcoin to price the withdrawal against" is a little bit cryptic. I don't think the user needs to know the details here.

docs/breez-sdk/src/guide/stable_balance.md
```diff
@@ -65,6 +65,8 @@ When you [prepare to send a payment](./send_payment.md#preparing-payments) witho
 1. If you have enough bitcoin balance, no conversion is needed
 2. If your bitcoin balance is insufficient, the SDK configures conversion options using your Stable Balance settings (token identifier and slippage)
 
+For [bitcoin address](./send_payment.md#bitcoin) destinations the prepare response marks its fee quote {{#name is_estimate}} when your bitcoin balance is zero, leaving nothing to price the withdrawal against. Any balance at all still gets a real quote. An estimate is always replaced with a real one before the payment goes out.
```
JssDWt: Same "leaving nothing to price the withdrawal against", I think is a bit of a cryptic implementation detail for users.

JssDWt (commented):
Looks good I think. We should probably consult with Lightspark whether the assumptions made here are correct?

crates/breez-sdk/core/src/sdk/payments/prepare/bitcoin_address.rs
dangeross: I tested it across different leaf counts of 1, 2, 5, 7 and 22 (all leaves), amounts from 1 to 37,796 sats, and both P2WPKH and P2TR destinations. User fee was always 750 sats, L1 broadcast fee was always 240 × rate.

danielgranhao (approved):
LGTM

PR #1090, opened 2026-08-31
fix: upgrade deadpool to 0.13

erdemyerebasmaz:
**Signer compatibility tests** job started to fail after a transitive dependency update pulled in an incompatible version of `deadpool`, causing a type mismatch. 

This PR aligns the dependencies.

<details>

`deadpool-postgres` `0.14.2` (published _2026-08-26_) moved its own dependency
from `deadpool` `0.12` to `deadpool` `0.13`, in a patch release. It re-exports
`deadpool` types in its public API (`PoolConfig`, `Timeouts`, `QueueMode`), so
`spark-postgres`, which names those types directly, ended up with two
incompatible `deadpool` versions in one graph.

The root workspace was shielded by `Cargo.lock`, but
`crates/spark-compat-itest` is a standalone workspace with a gitignored
lock, so CI resolved it fresh and picked up `0.14.2`. The **Signer
compatibility tests** job has failed on every PR since.

`deadpool` `0.13.0` only bumped `deadpool-runtime` and dropped a
`SpawnBlockingError` re-export we never used, so no source changes are
needed. `Runtime::Tokio1`, `PoolConfig`, `Timeouts` and `QueueMode` are
unchanged.

The lock update carries `deadpool-postgres` to `0.14.2` and, with it,
`tokio-postgres` `0.7.13` -> `0.7.18`.
</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2338,25 +2338,23 @@ checksum = "a4ae5f15dda3c708c0ade84bfee31ccab44a3da4f88015ed22f63732abe300c8"
 
 [[package]]
 name = "deadpool"
-version = "0.12.3"
+version = "0.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0be2b1d1d6ec8d846f05e137292d0b89133caf95ef33695424c09568bdd39b1b"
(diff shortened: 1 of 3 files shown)


roeierez (approved):

JssDWt:
Looks good, needs a rebase.

PR #1091, opened 2026-08-31
log missing htlc_amount

JssDWt:
A significant amount of webhooks has a null amount_sat. Log the payload coming from spark, to see whether the amount is missing from there.


roeierez (approved):
LGTM

PR #1092, opened 2026-09-01
Add optional nostr_zap_request in the lnurl webhook payload

yaslama:


crates/breez-sdk/lnurl/src/webhook_notify.rs
```diff
@@ -20,6 +20,8 @@ pub enum WebhookPayload {
         lightning_address: Option<String>,
         sender_comment: Option<String>,
         timestamp: i64,
+        #[serde(skip_serializing_if = "Option::is_none")]
```
roeierez: This is different than other optional fields which don't have the serde attribute. Is it on purpose? It means that on a non zap payment the field will be absent (rather than explicit null).

roeierez (commented):
Looks good. One comment.

crates/breez-sdk/lnurl/src/webhook_notify.rs
yaslama: Yes. I thought that it's better because zap payments are special. But I can remove it if you prefer

JssDWt: nit: I think this line can be removed. Docs also say `string | null`

JssDWt (commented):
Looks good

roeierez (approved):
LGTM

PR #1095, opened 2026-09-03
Fix resuming a unilateral exit

JssDWt:
- **`prepare_unilateral_exit` reads the chain** and prices only the work an exit has
  left. It was pricing a fresh exit every time.
- **`unilateral_exit` no longer walks the tree.** It reads only the funding: each
  supplied outpoint is followed to whatever it became, and the planner assigns the
  result to branches. Funding used to be tracked through the build by three separate
  mechanisms that did not know about each other.
- **New `check_unilateral_exit`.** Hand it the response you stored and it hands the
  same exit back with statuses up to date, plus a verdict: `Valid`, `Done`, or
  `Redo`. Needs no wallet, leaves, signer or funding, so an exit can be followed from
  that response alone.
- **The caller owns the money.** A funding UTXO an earlier attempt spent is followed,
  not refused, so you can hand back the `funding_inputs` the exit came with.
- **Import now checks leaf ownership**, and a leaf part-way out is no longer
  re-verified or offered for renewal on every refresh.

## Fixes

- A branch could be funded from another branch's change, then build over an output
  that was already gone. The package was rejected for as long as the exit ran and the
  branch never finished.
- A fan-out's outputs were matched to branches by position, and that position list
  followed what was left to do, so a branch settling shifted every branch after it
  onto outputs that were never theirs.
- `Auto` could drop a profitable leaf where some parents confirmed, because it assumed the parents were unconfirmed.
- An imported leaf was never ownership-checked, and never would be.

## API

New: `check_unilateral_exit`, and the types it returns.

Breaking: **`SdkError::FundingUtxoConflict` is removed.** It can no longer occur, and
`InsufficientCpfpFunds` covers what is left of it.

New fields, additive in Rust but required in the bindings:

| Type | Field |
|---|---|
| `PrepareUnilateralExitResponse` | `exit_chain_state` |
| `UnilateralExitResponse` | `funding_inputs` |
| `UnilateralExitTransaction` | `dependencies_met` |

`dependencies_met` does not cover `csv_timelock_blocks`, because
`BitcoinChainService` has no tip height. Maturity stays the caller's to check, which
is why it is not called `ready_to_broadcast`.


crates/breez-sdk/core/src/sdk/unilateral_exit.rs
```diff
@@ -567,6 +848,14 @@ async fn resolve_exit_observations(
 /// rather than treating it as confirmed or absent.
 async fn execute_chain_query(chain: &dyn BitcoinChainService, query: &ChainQuery) -> ChainResult {
     match query {
+        ChainQuery::TxConfirmed(txid) => match chain.get_transaction_status(txid.to_string()).await
```
roeierez, line 996: `TxStatus` also carries `block_height`, dropped here. This will help a lot callers as a CSV
timelock counts from a height, so it re-fetches every one of these itself.
Could `ConfirmationStatus::Confirmed` carry the height?

crates/breez-sdk/core/src/models/mod.rs
```diff
@@ -2883,6 +2883,10 @@ pub struct UnilateralExitTransaction {
     /// Txids of other entries in this list that must be confirmed before this
     /// one can be broadcast.
     pub depends_on: Vec<String>,
+    /// Whether every transaction in `depends_on` is confirmed. A `csv_timelock_blocks`
+    /// that has yet to mature still holds this one back, which is yours to check
+    /// against the chain tip.
+    pub dependencies_met: bool,
```
roeierez: I am thinking it would be great to change this field to `ready_to_broadcast` and callers would drop their own timelock calculation.

roeierez (commented):
API looks great. I just added two comments that may help a lot the implementors.

crates/breez-sdk/core/src/models/mod.rs
roeierez, line 3025: `total_fee_sat` mixes fees the CPFP children pay from the caller's funding with fees taken out of the leaves (refund, sweep), so a wallet cannot derive what actually reaches the destination.

Perhaps we can add to the response:
 * `max_refunds_fee` (the direct refunds' self-paid fees, the worst case when the watchtower publishes first)
* `sweep_fee`
* `min_deliverable_amount`  (leaves value - max_refunds_fee - sweep_fee)

Then an app can show a floor estimated amount that will land on the destination address.

roeierez (commented):
Looks good!
Almost working in glow.

crates/breez-sdk/core/src/sdk/unilateral_exit.rs
```diff
@@ -526,39 +611,356 @@ fn parse_xonly(pubkey: &str) -> Result<XOnlyPublicKey, SdkError> {
...
+    for (tx, met) in transactions.iter().zip(&met) {
+        // On-chain, or its own on-chain state unknown: neither leaves anything
+        // to wait for.
+        if matches!(
+            tx.status,
+            ExitTransactionStatus::Confirmed { .. } | ExitTransactionStatus::Unverified
+        ) {
+            resolved.push(tx.status);
+            continue;
+        }
```
dangeross, line 822: Looking here it seems that `Unverified` can only ever go to `Confirmed`, never `Ready`? Nothing writes a status after this, and the update loop in `check_unilateral_exit` only writes `Confirmed`. So a transaction left `Unverified` by a chain service issue in `prepare_unilateral_exit` would stay that way for the life of the exit and never get broadcast. Should `check_unilateral_exit` be re-checking it?

crates/spark-wallet/src/unilateral_exit.rs
```diff
@@ -1030,34 +1307,9 @@ pub(crate) fn build_exit(
...
-                                funding = combined;
-                            }
-                            None
-                        }
+                        // Confirmed: its child already paid, and what that child
+                        // left over came back as funding like any other input, so
+                        // the branch keeps what it was assigned.
                         Some(_) => None,
                         None => {
                             let child =
```
dangeross, line 1296: I'm wondering if the fan-out dependency is going on the right transaction. `first_in_branch` a few lines up puts it on the first node the branch emits, but the one that actually spends the fan-out output is the first to take the `None` arm here. It's the same node on a fresh exit. On a resume the confirmed ancestors are emitted first with `to_sign: None`, so I think it lands on one of those instead.

crates/spark-wallet/src/unilateral_exit.rs
```diff
@@ -213,6 +199,11 @@ pub struct CpfpChangeInput {
 pub enum ChainQuery {
     /// Is this output spent, and by which (confirmed?) transaction?
     Outspend(OutPoint),
+    /// Is this transaction in a block? The direct question, for a transaction we
+    /// already hold: it says nothing about what took its place if it is not, so
+    /// ask an [`ChainQuery::Outspend`] of one of its inputs only when the answer
+    /// matters.
```
dangeross: nit: Very cryptic comment

crates/spark-wallet/src/unilateral_exit.rs
```diff
@@ -306,162 +305,528 @@ impl<'a> ObservedIndex<'a> {
...
-                }
-            }
-        }
-    }
+    walk
+}
+
+/// The exit's on-chain state as the build sees it: the tree, read before the
+/// exit was funded, restated in the build's own terms. Pure, and asks the chain
+/// nothing: the funding was resolved before the plan was made.
```
dangeross: nit: I think it could be simplified and be more informational

crates/spark/src/tree/service.rs
```diff
@@ -68,6 +68,22 @@ struct RenewalOutcome {
...
+/// Splits a refresh's leaves into the ones worth renewing and the ones already on
+/// their way out. Only a spendable leaf is renewed: the statuses the exit added
+/// are either part-way out or already locked by the operators, and asking for a
+/// new timelock on one of those is at best refused, while a node that no longer
+/// carries a refund fails the check outright and would take the whole refresh
+/// down with it.
+///
+/// Named as what is kept rather than what is excluded, so a status added later is
+/// left out until someone decides it belongs. Both halves are stored: a leaf on
+/// its way out keeps its place in the pool, and its chain with it.
```
dangeross: nit: It this paragraph needed?

docs/breez-sdk/src/guide/unilateral_exit.md
```diff
@@ -18,10 +18,13 @@ Three things are important to know before you build an exit:
 
 Your balance is held in a tree of pre-signed Bitcoin transactions. Each leaf is a portion of the balance. To move a leaf on-chain you broadcast the chain of transactions from the tree down to that leaf, then a refund transaction, then a final sweep to your destination address. Because the pre-signed transactions pay no fee on their own, each one is broadcast together with a CPFP child that pays its fee.
 
-The exit is two calls:
+The exit is three calls, and you will use all three:
```
dangeross: Don't think its needed
```suggestion
The exit is three calls:
```

docs/breez-sdk/src/guide/unilateral_exit.md
```diff
@@ -49,12 +52,40 @@ Call {{#name prepare_unilateral_exit}} with the target {{#name fee_rate_sat_per_
...
+| Component | Paid by |
+|---|---|
+| {{#name cpfp_fee_sat}} | The funding UTXOs, through the CPFP children that fee-bump the tree transactions |
+| {{#name fanout_fee_sat}} | The funding UTXO, by the fan-out transaction. Zero when there is no fan-out |
+| {{#name sweep_fee_sat}} | The value being recovered, by the final sweep |
+
+The three always add up to the total:
+
+```text
+cpfp_fee_sat + fanout_fee_sat + sweep_fee_sat = total_fee_sat
```
dangeross: Not sure if it's possible to change, but these won't align with the table above if a different language is selected. Same for the two other fragments below

docs/breez-sdk/src/guide/unilateral_exit.md
```diff
@@ -49,12 +52,40 @@ Call {{#name prepare_unilateral_exit}} with the target {{#name fee_rate_sat_per_
...
+```
+
+The first two come out of the Bitcoin you supplied as funding and do not reduce what the exit recovers. The third is different: the sweep spends the refunds and pays out what is left after its own fee, so it comes off the money on its way to your address.
+
+**What arrives at {{#name destination}}** is therefore {{#name recoverable_value_sat}} less {{#name sweep_fee_sat}}, plus any funding that was not spent on fees. The sweep also collects the leftover change of the CPFP children it built, so unused funding is delivered to the same address rather than left behind.
+
+**What the exit costs in total** is {{#name total_fee_sat}}, counting both pockets. Beginning with {{#name recoverable_value_sat}} in Spark and a funding UTXO of `F`, the destination ends up with `recoverable_value_sat + F - total_fee_sat`.
```
dangeross: Not sure if `pockets` is clear

dangeross (commented):
Looks good!

crates/breez-sdk/core/src/sdk/unilateral_exit.rs
JssDWt: I think I solved this ok with a `status_after_check` function.

crates/spark-wallet/src/unilateral_exit.rs
JssDWt: It wasn't. Fixed.

crates/breez-sdk/cli/src/command/advanced.rs
```diff
@@ -208,9 +212,26 @@ fn print_exit_transactions(response: &UnilateralExitResponse) {
...
+                println!("      (waiting on the transactions it depends on)");
+            }
+            ExitTransactionStatus::WaitingForTimelock {
+                spendable_at_height,
+            } => match spendable_at_height {
+                Some(height) => println!("      (waiting for its timelock, until block {height})"),
+                None => println!("      (waiting for its timelock)"),
+            },
+            ExitTransactionStatus::Ready | ExitTransactionStatus::Unverified => {}
         }
```
dangeross, line 235: Is it worth adding a `CheckUnilateralExit` command also? Perhaps adapting `UnilateralExit` to write the response to `--output-file` and read with `--input-file`

roeierez (approved):
LGTM

dangeross (approved):
LGTM

crates/spark/src/tree/service.rs
```diff
@@ -1077,8 +1121,18 @@ fn bare_pedigrees(leaves: Vec<TreeNode>) -> Vec<LeafPedigree> {
...
+/// spendable one. `OnChain` and `Exited` are the two an exit drives a leaf
+/// through, and they have to be asked for: a refresh that requested only
+/// `Available` lost the leaf from every operator's response at once the moment
+/// its node txs confirmed, and the store, reading that absence as a spend,
+/// deleted the chain the exit still needed to resume.
+fn held_leaf_statuses() -> Vec<i32> {
+    vec![
+        ProtoTreeNodeStatus::Available as i32,
+        ProtoTreeNodeStatus::OnChain as i32,
+        ProtoTreeNodeStatus::Exited as i32,
```
danielgranhao, line 1134: `ParentExited` isn't in here. It's mapped from the proto and handled downstream (`partition_renewable` keeps it, `selects_leaf_in_any_status` calls it exitable), but since we never ask for it, `set_leaves` drops any leaf in that state along with its chain, and exit is the only way out of it by then. Deliberate?

crates/spark/src/tree/service.rs
```diff
@@ -171,6 +186,18 @@ impl TreeService for SynchronousTreeService {
...
+        // The only leaves whose data comes from outside the SDK, so the only ones
+        // whose ownership nothing else has established. A refresh will not do it
+        // later either: it skips a leaf whose stored keys match the ones reported,
+        // which an unchecked import satisfies by definition.
+        let owned = self
+            .retain_owned(&leaves.iter().map(|p| &p.leaf).collect::<Vec<_>>())
+            .await?;
+        let leaves: Vec<LeafPedigree> = leaves
+            .iter()
+            .filter(|p| owned.contains(&p.leaf.id))
```
danielgranhao, line 198: This drops the leaf silently, but `import_exit_state` has already counted it in `imported_leaves`, and its own filter is only the `owner_identity_public_key` field, not this derived check. So `ExitStateImport` can report a leaf written when nothing was. Return the count from here?

crates/spark/src/services/unilateral_exit.rs
```diff
@@ -131,7 +222,15 @@ pub fn plan_unilateral_exit(
         });
     }
 
-    let (per_branch_funding, fan_out_psbt) = if selected.len() == 1 {
+    let to_build: Vec<UnilateralExitSelectedLeaf> = selected
+        .iter()
+        .filter(|leaf| !is_complete_on_chain(on_chain, &tree_nodes, &leaf.id))
```
danielgranhao, line 227: Worth dropping `on_chain.stopped_leaves` here too? `build_exit` drives nothing for them, so each still takes a fan-out output and a branch's worth of funding it never spends. Their value also stays in `selected_leaves`, so `recoverable_value_sat` counts them even when no refund surfaced.

crates/spark/src/services/unilateral_exit.rs
```diff
@@ -538,8 +685,11 @@ pub fn evaluate_unilateral_exit_leaf_costs(
             if covered_txids.contains(&txid) {
                 continue;
             }
-            // On-chain ancestor is already confirmed, so its CPFP fee is already paid.
-            if ancestor.status == TreeNodeStatus::OnChain {
+            // A node already on-chain is not built again, so it is not counted at
+            // all rather than counted and subtracted: the first node that is
+            // still to be built is the one the funding pays for directly, and is
+            // costed on the weight of that funding.
+            if on_chain.has_node(&ancestor.id) || ancestor.status == TreeNodeStatus::OnChain {
```
danielgranhao, line 692: The walk deliberately doesn't trust this label (`resume_point`: "The label is only where to look"), but the cost pass still does. When the chain says the node_tx isn't in a block, it stays out of `state.nodes` and `build_exit` builds a child for it anyway, so the quote comes up a package short. `has_node` on its own?

danielgranhao (approved):
My previous comments aren't blockers

PR #1098, opened 2026-09-07
Keep a leaf until its spend is proven

JssDWt:
Stacked on #1095. Replaces #1028.

If a leaf was missing from the operators' response during a refresh, we would
take that as proof the leaf had been spent, and delete it along with its chain.
One operator going quiet, or a bug leaving a leaf out of a response, would be
enough to lose the only way of recovering that leaf. Cancelling a reservation
would drop every leaf a single operator's response did not include. A failed
query could return an empty response, so unreachable operators would take a
whole reservation's chains.

A leaf missing from the response is now marked, not deleted. It stops counting
towards the balance and cannot be spent, but its chain stays and it is still
exitable. The leaf is deleted only once our own transfer history shows we sent
it on, under a refund we signed. A new background task looks for that at startup
and then hourly, and an exit repeats the check before it picks the leaves to
quote.

Reading the leaf back from the operators would not work instead: they return a
leaf only to a wallet with read access to its owner, and the SDK enables private
mode for every wallet by default, so a leaf we have sent on is not ours to read.

Adds an `is_deleted` column to the SQL stores.


crates/breez-sdk/wasm/js/postgres-tree-store/index.cjs
```diff
@@ -524,33 +569,34 @@ class PostgresTreeStore {
...
           await client.query(
             "DELETE FROM brz_tree_ancestors WHERE user_id = $1 AND leaf_id = ANY($2)",
-            [this.identity, goneIds]
+            [this.identity, spentList]
+          );
+          await client.query(
+            "DELETE FROM brz_tree_leaves WHERE user_id = $1 AND reservation_id IS NULL AND id = ANY($2)",
+            [this.identity, spentList]
           );
         }
```
dangeross, line 595: This diverges from the other backends by deleting the ancestors even when the leaf row is kept because it's reserved. That leaf ends up with a row and no chain.

crates/spark-wallet/src/wallet.rs
```diff
@@ -2887,6 +2934,121 @@ async fn claim_transfer(
...
+) -> Result<usize, SparkWalletError> {
+    let kept = tree_service.list_leaves_kept_for_exit().await?;
+    if kept.is_empty() {
+        return Ok(0);
+    }
+    let kept_by_id: HashMap<&TreeNodeId, &TreeNode> =
+        kept.iter().map(|leaf| (&leaf.id, leaf)).collect();
+
+    let bitcoin_service = BitcoinService::new(network);
+    let mut proven: Vec<TreeNodeId> = Vec::new();
```
dangeross, line 2959: nit: could `proven` be a `HashSet`? Paging is offset based over a set the coordinator appends to, so the same transfer can come back on two pages and prove the same leaf twice. That inflates the count against `kept.len()` below and stops the walk a page early. Same hazard `reporting` guards against in `refresh_leaves`.

dangeross (commented):
Looking good

PR #1100, opened 2026-09-08
fix(deposits): key the claim guard on the parsed txid

danielgranhao:
The in-flight claim guard keys on `request.txid`, the caller's raw string, while
the background claim loop keys on the parsed `detailed_utxo.txid.to_string()`.
`Txid::from_str` accepts hex in any case, so an uppercase txid inserts a key the
sync loop never looks up: the claim runs normally and the mutual exclusion
silently does not apply.

`claim_deposit` now builds the guard key and the `DepositClaimInProgress` error
from `detailed_utxo`, matching the key the sync loop uses. The error's `tx` field
is consequently canonical lowercase.


dangeross (approved):
LGTM

roeierez (approved):

PR #1101, opened 2026-09-08
Claim unconfirmed on-chain deposits

### LINKED ISSUE #1041: Detect incoming on-chain deposits in the mempool
Deposits are discovered only via the Spark operator's `get_utxos_for_identity`, which exposes them from 1 confirmation. A deposit stays invisible for ~10 min+ after broadcast, even though we already model a not-yet-claimable deposit (#749, closing #453). We want it surfaced as soon as it's in the mempool. Claiming still waits for 3 confs, so no API change is needed.

Options:
1. Lightspark tracks the mempool and returns it from get_utxos_for_identity. No client work, fixes it for all Spark SDKs. TBD if feasible or what's the timeline.
2. SDK tracks it via the existing `BitcoinChainService` over the addresses from `list_static_deposit_addresses`. Costs: one REST call per rotated address per sync against a public Esplora, plus handling RBF/dropped txs and reconcile_deposits (which today deletes anything absent from the operator's set).

Option 1 is preferred if feasible. 

Getting this will also allow true 0-conf claims in https://github.com/breez/spark-sdk/pull/1012. The 0-conf feature there at least currently only allows claiming after 1 conf, as that is when the deposit is detected.

dangeross:
The instant-claim path already supports 0-conf claims end to end, but it could almost never fire. The Spark operators do not report a deposit UTXO until it has a confirmation, so the SDK never learned about the deposit in time. This adds a chain-side source, where the SDK polls the deposit addresses and finds the deposit while it is still in the mempool.

Requesting a receive address starts a 24-hour window on it, and requesting it again restarts that, so the watch tracks how recently you asked to be paid.

Two rules retire an address:
- **Deposit state**: once its deposits have confirmed, the address is dropped. The address you are currently receiving on is exempt from this, because a static deposit address takes repeat payments.
- **The window**: applies to every address including the current one. This prevents an address from being polled forever. An address holding a still-unconfirmed deposit is followed past its window until that deposit clears.

Claims are also now idempotent against the operator feed, which keeps reporting a UTXO the provider has already fronted: a deposit whose claim is settling or already credited is skipped by both claim paths, and an "already claimed" rejection counts as settled rather than failed. That skip is terminal, so a claim the provider accepts but never settles is recovered with an explicit `claim_deposit`, which ignores the status.

Nothing deletes a deposit record on credit any more. It is marked with a new `InstantClaimStatus::Claimed` and reconciliation removes it once the provider spends the output: deleting it earlier let both sources re-announce and re-claim the deposit. A claimed deposit therefore stays in `list_unclaimed_deposits` for a while after the amount arrives, so apps should branch on the status rather than treat every listed deposit as awaiting action. The new variant is a breaking change for exhaustive matches in Kotlin, Swift and TypeScript.

`Storage` gains two methods: `list_watched_deposit_addresses` and `update_watched_deposit_address`, backed by a new `watched_deposit_addresses` table across all seven backends.

Closes #1041


crates/breez-sdk/core/src/realtime_sync/storage.rs
```diff
@@ -605,6 +605,24 @@ impl Storage for SyncedStorage {
...
+    async fn update_watched_deposit_address(
+        &self,
+        address: String,
+        payload: crate::persist::UpdateWatchedAddressPayload,
+    ) -> Result<(), StorageError> {
+        self.inner
+            .update_watched_deposit_address(address, payload)
+            .await
+    }
+
```
roeierez, line 625: Should we consider using the get/set cached item already existing function instead of introducing new storage functions?

crates/breez-sdk/core/src/sdk/deposits.rs
```diff
@@ -121,6 +117,19 @@ impl BreezSdk {
                     payment: Some(payment),
                 })
             }
+            // An earlier claim took it, so the deposit is settled and its row
+            // goes. The caller still gets the error: the claim they asked for
+            // did not happen here.
+            Err(e) if is_already_claimed_error(&e.to_string()) => {
```
roeierez, line 123: claim_deposit never checks the `Submitted` flag, so it can send a second claim for a deposit the sync already claimed. The deposit is still in `list_unclaimed_deposits`, so apps show a Claim button for it.

Worse, line 634 always overwrites the status. At 0 conf `select_instant_claim_plan` returns NoPlan → Declined, which erases Submitted { claim_id }. After that the sync claims again, and `can_drop_unobserved_deposit` may delete the row while the first claim is still running.

crates/breez-sdk/core/src/sdk/sync.rs
```diff
@@ -394,6 +402,13 @@ impl BreezSdk {
             let Some(_claim_guard) = self.claim_guards.try_acquire(key.clone()) else {
                 continue;
             };
+            // An early claim has already gone out. The provider credits it
+            // asynchronously and keeps reporting the UTXO, eventually as mature,
+            // so without this the claim at maturity re-submits every pass and is
+            // rejected every time.
+            if claim_already_submitted(instant_status.get(&key)) {
```
roeierez, line 409: `instant_status` is read once on line 368, before the loop. Each deposit here makes several network calls, so the copy goes stale.

A `claim_deposit` that writes Submitted and releases its guard mid-loop passes both checks: `try_acquire` on line 402 succeeds, and this line reads the old copy.

Seems like a race for Submitted deposits?

crates/breez-sdk/core/src/sdk/sync.rs
```diff
@@ -39,6 +40,13 @@ fn instant_claim_worth_attempting(
     }
 }
 
+/// Whether an early claim has already gone out for this deposit. The provider
+/// credits it asynchronously and keeps reporting the UTXO, eventually as mature,
+/// so both claim paths have to leave such a deposit alone.
```
roeierez, line 45: The comment says "both claim paths", but only the `sync` uses this. `claim_deposit` never calls it.

crates/breez-sdk/core/src/sdk/sync.rs
```diff
@@ -39,6 +40,13 @@ fn instant_claim_worth_attempting(
...
+fn claim_already_submitted(status: Option<&InstantClaimStatus>) -> bool {
+    matches!(status, Some(InstantClaimStatus::Submitted { .. }))
```
roeierez, line 47: I am thinking also on recovering here from stale Submitted... 
If the ssp doesn't continue with the deposit how do we recover? Or the submitted status will stay forever?

crates/breez-sdk/core/src/utils/deposit_chain_syncer.rs
```diff
@@ -170,6 +203,43 @@ impl DepositChainSyncer {
...
+        }
+        match self
+            .chain_service
+            .get_outspend(deposit.txid.clone(), deposit.vout)
+            .await
+        {
+            Ok(Outspend::Spent { status, .. }) if status.confirmed => true,
+            Ok(_) => false,
+            Err(e) => {
+                warn!(
```
roeierez, line 234: Worth check if this error is 404 from the chain service? since zero conf transactions might disappear? If we don't do that looks like we will check this deposit forever?

roeierez (changes requested):

crates/breez-sdk/core/src/realtime_sync/storage.rs
dangeross: I considered it, but thought because it would be regularly used for reads/writes during on-chain deposits/syncing it's worth the storage trait fns. Also because it's easier to prevent concurrent writes.

crates/breez-sdk/core/src/utils/deposit_chain_syncer.rs
```diff
@@ -293,6 +363,181 @@ impl DepositChainSyncer {
...
+
+    let mut observations = HashMap::new();
+    let mut complete = true;
+    'watch_loop: for address in plan.poll {
+        let utxos = match chain_service.get_address_utxos(address.clone()).await {
+            Ok(utxos) => utxos,
+            Err(e) => {
+                warn!("Failed to read the UTXOs of watched address {address}: {e}");
+                complete = false;
+                break;
```
danielgranhao, line 452: maybe `continue` instead of `break`? Otherwise a failure can prevent other addresses from being read

crates/breez-sdk/core/src/sdk/deposits.rs
dangeross: The missing check is intentional, since `claim_deposit` is the recovery path from a stale claim, and the status is on `DepositInfo` for apps to branch on rather than offer a claim.

Fixed the overwrite.

crates/breez-sdk/core/src/sdk/sync.rs
dangeross: The status clears once reconciliation sees the outpoint spent, so it stays Submitted only if the provider never sweeps. Calling `claim_deposit` is the recovery path. It intentionally doesn't check the Submitted status to allow the claim to pass through to the SSP.

roeierez (approved):
LGTM

danielgranhao (approved):

PR #1102, opened 2026-09-08
feat(cross-chain): expose the Orchestra settlement tx hash

### LINKED ISSUE #1085: USDT/USDC settlement chain transaction hash
When moving from bitcoin to USDT/USDC, you usually get a spark transaction hash, but it would be helpful if we also got the settlement chain transaction hash for payment proof or anything one wants to use it for.

danielgranhao:
Closes #1085.

Adds `destination_tx_hash` to `ConversionInfo::Orchestra`, set when the order
reaches a terminal state. It identifies the settlement transaction on the
destination chain.

Orchestra returns it as `order.destinationTxHash` on `/v1/orchestration/status`,
declared on `OperationRecord` as nullable and not required, so it is modelled as
optional.

No storage migration: `conversion_info` is a JSON blob and the field defaults to
unset on existing rows. Boltz is unchanged and leaves it unset.

Testing: the mainnet cross-chain itest now requires a hash on Orchestra
completion and verifies it against the destination chain (receipt succeeded,
credits the recipient the delivered amount). Verified on mainnet: SDK
`delivered_amount`, on-chain balance delta, and the settlement tx credit all
agree at 1009285 USDC.


crates/flashnet/src/orchestra/models.rs
```diff
@@ -259,6 +259,11 @@ pub struct Order {
     pub deposit_address: String,
     pub destination_chain: String,
     pub destination_asset: String,
+    /// Settlement transaction on the destination chain, set once the order is
+    /// delivered. Absent from the documented status schema but returned by the
+    /// live API, so it is modelled as optional.
```
dangeross: FYI it's in the OpenAPI docs: https://orchestration.flashnet.xyz/docs

dangeross (approved):
LGTM

crates/flashnet/src/orchestra/models.rs
danielgranhao: Thanks, adjusted rustdocs

crates/breez-sdk/core/src/token_conversion/models.rs
```diff
@@ -236,6 +236,17 @@ pub enum ConversionInfo {
...
+        /// delivery on that chain's block explorer. Format follows the chain
+        /// (e.g. `0x`-prefixed hex on EVM, a base58 signature on Solana).
+        /// Unset until the order is delivered, and on orders that failed or
+        /// were refunded.
+        ///
+        /// On a receive the destination is Spark, so this is the Spark transfer
+        /// id (or token tx hash) of the inbound payment, not an external-chain
+        /// transaction. Note that `chain` names the source side on a receive.
+        #[serde(default)]
+        destination_tx_hash: Option<String>,
```
dangeross: I'm wondering if this should always be the external tx hash. Looking at the issue, they are looking for payment proof on the external chain, but this doesn't provide that for receive. Wdyt? Should we consider a more generic field name and using the source_tx_hash from the Orchestra order status for receive?

roeierez (approved):

dangeross (approved):

PR #1103, opened 2026-09-08
Add missing proxy field to itest config

dangeross:
`cargo clippy -p breez-sdk-itest --all-features` did not build: the itest Turnkey config was missing the `proxy` field added in #1076. Nothing in CI compiled that file, because `cargo xtask clippy` only ever built default features plus a one-off `local-itest` pass.

### Fix
- Add the missing `proxy: None` to the itest Turnkey config.
- Replace the `local-itest` special case with `FEATURE_CLIPPY_PASSES`: extra clippy passes for `breez-sdk-itest`, `cli`, `breez-sdk-bindings` (all features) and `breez-sdk-spark` (its native features; `--all-features` there needs `uniffi/tokio`, which only the bindings crate turns on). This picks up `turnkey`, `fido2`, `uniffi-cli` and `span-trace`, none of which were built before.
- Guard against the next gap: `cargo xtask clippy` now aborts when a workspace package declares a feature no pass builds, naming it and pointing at `FEATURE_CLIPPY_PASSES` or `UNLINTED_FEATURES` (features another target already compiles, namely `browser-tests` and `test-arbitrary-precision`).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -29,6 +29,11 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
+      # The CLI's `fido2` feature pulls in hidapi, whose build script needs
+      # libudev on Linux.
+      - name: Install libudev
+        run: sudo apt-get update && sudo apt-get install -y libudev-dev
+
       - name: Run Clippy
         run: make cargo-clippy-check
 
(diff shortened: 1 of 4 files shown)


danielgranhao:
Looks good. Would you mind adjusting CI/xtask to prevent this from happening in the future?

roeierez (approved):

danielgranhao:
Nice, but now clippy job is failing

danielgranhao (approved):

PR #1105, opened 2026-09-08
fix(cross-chain): unwire the Boltz provider

### LINKED ISSUE #1081: Boltz provider unusable on mainnet: 'BTC/TBTC pair not found. Is referral header configured?'
## What happens

On a mainnet wallet (`defaultConfig('mainnet')`, SDK 0.22.0/0.23.0, WASM in a WebView), every cross-chain prepare that falls back to Boltz fails:

```
[SendSheet] boltz quote failed, trying next provider: Error: Network error: Boltz API: BTC/TBTC pair not found. Is referral header configured?
```

Orchestra quotes on the same route work, so the Boltz leg is dead as a fallback provider.

## Where it comes from

`boltz-client` `crates/lib/src/swap/reverse.rs` looks up `pairs["BTC"]["TBTC"]` and errors with exactly this message when absent - the string itself anticipates the cause: the pair appears to be partner-gated, served by Boltz only when the referral header is recognized. `referral_id` is sent on every request per `config.rs`, so either the shipped referral is not whitelisted for this pair, or the header is not reaching Boltz in this build.

## Impact

No coverage loss where Orchestra overlaps, but: (1) no provider redundancy - an Orchestra outage takes cross-chain down entirely; (2) Boltz-exclusive destinations can never quote; (3) every cross-chain prepare pays a doomed Boltz round-trip and a noisy error.

Happy to test a fix - we hit this daily in device testing.

danielgranhao:
The Boltz service is not operational, and it is unclear when or whether
it will be again. Registering it also slowed route discovery: route
fetching retries a network-bound client construction cached only on
success, and providers are queried sequentially, so an unreachable
backend put a 60s timeout in front of every route list.

Stop registering the provider. The modules stay compiled and tested so
the wiring can be restored in one place. The public enum variants are
retained so persisted payment history keeps decoding.

### DIFF
--- a/crates/breez-sdk/core/src/cross_chain/mod.rs
+++ b/crates/breez-sdk/core/src/cross_chain/mod.rs
@@ -132,9 +132,7 @@ pub(crate) fn derive_btc_leg_transfer_id(
 #[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
 pub enum CrossChainProvider {
     Orchestra,
-    /// Not operational: no routes are offered under this provider, and it is
-    /// unclear when or whether that will change. The variant is retained so
-    /// persisted payment history keeps decoding.
+    /// Not operational: no routes are currently offered under this provider.
     Boltz,
 }
 
--- a/crates/breez-sdk/core/src/cross_chain/mod.rs
+++ b/crates/breez-sdk/core/src/cross_chain/mod.rs
@@ -3,14 +3,19 @@
 //! The [`CrossChainService`] trait abstracts route discovery, quoting, and
 //! sending. Each provider module (e.g. `orchestra`, `boltz`) implements it.
 
+// Boltz is not registered as a provider (see `sdk_builder`): the service is not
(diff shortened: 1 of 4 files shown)


crates/breez-sdk/core/src/cross_chain/mod.rs
```diff
@@ -127,6 +132,9 @@ pub(crate) fn derive_btc_leg_transfer_id(
 #[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
 pub enum CrossChainProvider {
     Orchestra,
+    /// Not operational: no routes are offered under this provider, and it is
+    /// unclear when or whether that will change. The variant is retained so
+    /// persisted payment history keeps decoding.
```
dangeross: nit: For a dev facing doc I wonder if this is better?
```suggestion
    /// Not operational: no routes are currently offered under this provider.
```

dangeross (approved):
LGTM

roeierez (approved):

PR #1106, opened 2026-09-08
Attribute a Bolt11 invoice settled over Spark to the invoice it paid

### LINKED ISSUE #432: No paid event emitted when invoice is settled via Spark

danielgranhao:
Closes #432

When a payer takes the Spark destination advertised by a Bolt11, the transfer never reaches the SSP, and both ends were left holding a bare Spark transfer where a Lightning payment belonged. The receiver got no paid event for the invoice it issued, so anything waiting on that invoice timed out. The payer lost the invoice, its description and the payee.

Both ends now record which Bolt11 the transfer settled, the payer when it sends and the receiver when it mints the invoice, and report the payment as that invoice being paid. Each half is shared with the wallet's other devices over real-time sync, so a device that neither paid nor minted reports the same thing.


crates/breez-sdk/core/src/sdk/payments/client_signing.rs
```diff
@@ -104,25 +105,48 @@ pub(in crate::sdk) async fn build_unsigned_transfer_package(
                 _ => (sdk.config.prefer_spark_over_lightning, None),
             };
             if prefers_bolt11_spark_route(prefer_spark, prepare_response) {
-                let spark_address = sdk
+                let fallback = sdk
                     .spark_wallet
-                    .extract_spark_address(&invoice_details.invoice.bolt11)?
+                    .extract_spark_fallback(&invoice_details.invoice.bolt11)?
```
dangeross, line 110: Just wondering if we should run `validate_fallback_to_pay` here too? Is it possible for the client-signing path to be symmetric?

crates/breez-sdk/core/src/persist/mod.rs
```diff
@@ -784,6 +867,28 @@ impl ObjectCacheRepository {
         }
     }
 
+    /// Notes that this wallet has minted a Bolt11 that can settle over Spark.
```
dangeross: I notice there is a lot of occurrences of "minting"/"minted" an invoice in the comments. Minting has a different connotation for me, specifically relating to tokens. Can we avoid using "mint"?

crates/breez-sdk/core/src/persist/mod.rs
```diff
@@ -44,6 +44,28 @@ const TX_CACHE_KEY: &str = "tx_cache";
...
+pub(crate) fn spark_invoice_expiry_secs(spark_invoice: &str) -> Option<u64> {
+    let address: spark_wallet::SparkAddress = spark_invoice.parse().ok()?;
+    let expiry = address.spark_invoice_fields?.expiry_time?;
+    Some(
+        expiry
+            .duration_since(platform_utils::time::UNIX_EPOCH)
+            .ok()?
+            .as_secs(),
+    )
+}
```
dangeross, line 72: Seems to have been dropped into a const block

crates/breez-sdk/wasm/js/mysql-storage/index.cjs
```diff
@@ -109,6 +128,17 @@ function toBool(value) {
  * Maps a brz_cross_chain_swaps row to the camelCase StoredCrossChainSwap shape
  * the SDK expects, coercing TINYINT(1) to a boolean.
  */
```
dangeross, line 144: Looks like crossChainSwapFromRow lost its docs

crates/breez-sdk/wasm/js/node-storage/index.cjs
```diff
@@ -787,7 +808,7 @@ class SqliteStorage {
   setLnurlMetadata(metadata) {
     try {
       const stmt = this.db.prepare(
-        "INSERT OR REPLACE INTO lnurl_receive_metadata (payment_hash, nostr_zap_request, nostr_zap_receipt, sender_comment) VALUES (?, ?, ?, ?)"
+"INSERT OR REPLACE INTO lnurl_receive_metadata (payment_hash, nostr_zap_request, nostr_zap_receipt, sender_comment) VALUES (?, ?, ?, ?)"
```
dangeross: Lost its indent?

dangeross (commented):
Lookin good

crates/breez-sdk/core/src/sdk/payments/client_signing.rs
danielgranhao: Made it symmetric

crates/breez-sdk/core/src/persist/mod.rs
danielgranhao: Fixed.

crates/breez-sdk/core/src/persist/mysql/storage.rs
```diff
@@ -567,6 +567,41 @@ impl MysqlStorage {
...
+                   MODIFY htlc_status VARCHAR(64) NULL,
+                   MODIFY htlc_expiry_time BIGINT NULL",
+            )],
+            // Migration 26: Bolt11s settled over Spark, one table per direction.
+            // Sends are keyed by the transfer that paid. Receives are keyed by
+            // the Spark invoice the Bolt11 embeds, written when it is created and
+            // dropped once it has expired. Born multi-tenant.
+            vec![
+                Migration::sql(
+                    "CREATE TABLE IF NOT EXISTS brz_spark_settled_bolt11_sends (
```
JssDWt, line 584: Should these tables be joined when selecting payments, so that the bolt11 is always taken from there? Even if the payment itself is overwritten at some point? Like payment_metadata I mean?

crates/breez-sdk/breez-itest/tests/bolt11_spark_fallback.rs
```diff
@@ -0,0 +1,134 @@
...
+        })
+        .await?
+        .payment;
+    assert_bolt11_settled_over_spark(&alice_stored, &invoice)?;
+
+    // Bob's payment must report as the Bolt11 being paid, not as a bare Spark
+    // receive, and must carry no HTLC: none was ever created.
+    let received =
+        wait_for_payment_succeeded_event(&mut bob.events, PaymentType::Receive, 60).await?;
+    assert_eq!(received.amount, u128::from(AMOUNT_SATS));
```
JssDWt, line 98: I think the event should contain the bolt11 here?

crates/breez-sdk/core/src/sdk/payments/receive.rs
```diff
@@ -355,28 +359,101 @@ pub(super) async fn receive_bolt11_invoice_inner(
...
+    // transfer settling it would be attributed to an invoice already expired.
+    let expiry_secs = expiry_secs.unwrap_or(DEFAULT_RECEIVE_EXPIRY_SECS);
+    let expiry_time = SystemTime::now().checked_add(Duration::from_secs(u64::from(expiry_secs)));
+    let invoice = sdk
+        .spark_wallet
+        .create_spark_invoice(
+            amount_sats.map(u128::from),
+            None,
+            expiry_time,
+            Some(description.to_string()),
```
JssDWt: I think we should skip the description here, as it may be too large. The JS SDK also does this:
https://github.com/buildonspark/spark/blob/0b3a32a05c9ac06cc411683551dd1f1bde9d0caa/sdks/js/packages/spark-sdk/src/spark-wallet/spark-wallet.ts#L4262-L4272

crates/breez-sdk/core/src/sdk/payments/polling.rs
```diff
@@ -71,6 +77,29 @@ pub(super) async fn finalize_payment(sdk: &BreezSdk, mut payment: Payment) -> bo
...
+/// ingests that transfer and records it against the Bolt11, so storage is where
+/// the Spark-settled payment shows up.
+async fn poll_lightning_receive(
+    sdk: &BreezSdk,
+    ssp_id: &str,
+    invoice: &str,
+) -> Result<Option<Payment>, SdkError> {
+    if let Some(payment) = sdk
+        .storage
+        .get_payment_by_invoice(invoice.to_string())
```
JssDWt, line 88: Claude notifies me that:

> The IndexedDB invoice index uses key path details.invoice, but details is stored as a JSON string, so the index is always empty. The bug already exists on main; this is the first code that depends on it.

> Rust Postgres's get_payment_by_invoice errors when two payments share an invoice. That's now easier to hit (one Bolt11 paid over both Spark and Lightning), and the error propagates out of the poll.

crates/breez-sdk/core/src/sdk/payments/receive.rs
```diff
@@ -355,28 +359,101 @@ pub(super) async fn receive_bolt11_invoice_inner(
...
+async fn build_fallback(
+    sdk: &BreezSdk,
+    amount_sats: Option<u64>,
+    expiry_secs: Option<u32>,
+    description: &str,
+    receiver_identity_public_key: Option<PublicKey>,
+) -> Result<LightningReceiveFallback, SdkError> {
+    if !sdk.config.prefer_spark_over_lightning || receiver_identity_public_key.is_some() {
+        return Ok(LightningReceiveFallback::None);
+    }
```
JssDWt: This is different from before, because invoices with receiver_identity_public_key would use LightningReceiveFallback::Address. Is the change intended?

crates/spark/src/utils/bolt11_fallback.rs
```diff
@@ -0,0 +1,126 @@
...
+    /// A BOLT11 carrying a bare Spark address in the version-31 fallback field.
+    const INVOICE_WITH_FALLBACK_ADDRESS: &str = "lnbcrt10u1p57lj4upp5qxa002jtss48lwgqrzqwhsr5388gv09k4tllfy9hlv78akygcmzssp5hkpfxyy7xrd7qwwc043cfjma9u4z5vasxuyrgqrh6wrvyx6xkevqxq9z0rgqnp4qtlyk6hxw5h4hrdfdkd4nh2rv0mwyyqvdtakr3dv6m4vvsmfshvg6rzjqf6plzwgkgyrrwdygdekj8w8frztu8k7dz2x9nefwyc70vergwrqeapyqr6zgqqqq8hxk2qqae4jsqyugqcqzpudqswfhh2ar9dp5kuarn9qyyssqn9tv97k2a24a45vkt3zvckjt80ph2luwjje5c8ymtf7qr3m4nkaqrqzuc9gzvjufxvvhk2lvzfxerccvtclakmyq43hfyfuwzgkx7hspfh5vzf";
```
JssDWt: nit: This one actually contains a route hint, not a fallback field

JssDWt (commented):
Looking nice

crates/breez-sdk/core/src/persist/mysql/storage.rs
danielgranhao: Good idea. I like it, especially because we drop reliance on ingest time resolution

crates/breez-sdk/core/src/sdk/payments/receive.rs
danielgranhao: Reverted

PR #1110, opened 2026-09-10
feat(cross-chain): expose Orchestra route amount limits

### LINKED ISSUE #1080: Cross-chain: no way to learn a route's min/max before a quote fails
## Problem

`getCrossChainRoutes` returns provider/chain/asset/decimals but no amount limits, and there is no other API to fetch them. A wallet cannot validate an amount before quoting, so users run into avoidable errors:

- Sending $1 USDT to a Tron address via Orchestra fails at prepare with `Network error: Amount too small (code: 400)` - only after address entry, amount entry, and a provider round-trip.
- The client has no number to show; apps are tempted to cache past rejections as pseudo-limits, which goes stale.

## What we found investigating

- **Orchestra**: `GET /v1/orchestration/routes` (public) carries no min/max on any of its ~2,861 routes, so the SDK currently has nothing to plumb - this likely needs coordination with Flashnet to add limits to the routes (or estimate) payload.
- **Orchestra errors**: `extract_error_message` keeps only `error.message` from the error body. If Flashnet includes structured limit data in rejections (or adds it), surfacing it as a typed error (e.g. `AmountTooSmall { min }`) instead of a plain string would at least let wallets show the real number at failure time - Boltz-side code already does this (`BoltzError::AmountOutOfRange { min }`).
- **Boltz**: pair limits exist in its API and `boltz-client` knows them at prepare, but they are not exposed on the route pairs either.

## Ask

1. Expose optional `limits` (min/max, documented denomination) on `CrossChainRoutePair` where the provider publishes them (Boltz today; Orchestra once Flashnet provides them).
2. Until then: make the too-small rejection structured with the minimum included, for both providers, so wallets can be honest at the earliest knowable moment.

Context: building Digital Dollars flows on Breez SDK - Spark 0.23.0. Happy to test.

danielgranhao:
Closes #1080.

`getCrossChainRoutes` returned no amount bounds, so a wallet could not validate an amount before quoting, and an out-of-range payment only failed at prepare with an opaque `Network error: Amount too small (code: 400)`. Route discovery now reads Orchestra's limits endpoint, so routes come back with their bounds attached, and a rejected amount maps to `SdkError::CrossChainAmountOutOfRange` carrying the published bound.

Bounds are per accepted asset rather than per route, because the same endpoint can carry a dust floor when funded with sats and none when funded with USDB. `accepted_assets` therefore becomes `Vec<CrossChainAcceptedAsset>`, which is breaking for anyone reading that field.


crates/breez-sdk/core/src/cross_chain/mod.rs
```diff
@@ -155,6 +155,48 @@ pub enum SparkAsset {
...
+    /// Smallest order value accepted, in USD cents.
+    pub min_usd_cents: Option<u64>,
+    /// Largest order value accepted, in USD cents.
+    pub max_usd_cents: Option<u64>,
+    /// Whether the provider can still reject an amount that satisfies the
+    /// bounds above. Live routing legs impose moving minimums and liquidity
+    /// ceilings that the published bounds do not capture, so when this is set
+    /// the bounds are a floor on what will be rejected, not the whole truth.
+    /// Validate a concrete amount by preparing the payment.
+    pub dynamic_limits_possible: bool,
```
roeierez, line 185: Just wondering if we need these and what the user should do with these values.

roeierez (commented):
Just a comment about possibly minimize the limits struct, otherwise LGTM.

crates/breez-sdk/core/src/cross_chain/mod.rs
danielgranhao: Goal was for integrators to avoid "guaranteeing" acceptance of an amount when this was true but in practice it's true for almost all our supported assets, so it doesn't even help much. Removed

roeierez (approved):

crates/breez-sdk/core/src/cross_chain/mod.rs
```diff
@@ -155,6 +155,47 @@ pub enum SparkAsset {
     Token { token_identifier: String },
 }
 
+/// Amount bounds a provider publishes for moving a route with one Spark-side
+/// asset.
+///
+/// The bounds are a reliable no, never a reliable yes. An amount outside them
```
dangeross: Not sure what this means, probably can be removed

docs/breez-sdk/src/guide/cross_chain.md
```diff
@@ -57,6 +57,23 @@ URIs whose recipient address doesn't match the scheme's address family (e.g. a `
...
+| ------------------------------ | -------------------------------------------------------------------- |
+| {{#name min_amount}} / {{#name max_amount}}     | Bounds in the base units of the asset paid in: the Spark-side asset on a send, the external asset on a receive |
+| {{#name min_usd_cents}} / {{#name max_usd_cents}} | Bounds on the order's value, in USD cents                          |
+
+Bounds are per asset rather than per route, because the same external endpoint can carry a dust floor when moved as sats and none when moved as a token. Either denomination can be absent, and a provider may publish none at all, so treat a missing bound as "no published limit" rather than as zero.
+
+**The bounds are a reliable no, never a reliable yes.** Reject an amount that falls outside a published bound before preparing the payment: the provider rejects it too, so the check costs nothing and never blocks a payment that would have gone through. Satisfying the bounds is not a guarantee. Some routes enforce a tighter bound than they publish, so a payment inside the published band can still be rejected, and preparing it is what decides a concrete amount.
```
dangeross: Same here if dropped

dangeross: This also sounds like its describing internals that I think is not needed in the docs. Wdyt?

docs/breez-sdk/src/guide/cross_chain.md
```diff
@@ -57,6 +57,23 @@ URIs whose recipient address doesn't match the scheme's address family (e.g. a `
...
+
+Because of that, the bounds are for catching an amount the user could not have meant, not for guiding them to a sensible one. Fees weigh heavily near the floor, so the smallest amount worth sending is usually well above the smallest one accepted.
```
dangeross: Doesn't seem needed. I think the limits are more useful as a UI guide

dangeross (commented):
Looks good. Just docs nits

crates/breez-sdk/core/src/cross_chain/mod.rs
danielgranhao: I clarified it but didn't remove it completely. Let me know if that's ok.

docs/breez-sdk/src/guide/cross_chain.md
danielgranhao: Same

dangeross (approved):
LGTM

PR #1111, opened 2026-09-10
Retry transient chain service errors, disable the Spark HTLC tests

danielgranhao:
The itests have been failing on most runs for the past week. This fixes both causes.

Most of the failures were connection-level. The shared regtest esplora reaps idle connections, and the chain service only retried on HTTP status codes, so the next request off a reaped connection surfaced a broken pipe to the caller. Those are now retried, on the existing backoff and within a single request timeout.

The rest are the direct Spark HTLC tests, which cannot pass while HTLCs are restricted to transfers to and from the SSP. They are `#[ignore]`d and tracked in #1123.

Only `rest_client.rs` is production code; everything else here is test-only.

### DIFF
--- a/crates/breez-sdk/breez-itest/tests/spark_htlcs.rs
+++ b/crates/breez-sdk/breez-itest/tests/spark_htlcs.rs
@@ -191,7 +191,13 @@ async fn test_01_htlc_success(
 }
 
 /// Test 2: Send payment from Alice to Bob using Spark transfer and fail to claim before expiry
+///
+/// TODO: re-enable once direct HTLCs are available again. While they are
+/// disabled, the operators stop serving a returned HTLC's preimage request, so
+/// the refunded transfer syncs with no HTLC details and the
+/// `SparkHtlcStatus::Returned` filter below matches nothing.
 #[rstest]
+#[ignore = "direct HTLCs are disabled; the refund carries no HTLC details"]
 #[test_log::test(tokio::test)]
 async fn test_02_htlc_refund(
     #[future] alice_sdk: Result<SdkInstance>,
--- a/crates/breez-sdk/core/src/chain/rest_client.rs
+++ b/crates/breez-sdk/core/src/chain/rest_client.rs
@@ -200,6 +200,18 @@ impl RestClientChainServiceInner {
         Ok(response)
     }
 
+    /// Sleeps out the current backoff and charges an attempt, or reports that
+    /// the budget is spent so the caller surfaces the error it is holding.
+    async fn backoff(&self, attempts: &mut usize, delay: &mut Duration) -> bool {
+        if *attempts >= self.max_retries {
+            return false;
+        }
+        tokio::time::sleep(*delay).await;
+        *attempts = attempts.saturating_add(1);
+        *delay = delay.saturating_mul(2);
+        true
+    }
+
     async fn get_with_retry(
         &self,
         url: &str,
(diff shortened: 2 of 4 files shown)


crates/breez-sdk/core/src/chain/rest_client.rs
```diff
@@ -423,10 +458,25 @@ fn is_status_retryable(status: u16) -> bool {
...
+///
+/// A shared esplora endpoint reaping idle HTTP/2 flows answers the next request
+/// off that connection with a broken pipe, which reqwest reports as a request
+/// error. Retrying the POST is safe for the same reason: the caller was about to
+/// see an error either way, and a re-broadcast of a transaction the server did
+/// accept is answered "already known" rather than acted on twice.
+fn is_transport_retryable(error: &HttpError) -> bool {
+    matches!(
+        error,
+        HttpError::Request(_) | HttpError::Connect(_) | HttpError::Timeout(_) | HttpError::Body(_)
```
dangeross: Wondering if we should leave out `HttpError::Timeout`. With `max_retries = 5` and the 60s `REQUEST_TIMEOUT`, retrying it turns a 60s worst case into ~6 minutes on paths that have no deadline of their own (`check_and_claim_static_deposits`, `refund_fee_to_outbid`, `deposit_confirmations_at_tip`).

dangeross: I'm wondering if it's worth splitting `HttpError::Body` into two errors. Currently `read_capped_bytes` reports the `MAX_RESPONSE_BYTES` refusal with it, and a body that failed mid-stream gets the same variant. Only the latter is probably worth another attempt here. Wdyt?

crates/breez-sdk/breez-itest/tests/spark_htlcs.rs
```diff
@@ -191,7 +191,13 @@ async fn test_01_htlc_success(
 }
 
 /// Test 2: Send payment from Alice to Bob using Spark transfer and fail to claim before expiry
+///
+/// TODO: re-enable once direct HTLCs are available again. While they are
+/// disabled, the operators stop serving a returned HTLC's preimage request, so
+/// the refunded transfer syncs with no HTLC details and the
+/// `SparkHtlcStatus::Returned` filter below matches nothing.
```
dangeross: I wonder if it's worth raising an issue to track this

crates/breez-sdk/core/src/chain/rest_client.rs
```diff
@@ -423,10 +458,25 @@ fn is_status_retryable(status: u16) -> bool {
     RETRYABLE_ERROR_CODES.contains(&status)
 }
 
+/// Whether a failure that produced no response at all is worth another attempt.
+///
+/// A shared esplora endpoint reaping idle HTTP/2 flows answers the next request
+/// off that connection with a broken pipe, which reqwest reports as a request
+/// error. Retrying the POST is safe for the same reason: the caller was about to
+/// see an error either way, and a re-broadcast of a transaction the server did
+/// accept is answered "already known" rather than acted on twice.
```
dangeross: nit: this comment seems very specific to `do_broadcast_transaction`, I'm not sure it belongs here

dangeross (commented):
Lookin good

crates/breez-sdk/core/src/chain/rest_client.rs
danielgranhao: Remove it

danielgranhao: No longer retrying on body errors

crates/breez-sdk/breez-itest/tests/spark_htlcs.rs
danielgranhao: Created https://github.com/breez/spark-sdk/issues/1123

crates/breez-sdk/core/src/chain/rest_client.rs
danielgranhao: fixed

crates/breez-sdk/core/src/chain/rest_client.rs
```diff
@@ -423,10 +460,23 @@ fn is_status_retryable(status: u16) -> bool {
...
 }
 
+/// Whether a failure that produced no response at all is worth another attempt.
+///
+/// A shared esplora endpoint reaping idle HTTP/2 flows answers the next request
+/// off that connection with a broken pipe, which reqwest reports as a request
+/// error. Timeouts stay out: the per-request budget is already 60s, so retrying
+/// one stalls the caller for minutes rather than riding out a blip. So does
+/// `Body`, which a response over the size cap shares with a truncated one, and
+/// the cap is refused the same way every attempt.
```
JssDWt: This comment can probably be deleted.

JssDWt (approved):
Looks good. 
Having retries here make we wonder whether we should have a total timeout configured on this level? We have the timeout from reqwest, but retrying here can duplicate that. It would be nice to have a total timeout that actually returns after 60 seconds always, regardless of retries.

crates/breez-sdk/core/src/chain/rest_client.rs
```diff
@@ -423,10 +480,30 @@ fn is_status_retryable(status: u16) -> bool {
     RETRYABLE_ERROR_CODES.contains(&status)
 }
 
+fn is_transport_retryable(error: &HttpError) -> bool {
+    matches!(error, HttpError::Request(_))
+}
+
+/// Fails `work` once `budget` is gone, whatever attempt it is on.
+async fn within_budget<T>(
```
roeierez: NIT: budget => timeout ?

roeierez (approved):

PR #1112, opened 2026-09-11
Fix conversions failing on a null hostname in the Flashnet pool listing

dangeross:
Mainnet BTC/USDB conversions failed with `No conversion pools available` while `GET /v1/pools` returned 200 with 11 pools. A pool created 2026-09-11 carries `"hostName": null`, which the Flashnet API documents as nullable, but
`Pool::host_name` was a required `String`. One pool failed the whole `ListPoolsResponse`, and the resulting error was discarded, so an unparseable listing was reported as an absence of liquidity.

### Schema corrections
Audited the response models against the published schema. Three fields were declared required that the spec marks optional:

| Field | Was | Now |
|---|---|---|
| `Pool.host_name` | `String` | `Option<String>` |
| `Pool.created_at` / `updated_at` | `String` | `Option<String>` |
| `ClawbackTransfer.lp_identity_public_key` | `PublicKey` | `Option<PublicKey>` |

### Hardening

- `ListPoolsResponse.pools` skips individual pools that fail to deserialize, logging each, rather than failing the listing.
- `list_pools` warns when the server reports pools but none parsed.
- `get_conversion_pool` logs the underlying `FlashnetError` instead of discarding it.

### Pool selection

A V3 pool is quoted from its advertised price rather than from reserves, so one omitting reserves was priced against nothing. Quoting now requires a covering output reserve or a non-zero `total_liquidity`.

### Refund path

`lp_identity_public_key` is signed into the clawback intent, so a transfer cannot be clawed back without it. Reconcile falls back to the pool recorded on our own conversion (`ConversionInfo::Amm`), and skips a row it cannot address instead of aborting the pass.

### DIFF
--- a/crates/breez-sdk/core/src/token_conversion/flashnet.rs
+++ b/crates/breez-sdk/core/src/token_conversion/flashnet.rs
@@ -537,13 +537,16 @@ impl FlashnetTokenConverter {
                 res.skipped = res.skipped.saturating_add(1);
                 continue;
             }
+            let Some(pool_id) = self.resolve_clawback_pool(&transfer).await else {
+                warn!(
+                    "Reconcile: cannot claw back {}: no pool in the listing or the local record",
+                    transfer.id
+                );
+                res.failed = res.failed.saturating_add(1);
+                continue;
+            };
             match self
-                .clawback_and_record_refunded(
-                    &transfer.id,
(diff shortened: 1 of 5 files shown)


danielgranhao (approved):
LGTM

roeierez (approved):

PR #1113, opened 2026-09-11
flashnet: parse the pool listing one pool at a time; allow a null host

satoshisound:
`ListPoolsResponse.pools` was a `Vec<Pool>` straight from serde, so one entry the model can't represent fails the whole response, `list_pools` returns an error in both directions, `get_conversion_pool` swallows both and reports `NoPoolsAvailable` — and every Stable Balance conversion for every client stops. Pool creation is permissionless and the listing is server-shaped, so that is a standing single point of failure for the feature.

It fired on mainnet BTC/USDB at 16:19 UTC on 2026-09-11: a pool (`034f31eb0f56…`, empty, tick 0) was created with `"hostName": null`, and `Pool.host_name` is a plain `String` in every release. Our wallets logged:

```
WARN breez_sdk_spark::token_conversion::flashnet: No conversion pools available
WARN breez_sdk_spark::stable_balance::queue: Auto-conversion failed: NoPoolsAvailable
```

while `GET /v1/pools?assetAAddress=…&assetBAddress=…` returned 11 pools including the live one.

**Fix**
- Deserialise `pools` element by element and skip, with a warning naming the pool, any entry that doesn't parse.
- Make `host_name` `Option<String>`, which is what the backend actually sends.

Regression tests use the listing from that day (the live pool plus the null-host pool), and a mangled variant to show one bad entry no longer drops the rest. `totalCount` is left as the server's figure.

**Verified locally**: `cargo fmt`, `cargo clippy -p flashnet --all-targets -- -D warnings`, `cargo test -p flashnet`. Also worth asking Flashnet why a pool can be created without a host, but the client has to survive it either way.

### DIFF
--- a/crates/flashnet/src/amm/models.rs
+++ b/crates/flashnet/src/amm/models.rs
@@ -473,8 +473,40 @@ impl ListPoolsRequest {
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(rename_all = "camelCase")]
 pub struct ListPoolsResponse {
+    /// Pools are parsed one at a time and a pool that doesn't parse is
+    /// skipped, not fatal. Pool creation is permissionless and the listing is
+    /// server-shaped, so a single entry with a field this client doesn't
+    /// expect must not empty the whole listing: on mainnet BTC/USDB on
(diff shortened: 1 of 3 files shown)


dangeross:
Duplicate of #1112

PR #1114, opened 2026-09-11
flashnet: drop pools priced far from the market before scoring

satoshisound:
`validate_for_swap` checks a pool against itself: price vs tick, price vs reserves, fees in range. Nothing checks it against its peers. Pool creation is permissionless, and a concentrated pool is quoted from its advertised price alone, so a pool holding just enough of the output asset to pass the reserve check, at an absurd but tick-corroborated price, quotes a near-zero `amount_in`, takes the full fee-efficiency score (50% weight) and beats the real pool; the swap then fails at the AMM with FSAG-4201.

This is the residual of the mainnet BTC/USDB incident of 2026-09-07: two empty V3 pools (host `auditlab`, ticks ±690810, prices 1e30 / 1e-30, one poisoning each direction) out-scored the real pool (`037579ae…`, 1.6 BTC / $136k, $86M 24h volume) on 0.19.0 — scoring 7,300 vs 4,978 — and every conversion failed until we upgraded past 7ba9363a. The output-reserve check now refuses those two, but nothing stops the same trick with a few sats of funding.

**Fix**: after viability, drop pools whose `currentPriceAInB` is more than 2× away, in either direction, from the deepest viable pool's (TVL, else asset-B reserve). Real pools for one pair sit within arbitrage distance of each other. Pools with no price, or with no deep pool to compare against, are kept — the filter only removes what it can positively call an outlier.

Tests use the API's figures for the real pool from that day, a thinly-funded lure with a coherent tick at a millionth of the market price, and the no-reference case.

**Verified locally**: `cargo fmt`, `cargo clippy -p flashnet --all-targets -- -D warnings`, `cargo test -p flashnet`.

### DIFF
--- a/crates/flashnet/src/amm/pool_selection.rs
+++ b/crates/flashnet/src/amm/pool_selection.rs
@@ -76,6 +76,17 @@ pub fn select_best_pool(
         ));
     }
 
+    // Drop pools whose advertised price is nowhere near the market for the
+    // pair. Validation above checks each pool against itself (price vs tick,
+    // price vs reserves); this checks it against its peers. Pool creation is
+    // permissionless, and a concentrated pool is quoted from its price alone,
+    // so a pool holding just enough of the output asset to pass the reserve
+    // check but priced at an absurd, tick-corroborated level quotes a
+    // near-zero amount_in, takes the fee-efficiency score outright, and the
+    // swap then fails at the AMM. Real pools for one pair sit within arbitrage
+    // distance of each other, far inside the band used here.
+    let viable_pools = filter_price_outliers(viable_pools);
+
     // Filter out non V3 concentrated pools if any V3 concentrated pool exists
     let has_v3_concentrated = viable_pools
         .iter()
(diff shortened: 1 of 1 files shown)


crates/flashnet/src/amm/pool_selection.rs
```diff
@@ -149,6 +160,53 @@ pub fn select_best_pool(
...
+        return pools;
+    };
+    let (low, high) = (
+        reference_price / PRICE_SANITY_FACTOR,
+        reference_price * PRICE_SANITY_FACTOR,
+    );
+    pools
+        .into_iter()
+        .filter(|(pool, _)| match pool.current_price_a_in_b {
+            Some(price) if price < low || price > high => {
```
dangeross: Should we normalise the price to the swap direction before comparing? `current_price_a_in_b` is B per A in the pool's *own* ordering, and the listing carries both orderings for the same pair. `LIVE_LISTING` in `pool_validation.rs` has two BTC/USDB pools with `assetAAddress` = USDB and two with it = BTC, and `get_conversion_pool` queries both directions and merges them into one list. So we build the band from a reference in one denomination and apply it to prices in the other.

Claude checked both directions by adding temporary tests to this module:

- an honest reversed pool (`assetA` = USDB, price `1/773`) gets dropped as an outlier against the real pool's `773`. That one is new behaviour from this PR
- a lure with its pair flipped and `currentPriceAInB = 773`, meaning 773 sats per USDB unit, lands dead centre of the band, survives the filter, quotes `260` where the real pool quotes `154_889_949`, and `select_best_pool` returns it. `price_matches_tick` doesn't catch it either, it accepts `p` or `1/p`

`select_best_pool` already has `asset_in_address`. Could we pass it in and compare `if asset_in == pool.asset_a_address { price } else { 1.0 / price }`?

crates/flashnet/src/amm/pool_selection.rs
```diff
@@ -149,6 +160,53 @@ pub fn select_best_pool(
...
+/// [`PRICE_SANITY_FACTOR`] away, in either direction, from that of the deepest
+/// pool (by TVL, else asset-B reserve). Pools without a price, or with no deep
+/// pool to compare against, are kept: this guard only removes a pool it can
+/// positively call an outlier.
+fn filter_price_outliers(pools: Vec<(Pool, u128)>) -> Vec<(Pool, u128)> {
+    let depth = |pool: &Pool| -> u128 {
+        pool.tvl_asset_b
+            .map_or(pool.asset_b_reserve, |v| Some(u128::from(v)))
+            .unwrap_or(0)
+    };
```
dangeross: `tvl_asset_b` is denominated in the pool's own asset B, so this takes a max over two different units. Claude reconstructed the field from the live payload to confirm that reading: `assetBReserve + assetAReserve * currentPriceAInB` gives `197574061893` exactly, so it's the whole pool valued in asset B.

For BTC/USDB that's a 642x skew. The same pool that reports `197,574,061,893` with USDB as its asset B would report `307,785,755` with BTC as its asset B. So a BTC-as-B pool needs roughly 642x the real value to become the reference, which makes the winner a function of which token has more decimals rather than which pool is deepest.

`max_tvl` on line 117 has the same issue, but there it only normalises a score. Here it decides which pools get dropped.

Worth taking depth in the output asset instead?

crates/flashnet/src/amm/pool_selection.rs
```diff
@@ -149,6 +160,53 @@ pub fn select_best_pool(
...
+    let Some(reference_price) = pools
+        .iter()
+        .filter(|(pool, _)| {
+            pool.current_price_a_in_b
+                .is_some_and(|p| p.is_finite() && p > 0.0)
+        })
+        .max_by_key(|(pool, _)| depth(pool))
+        .filter(|(pool, _)| depth(pool) > 0)
```
dangeross: Nothing corroborates `tvl_asset_b`. `validate()` checks price, price change, bonding progress, fees, curve and tick, and never looks at TVL, so the pool that sets the reference is picked by a field it supplies itself.

Claude took the lure from `thinly_funded_pool_with_an_absurd_price_is_excluded_as_an_outlier` and changed only `tvl_asset_b` to a large value. It still passes `validate()`, becomes the reference, and the real pool gets dropped as the outlier instead. `select_best_pool` returns the lure.

Not a regression, it wins on `main` too. But is the backend's TVL something we're treating as trusted? The module header in `pool_validation.rs` reads like we aren't, and this is the first place TVL decides what gets eliminated rather than just how things rank.

crates/flashnet/src/amm/pool_selection.rs
```diff
@@ -149,6 +160,53 @@ pub fn select_best_pool(
...
+            .map_or(pool.asset_b_reserve, |v| Some(u128::from(v)))
+            .unwrap_or(0)
+    };
+    let Some(reference_price) = pools
+        .iter()
+        .filter(|(pool, _)| {
+            pool.current_price_a_in_b
+                .is_some_and(|p| p.is_finite() && p > 0.0)
+        })
+        .max_by_key(|(pool, _)| depth(pool))
```
dangeross: nit, and probably never hits in practice: `max_by_key` returns the *last* maximum, and the caller hands us `HashMap::into_values()` (`get_conversion_pool` in `token_conversion/flashnet.rs`), so input order varies per call. Claude confirmed both, 24 distinct orderings from the same 4 keys in one process.

So two pools tied on depth with different prices means the surviving set flips between runs. Exact ties are unlikely, but it would present as conversions intermittently picking a different pool with no code change.

The scoring `max_by` below already tiebreaks on volume. Worth doing the same here?

PR #1115, opened 2026-09-11
stable_balance: back off and retry failed auto-convert and deactivation

satoshisound:
The conversion worker treated a failed `AutoConvert` or `Deactivation` as done: `complete_task`, warn, move on. But a failed swap is refunded by the AMM as an incoming sats transfer, that receive re-queues `AutoConvert`, and the loop runs every few seconds for as long as the wallet is open. Observed on mainnet on 2026-09-11 while the selected pool had no liquidity: **127 attempts in 13 minutes**, each leaving a ±11,254-sat pair in the payment history:

```
WARN breez_sdk_spark::stable_balance::queue: Auto-conversion failed: ConversionFailed("Convert token failed, refund in progress: Pool has no liquidity")
```

A failed `Deactivation` was simply dropped, leaving the token holding in place behind a toggle that reads "off".

**Fix**: keep the failed task queued and schedule its next attempt with exponential backoff — 30 s, doubling, capped at one hour.
- `next_task` withholds a task until it is due; the worker sleeps for exactly the remaining delay alongside its usual notify/shutdown wakeups, since nothing else fires when the clock runs out.
- Success or replacement resets the backoff.
- A deactivation pushed while an auto-convert is backing off runs immediately, because it is the user acting.
- A failure reported for a task that has since been replaced is ignored, so the new task is not penalised for the old one's outcome.

The cap keeps a long outage self-healing: once liquidity is back, the next hourly attempt goes through with no restart.

Tests cover the schedule, due/not-due, reset on completion, deactivation preempting a backed-off auto-convert, and the replaced-task case. `spawn_conversion_worker`'s per-task match moved to `process_task` to stay under the line limit.

**Verified locally**: `cargo fmt`, `cargo clippy -p breez-sdk-spark --features sqlite --all-targets -- -D warnings`, `cargo test -p breez-sdk-spark --features sqlite`.

Related: #1113 (listing parse), and the follow-up that reports these failures as an event and persists a pending deactivation.

### DIFF
--- a/crates/breez-sdk/core/src/stable_balance/queue.rs
+++ b/crates/breez-sdk/core/src/stable_balance/queue.rs
@@ -4,6 +4,7 @@
 //! through a single queue to eliminate race conditions between the paths.
 
 use std::sync::Arc;
+use std::time::Duration;
 
 use platform_utils::tokio;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


dangeross:
Thanks @satoshisound. Closing this as we're already doing a round of hardening to handle this

PR #1116, opened 2026-09-11
stable_balance: report failed conversions as an event; persist a pending deactivation

satoshisound:
**Stacked on #1115** — the first commit here is that PR; only the second (`stable_balance: report failed conversions; persist a pending deactivation`) is new. Rebases to one commit once #1115 lands.

Two things an integrator could not see:

- **A failed sweep or deactivation conversion was only a WARN log line.** We currently detect them by scraping the log through a `tracing` layer, which is not a contract. Emit `SdkEvent::StableBalanceConversionFailed { conversion: AutoConvert | Deactivation, error, retry_in_secs }` each time one fails, so an app can show the user why the balance isn't moving, or deactivate Stable Balance rather than wait out the backoff. Balances are unchanged when it fires: the AMM refunds a failed swap. Mirrored in the wasm and flutter bindings and the Rust snippet.

- **A `Deactivation` task lived only in memory.** The active label is cleared the moment the user switches Stable Balance off, so after a failed conversion and a restart the wallet reported "off" with the whole token holding still in place and nothing left to move it. (That is the state one of our users was in: ~$15k USDB behind an "off" toggle, after `Deactivation conversion failed: … Pool has no liquidity`.) Persist the token to convert in the object cache when the deactivation is queued, clear it when the conversion succeeds — or when the user re-activates a token, which makes the holding wanted again — and re-queue it when the worker starts.

Tests cover persistence across queue/complete/clear and that a failure keeps the persisted entry.

**Verified locally**: `cargo fmt`, `cargo clippy -p breez-sdk-spark --features sqlite --all-targets -- -D warnings`, `cargo test -p breez-sdk-spark --features sqlite`, and the Rust doc snippet compiles. I could not build the wasm target on this machine (no rustup); the wasm mirror follows the existing `AutoOptimizationEvent` pattern exactly, so `xtask wasm-clippy` in CI is the check for that.

### DIFF
--- a/crates/breez-sdk/core/src/events.rs
+++ b/crates/breez-sdk/core/src/events.rs
@@ -56,6 +56,28 @@ pub enum SdkEvent {
     /// Emitted when the data a unilateral exit is built from has changed, so an
     /// exit state exported earlier is out of date and should be exported again.
     UnilateralExitStateChanged,
+    /// Emitted when a Stable Balance conversion the SDK runs on its own failed:
+    /// sweeping received bitcoin into the stable token, or converting the token
+    /// back to bitcoin on deactivation. The SDK keeps retrying with backoff;
+    /// `retry_in_secs` is how long until the next attempt. Balances are
+    /// unchanged: the AMM refunds a failed swap. Integrators that would rather
+    /// stop than wait can deactivate Stable Balance in response.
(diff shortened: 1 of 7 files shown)


dangeross:
Thanks @satoshisound. Closing this as we're already doing a round of hardening to address this

PR #1117, opened 2026-09-14
spark: bump protos

JssDWt:
Updates the vendored protos and the operator image to buildonspark/spark `0b3a32a05`.


JssDWt:
Recreating from a breez branch so the stack sits in one repo.

PR #1118, opened 2026-09-14
spark: add what a service provider needs from this crate

JssDWt:
Stacked on #1117.

- **`Signer::split_signing_key` / `combine_signing_keys`** — a deposit tree needs every node to hold its own key while its leaves still sum to the parent's.
- **`store_preimage_shares`** — the receiver side of a HODL receive: splits the preimage across the operators so the SSP's swap can return it atomically.
- **`HtlcService::receive_htlc_with_tweaks`** — the SSP side of a lightning receive: hands over the fronted leaves and gets the preimage back atomically.
- **Expose SSP-related operator functions** — `create_spark_tx`, `create_initial_timelock_refund_txs`, `RefundTransactions`, `signing_path`, `ephemeral_anchor_output`, `SigningResult`, `SparkRpcClient::transport`, `build_interceptor`, `Transport`.
- **`TreeNode::can_renew_refund_tx`** — so the SSP can refuse a leaf it could not later exit, before taking it on.
- **`TreeService::new` takes `Option<Arc<Swap>>`** — the SSP is itself the service provider, so it has none to swap with, but still wants the rest of the service.


JssDWt:
Recreating from a breez branch so the stack sits in one repo.

PR #1120, opened 2026-09-14
add what a service provider needs from the spark crate

JssDWt:
- **`Signer::split_signing_key` / `combine_signing_keys`**: a deposit tree needs every node to hold its own key while its leaves still sum to the parent's.
- **`store_preimage_shares`**: the receiver side of a HODL receive. It splits the preimage across the operators so the SSP's swap can return it atomically.
- **`HtlcService::receive_htlc_with_tweaks`**: the SSP side of a lightning receive. It hands over the fronted leaves and gets the preimage back atomically.
- **Expose SSP-related operator functions**: `create_spark_tx`, `create_initial_timelock_refund_txs`, `RefundTransactions`, `signing_path`, `ephemeral_anchor_output`, `SigningResult`, `SparkRpcClient::transport`, `build_interceptor`, `Transport`, `serialize_static_deposit_claim_payload`, `serialize_instant_static_deposit_claim_payload`.
- **`TreeService::new` takes `Option<Arc<Swap>>`**: the SSP is itself the service provider, so it has none to swap with, but still wants the rest of the service.


crates/spark/src/services/htlc.rs
```diff
@@ -122,6 +124,81 @@ impl HtlcService {
...
+                    .ok_or(ServiceError::SSPswapError(
+                        "Swap response did not contain a transfer".to_string(),
+                    ))?
+                    .try_into()?;
+                // Empty preimage means the share was not present (a HODL invoice):
+                // the transfer is committed but the preimage must be provided later.
+                let preimage = if response.preimage.is_empty() {
+                    None
+                } else {
+                    Some(Preimage::try_from(response.preimage)?)
```
roeierez: Shouldn't we ensure the preimage is the right one also? 

```
let preimage = Preimage::try_from(response.preimage)?;
if preimage.compute_hash() != *payment_hash {
    return Err(ServiceError::SSPswapError(
        "preimage does not match the payment hash".into(),
    ));
}
```

roeierez (changes requested):

crates/spark/src/services/htlc.rs
JssDWt: Yes, done!

roeierez (approved):

crates/spark/src/services/lightning.rs
```diff
@@ -35,6 +38,70 @@ const DEFAULT_RECEIVE_EXPIRY_SECS: u32 = 60 * 60 * 24 * 30; // 30 days
...
+///
+/// `invoice_string` is the bolt11 the operators validate the swap amount against;
+/// `identity_pubkey` is the share owner (the receiver), which the operators check
+/// against the swap's receiver.
+pub async fn store_preimage_shares(
+    operator_pool: &OperatorPool,
+    signer: &Arc<dyn Signer>,
+    split_secret_threshold: u32,
+    payment_hash: sha256::Hash,
+    preimage: Vec<u8>,
```
dangeross: Should we check that `preimage` hashes to `payment_hash` here? The two are supplied separately and nothing checks they agree at the API boundary.

dangeross (commented):
Looks good, just a validation question

crates/spark/src/services/lightning.rs
JssDWt: It certainly doesn't hurt! Done by taking only the preimage as a parameter.

dangeross (approved):
LGTM

PR #1122, opened 2026-09-14
allow signing with other key

JssDWt:
Stacked on #1120.

`LeafKeyTweak` and `TransferLeafInput` gain a non-optional `signing_key: FrostDerivation`, so a leaf can be signed with a key other than the one derived from its operator-assigned id. The SSP needs that: it builds trees before the operators name the nodes, so the keys committed at creation cannot derive from ids that do not exist yet.


kingonly:
What's the meaning of the spark: prefix to PRs?

JssDWt:
> What's the meaning of the spark: prefix to PRs?

It points at the `spark` rust library. The lowest level library in the repo.

kingonly:
> > What's the meaning of the spark: prefix to PRs?
> 
> It points at the `spark` rust library. The lowest level library in the repo.

Is that the team's convention? It's weird just to have some PRs using it. It's mostly llm driven and I'm not sure it's necessary (unless you see value in it). I just think more consistency would be nice.

crates/spark/src/services/transfer.rs
```diff
@@ -1019,6 +1025,7 @@ impl TransferService {
             let cpfp_sighash = sighash_from_tx(&cpfp_refund_tx, 0, node_tx_out)?;
             let cpfp = build_refund_signing_job(
                 &leaf.node.id,
+                &leaf.signing_key,
```
roeierez: Shouldn't we always use the derived key from the node id in the claim flow? 

```
FrostDerivation::SigningLeaf {
    leaf_id: leaf.node.id.clone(),
}
```

roeierez (commented):
Other than one comment LGTM.

crates/spark/src/signer/spark_signer.rs
dangeross: Does this need updating? Still says the old leaf key is derived from the `node.id`

crates/spark/src/services/models.rs
```diff
@@ -474,9 +473,13 @@ pub(crate) fn split_signing_commitments_by_variant<T>(
...
     /// For a claim, the incoming leaf key (ECIES-encrypted to our identity key
-    /// by the sender). `None` for outbound leaves, whose signing key is derived
-    /// from `node.id`.
+    /// by the sender). `None` for outbound leaves.
     pub incoming_key: Option<EncryptedSecret>,
+    /// The key this leaf is currently held under. Usually
+    /// `SigningLeaf { leaf_id: node.id }`, which is where a leaf received in a
+    /// transfer ends up, but a leaf created under an id its owner chose is held
+    /// under that one instead.
+    pub signing_key: FrostDerivation,
```
dangeross: Does renewal need the same? `renew_node` derives the refund key from `node.id`, so for a leaf held under a chosen id it builds a refund paying to the wrong key. I asked Claude to check with a local itest: a chosen-id leaf deposits, sends and claims fine, but renewal is rejected on the output pkscript, `timelock_manager` swallows it and retries, and the leaf ends up unspendable with `LEAF_RENEWAL_REQUIRED`. `retain_owned` looks like it makes the same assumption (not verified).

dangeross: Does renewal need the same? `renew_node` derives the refund key from `node.id`, so for a leaf held under a chosen id it builds a refund paying to the wrong key. `retain_owned` looks like it makes the same assumption (not verified).

I asked Claude to check with a local itest: a chosen-id leaf deposits, sends and claims fine, but the operators reject the renewal on the output pkscript, `timelock_manager` swallows that and retries, and once the timelock runs out they fail the send too, with reason `LEAF_RENEWAL_REQUIRED`.

crates/spark/src/services/transfer.rs
JssDWt: Agreed! The claim side now has its own type without signing_key, and it always signs with the node id key.

crates/spark/src/signer/spark_signer.rs
JssDWt: Fixed. Also fixed some other similarly mistaken comments.

crates/spark/src/services/models.rs
JssDWt: You are right. I did a better pass. Every path that spends a leaf now takes the key from the caller, including renewal and the unilateral exit sweep.

After fixing that there was still a discrepancy in SynchronousTreeService. It only works with node id keys, so it now checks a leaf is held under that key before renewing or swapping it, and leaves it alone otherwise. Injecting the custom key there was too cumbersome (and not needed for the SSP).

retain_owned I think is good.

roeierez (approved):
LGTM

dangeross (approved):
LGTM

PR #1125, opened 2026-09-15
docs: remove API docs pills from guide headings

dsbreez:
Removes the 71 `API docs` pills from the Breez SDK Spark guide headings.

## Why

The pills were added in b0aa06a8 (Aug 2025) with the original mdBook, and every page since followed the pattern. Four problems:

- **Rust-only in a nine-language book.** The pill links to rustdoc regardless of which language tab the reader has selected.
- **Three of the 61 targets are already broken**, and nothing in CI checks them:
  - `signer/trait.ExternalSparkSigner.html` — 404; the traits are `BreezSigner` / `EciesSigner` / `HmacSigner`
  - `turnkey/fn.create_turnkey_signer.html` — 404; `turnkey` is feature-gated, so `cargo doc --no-deps` never builds it
  - `struct.BreezSdk.html#method.connect` — page resolves, anchor does not exist (there is no `connect` method). This one is the h1 pill on `initializing.md`.
- **The rustdoc site has no index** — `https://breez.github.io/spark-sdk/` is a 404, so a pill drops the reader into a site with no navigation.
- **They leak into the AI corpus.** The snippets processor renders each pill as an `API docs: <url>` line in the published `.md` pages, so `llms-swift-full.txt` and `llms-kotlin-full.txt` each carried 71 Rust URLs.

## What changed

- 54 headings that existed only to hold a pill collapse back to plain markdown headings.
- 17 headings keep their raw HTML because other pages deep-link to their custom anchors (`config.md#usdc-usdt`, `optimize.md#optimize-leaves-full`, the four in `token_payments.md`, and so on) — only the pill line is removed.
- `styles.css`: the `a.tag` rules are now unused and removed.

Only the heading pills go. The two prose rustdoc links in `parse.md` stay: they point at the complete `InputType` list and the default-parser constant, neither of which is documented anywhere else in the book, so they carry information the guide would otherwise lose. (An earlier revision of this branch removed them; that is reverted.)

The `pagetoc.js` `.tag` filter and the optional tag group in the snippets processor's heading regex are left in place — both are harmless no-ops now, and the processor still needs that regex for the 17 raw-HTML headings.

## Verification

- No `class="tag"` remains anywhere under `docs/breez-sdk/src/`.
- All 65 in-book anchor links re-checked against the heading ids each page now exposes — no regressions. Every collapsed heading's id matched mdBook's generated slug exactly, so the anchors are unchanged.

Requested by Roy.


erdemyerebasmaz (approved):
Looks good

roeierez (approved):

kingonly:
No need for a PR and should be merged into the release branch

dsbreez:
Landed directly on `release` as 5ff8edb8 per Roy — no PR needed for docs. The commit is byte-identical to the approved diff here (squashed into one). `release` merges back into `main` on the usual cadence, so `main` picks it up there.

PR #1126, opened 2026-09-16
Add Signet support to the wallet SDK

benthecarman:
With https://github.com/benthecarman/open-ssp now we can use spark on any network, not just the ones lightspark blesses. This adds support for signet to the breez sdk so we can utilize it for signet and mutinynet.

--

Expose Signet through the SDK, CLI, and binding definitions while preserving public function signatures. Require explicit operators and SSP configuration, retaining the public Signet chain API as the default.

Allow Signet initialization without a Breez API key and use test-network sync keys. Reject unsupported Turnkey Signet accounts before making API requests. Cover configuration validation, initialization, and routing.


kingonly:
@danielgranhao please follow up

crates/breez-sdk/core/src/signer/rtsync.rs
```diff
@@ -28,12 +28,12 @@ impl RTSyncSigner {
...
         let signing_path: DerivationPath = match network {
             Network::Mainnet => SIGNING_DERIVATION_PATH,
-            Network::Regtest => SIGNING_DERIVATION_PATH_TEST,
+            Network::Signet | Network::Regtest => SIGNING_DERIVATION_PATH_TEST,
         }
         .parse()?;
         let encryption_path: DerivationPath = match network {
             Network::Mainnet => ENCRYPTION_DERIVATION_PATH,
-            Network::Regtest => ENCRYPTION_DERIVATION_PATH_TEST,
+            Network::Signet | Network::Regtest => ENCRYPTION_DERIVATION_PATH_TEST,
```
danielgranhao: With regtest and signet sharing the sync signing key, using the same seed on both networks will make them share the same namespace. It's worth giving signet its own account index. 

The encryption key being the same isn't problematic, but also worth changing to avoid the asymmetry

crates/breez-sdk/cli/src/main.rs
```diff
@@ -342,7 +343,12 @@ async fn main() -> Result<(), anyhow::Error> {
     let network = match cli.network.to_lowercase().as_str() {
         "regtest" => Network::Regtest,
         "mainnet" => Network::Mainnet,
-        _ => return Err(anyhow!("Invalid network. Use 'regtest' or 'mainnet'")),
+        "signet" => Network::Signet,
```
danielgranhao, line 386: The CLI currently always builds its config without `spark_config` so choosing signet will always fail. Can we add a param to load the operators/SSP from a JSON file so it's actually usable?

danielgranhao (commented):
Looks good!

crates/breez-sdk/core/src/signer/rtsync.rs
benthecarman: fixed

crates/breez-sdk/cli/src/main.rs
benthecarman: added

danielgranhao (approved):
LGTM

crates/breez-sdk/core/src/sdk_builder.rs
```diff
@@ -901,6 +902,14 @@ fn resolve_chain_service(
                 None,
                 ChainApiType::MempoolSpace,
             )),
+            Network::Signet => Arc::new(RestClientChainService::new(
+                "https://blockstream.info/signet/api".to_string(),
```
JssDWt: Perhaps we should require users using signet to pass their own chain service? I'm not sure how useful this default is.

crates/breez-sdk/core/src/sdk/mod.rs
```diff
@@ -231,7 +231,7 @@ pub async fn connect_with_signing_only_signer(
 pub fn default_config(network: Network) -> Config {
     let lnurl_domain = match network {
         Network::Mainnet => Some("breez.tips".to_string()),
-        Network::Regtest => None,
+        Network::Signet | Network::Regtest => None,
```
JssDWt, line 234: In crates/spark/src/signer/default_signer.rs, there's the `default_account_number` function. I think that should return 0 for Signet, so it matches Regtest, not Mainnet.

JssDWt (changes requested):
Thanks for the PR! A couple of small things:

crates/breez-sdk/core/src/sdk/mod.rs
benthecarman: fixed, also did same for testnet

crates/breez-sdk/core/src/sdk_builder.rs
benthecarman: added

crates/spark/src/signer/default_signer.rs
```diff
@@ -489,6 +490,37 @@ pub(crate) mod tests {
...
+        let signet_key = identity_public_key(&seed, Network::Signet, None).unwrap();
+        assert_eq!(
+            signet_key,
+            identity_public_key(&seed, Network::Regtest, None).unwrap()
+        );
+        assert_ne!(
+            signet_key,
+            identity_public_key(&seed, Network::Mainnet, None).unwrap()
+        );
+        // Pinning account 1 still recovers wallets created with the old default.
```
JssDWt: nit: there is no old default
```suggestion
```

JssDWt (approved):
LGTM

PR #1131, opened 2026-09-17
Document treasury wallets

danielgranhao:
Partners asking about treasury wallets had nothing to read. Our only server-side docs are for server mode, which is a different shape: a wallet per user, an SDK instance per request. A treasury is one long-lived wallet the service itself owns, an ordinary deployment with a few defaults worth changing.

The advanced section now has a "Server-side configuration" page with two under it: "Multi-user configuration" (the old server mode page, renamed) and a new "Treasury configuration".

Writing it turned up that with `exit_chain_auto_fetch_enabled` off there was no way to download ancestors at all, which is what a treasury with many leaves most wants to control. The flag now governs only the automatic collection, and an explicit `sync_wallet` always collects.


roeierez (approved):

docs/breez-sdk/snippets/flutter/lib/sdk_building.dart
```diff
@@ -54,66 +54,6 @@ Future<void> withAccountNumber(SdkBuilder builder) async {
 // ANCHOR: with-session-store
 // ANCHOR_END: with-session-store
 
-Future<BreezSdk> initSdkServer() async {
-  // ANCHOR: init-sdk-server
```
JssDWt, line 58: Accidentally deleted? Or on purpose?

init-sdk-server, server-mode-request-handler and server-mode-provisioning from flutter/lib/sdk_building.dart and react-native/sdk_building.ts

docs/breez-sdk/src/guide/treasury.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Choose when to download unilateral exit data
+
+By default the SDK collects the data a [unilateral exit](unilateral_exit.md) needs as funds arrive, after an operation rather than during it. On a treasury that pays frequently and holds many leaves, that is a round trip behind almost every operation.
+
+Set {{#name exit_chain_auto_fetch_enabled}} to `false` to stop it, then call {{#name sync_wallet}} on a cadence of your own to collect:
+
+{{#tabs unilateral_exit:sync-exit-data}}
+
+The flag governs only the automatic collection, so a sync you call yourself collects regardless of it, and waits for the pass before returning. Nothing else has to change: the leaf set is still kept current for you by the background sync and the operator event stream.
```
JssDWt: This sentence is not readable.

docs/breez-sdk/src/guide/treasury.md
```diff
@@ -0,0 +1,35 @@
...
+
+Between those syncs, recently received funds are not exitable without the operators. The gap you choose is the window you are accepting, so leave the default alone if the automatic collection does not show up in your metrics.
```
JssDWt: "The gap you choose is the window you are accepting, so leave the default alone if the automatic collection does not show up in your metrics." -> I think this could use some rephrasing in terms of human language as well.

docs/breez-sdk/src/guide/unilateral_exit.md
```diff
@@ -10,7 +10,7 @@ A unilateral exit is a last resort. It is a multi-step, on-chain process that ne
...
+- **The exit data has to already be on the device.** Quoting and building an exit read each leaf's pre-signed transactions from local storage, so both work with the operators unreachable. What they cannot do is obtain that data: a leaf can be exited this way only once it has been synced at least once while the operators were reachable. The SDK collects it as funds arrive, in the background where background services run, which you can turn off with [{{#name exit_chain_auto_fetch_enabled}}](./config.md#unilateral-exit-data). {{#name sync_wallet}} collects regardless of that flag and waits for the pass before returning, so calling a sync yourself runs the collection at a moment of your choosing: that is how the data is kept current with the automatic collection off, and it is worth a call before going offline either way. Once collected it can be kept outside the SDK's storage, see [Back up the exit data](#back-up-the-exit-data).
```
JssDWt:
```suggestion
- **The exit data has to already be on the device.** Quoting and building an exit read each leaf's pre-signed transactions from local storage, so both work with the operators unreachable. What they cannot do is obtain that data: a leaf can be exited this way only once it has been synced at least once while the operators were reachable. The SDK collects it as funds arrive, in the background where background services run, which you can turn off with [{{#name exit_chain_auto_fetch_enabled}}](./config.md#unilateral-exit-data). {{#name sync_wallet}} collects regardless of that flag and waits for the pass before returning, so calling a sync yourself runs the collection at a moment of your choosing. Once collected it can be kept outside the SDK's storage, see [Back up the exit data](#back-up-the-exit-data).
```

docs/breez-sdk/src/guide/server_deployments.md
```diff
@@ -0,0 +1,17 @@
+# Server deployments
+
+The SDK runs on a server in two deployments that have little in common beyond the machine they share. They pull in opposite directions, so pick the one that matches yours before taking any configuration advice.
```
JssDWt: I know this needs changing, but not sure what.

JssDWt (commented):
Looks good. Some doc nits.

kingonly:
I think we should be  clearer and more consistent with the terminology. 
* Server-side configuration
* * Multi-user configuration 
* * Treasury configuration 

I haven't read this docs, but keep it succinct, technical, readable, short with no sloppy text (that's an instruction to Claude. Fable 5.1 only).

docs/breez-sdk/snippets/flutter/lib/sdk_building.dart
danielgranhao: On purpose. Flutter and react native don't make sense on the server side.

kingonly:
Go ahead and merge to the release branch, must be better than what we have now...

PR #1136, opened 2026-09-18
docs: Add a table to clarify quote state and what the SDK will do

dangeross:
Add more explanation about the early quote, what the SDK does depending on the quote, depth and fee ceiling, and UX suggestions on how to present the deposit


kingonly:
No need for approval, go ahead

Issue #1138, opened 2026-09-18
Cross-chain receive fails on higher-fee networks


erdemyerebasmaz:
Receiving USDT or USDC fails on some networks because the SDK turns down the provider's quote: the deposit it asks for doesn't cover that network's fees.

- At $5 it fails on Ethereum, Tron, BNB and HyperEVM. At $50, Ethereum and Tron still fail.
- Arc, TON and HyperCore can't be used at all: 
  - Arc has no Bitcoin destination, 
  - TON needs a refund address, 
  - and the provider refuses HyperCore.

We'd like every listed network to work at everyday amounts, or not be listed.

dangeross:
- Arc has a Bitcoin destination and I can create a $2 receive address (min amount is $2)
- Can create a $5 receive address for BNB and HyperEVM. Lower amounts exceed the max slippage bps
- Just tested TON and HyperCore, they do return provider errors ("A TON refund address is required", "This route is not supported")
- Just tested Ethereum and Tron, their quotes exceed the 100 max slippage at 180 and 582 bps respectively. Maybe the error returned needs improving and a more informative error displaying in Glow

kingonly:
Should be in the release. @dangeross if settings a something like $5 minimum across the board makes sense - I don't mind.

erdemyerebasmaz:
Tested in sats mode:

> Just tested Ethereum and Tron, their quotes exceed the 100 max slippage at 180 and 582 bps respectively. Maybe the error returned needs improving and a more informative error displaying in Glow

- Ethereum and Tron come back short because `/estimate` leaves out the source network's own cost. #1140 quotes again with the shortfall added.

> Arc has a Bitcoin destination and I can create a $2 receive address (min amount is $2)

- Arc only works for a wallet holding USD balance.

Base and Polygon fail now and then on provider errors, then work minutes later.

Would help apps: typed errors for these cases, leaving out routes a wallet can't complete, and one retry on a provider 503.

<details><summary>Details</summary>

- `/estimate` gives Ethereum and Arbitrum the same 6,147 sats for $5, while the firm quote takes about 2,000 sats of gas out of the output without a fee field. With #1140, a $5 receive asks the sender for about $6.16 from Ethereum and $8.43 from Tron; no other route needed the second quote.
- Arc's only Spark-side asset is USDB, so in sats mode the SDK finds no destination ("Route exposes no usable Spark destination").
[...]

PR #1140, opened 2026-09-18
cross-chain: quote again when a quote comes back short

erdemyerebasmaz:
Part of #1138

Receiving USDT or USDC from Ethereum or Tron failed at everyday amounts. The SDK sized the deposit from a price estimate that leaves out the network's own costs, so the firm quote came back short and was rejected.

When a quote comes back short, the SDK now adds the shortfall to the deposit and quotes once more. A $5 receive now asks the sender for about $6.16 from Ethereum and $8.43 from Tron, where it used to fail. Quotes that come back close enough are unaffected.


dangeross (commented):
I'm trying to look at this with my security hat on. verify_quote_not_drifted is there to catch a quote coming back short of what we sized from the estimate, to the max_slippage_bps limit. This fix is just bypassing that. The provider controls estimatedOut, so they control the shortfall we read off it and how much we top up in response.

I'm thinking we need something like a max_total_cost_bps knob on the request which bounds the requote, and a clear typed error that gives the max_total_cost_bps needed

erdemyerebasmaz:
Added `max_requote_cost_bps`: the re-quote now has a ceiling in bps of the deposit it was sized at (request override, config default, built-in 5000), and a quote needing more fails with `CrossChainRequoteTooExpensive` carrying the bps it would have taken. The drift check still runs on whatever quote comes back.

erdemyerebasmaz:
> I'm trying to look at this with my security hat on. verify_quote_not_drifted is there to catch a quote coming back short of what we sized from the estimate, to the max_slippage_bps limit. This fix is just bypassing that. The provider controls estimatedOut, so they control the shortfall we read off it and how much we top up in response.
> 
> I'm thinking we need something like a max_total_cost_bps knob on the request which bounds the requote, and a clear typed error that gives the max_total_cost_bps needed

~~@dangeross 6131c69 adds `max_requote_cost_bps` to cap how much extra a re-quote can cost relative to the original deposit. The limit can be set per request or configured globally (defaults to 5000 bps).~~

~~Re-quotes exceeding the limit now fail with `CrossChainRequoteTooExpensive`, carrying the bps it would have taken. The existing drift check still applies to the resulting quote.~~

Edit: **Reverted**

erdemyerebasmaz:
Good catch: `quoted_in` is the deposit sized from the estimate, so the provider sets the denominator, and a padded estimate makes the same increase look smaller in bps. I will anchor it on the requested amount instead, valued with the SDK's own rates (par for USD-stable, the fiat service for BTC), which the provider does not control. That also catches a first quote that is simply expensive, so the knob becomes a cap on total cost rather than on the increase.

That needs a number picked: measured that way a $5 receive costs about 2300 bps from Ethereum and about 6900 from Tron, so a 5000 bps default would reject Tron and return the bps it needs. Happy to raise the default instead if you would rather those routes keep working.

crates/breez-sdk/core/src/cross_chain/orchestra.rs
```diff
@@ -868,10 +868,66 @@ impl OrchestraService {
...
+        max_slippage_bps: u32,
+    ) -> Result<QuoteResponse, SdkError> {
+        let delivered = parse_amount(&quote.estimated_out, "estimatedOut")?;
+        if delivered >= min_acceptable_out(drift_target, max_slippage_bps) {
+            return Ok(quote);
+        }
+        let quoted_in = parse_amount(&quote.amount_in, "amountIn")?;
+        let top_up = top_up_for_shortfall(
+            sized.target.saturating_sub(delivered),
+            sized.probe_in,
```
dangeross: Could we check the finished deposit against `probe_in` on its own? At the moment the shortfall and the rate both come from them, so the top-up ends up being whatever they say it is.

dangeross (approved):

crates/breez-sdk/core/src/cross_chain/orchestra.rs
```diff
@@ -1269,18 +1347,25 @@ impl CrossChainService for OrchestraService {
...
             .client
-            .quote(request)
+            .quote(request.clone())
             .await
             .map_err(|e| with_limits(SdkError::from(e)))?;
         debug!("Orchestra: quote response: {:?}", quote);
+        verify_quote_amount_in(source_amount, parse_amount(&quote.amount_in, "amountIn")?)?;
+        let quote = match &sized {
+            Some(sized) => self
+                .requote_if_short(request, quote, sized, sized.target, max_slippage_bps)
```
danielgranhao: Do we also want to requote on sends?

danielgranhao (commented):
Looks good

danielgranhao (approved):

PR #1141, opened 2026-09-18
cross-chain: leave out receive routes a wallet can't complete

erdemyerebasmaz:
Part of #1138

The receive route list included routes that fail on every attempt: Arc when the wallet isn't holding USD balance (its only destination is USDB), and TON, which the provider refuses without a refund address that a receive request can't carry.

Receive routes are now listed only when a receive lands in Bitcoin or the wallet's active stable-balance token without naming a destination, and TON is left out.

An app that names a token destination itself would no longer see a token-only route in sats mode. Flagging these routes instead of hiding them would keep that open, if anyone needs it.


dangeross (approved):
LGTM

PR #1142, opened 2026-09-18
cross-chain: tell a refused or unavailable route apart from other errors

erdemyerebasmaz:
Part of #1138

When the provider won't serve a route, apps got a generic network error carrying the provider's text, so they couldn't tell "try again in a moment" from "pick another network".

This adds `SdkError::CrossChainRouteUnavailable`, with `temporary` saying whether the same request can succeed later. Its message starts with a fixed phrase, so apps that only see the text, like those on the WASM package, can recognise it too.

The refund address case isn't covered: its error code isn't documented, and #1141 leaves those routes out instead.


docs/breez-sdk/src/guide/cross_chain.md
```diff
@@ -72,6 +72,8 @@ Validate the amount against the bounds before preparing the payment. A route can
 
 A rejected amount surfaces as {{#enum SdkError::CrossChainAmountOutOfRange}}, carrying {{#name too_small}} for the direction and the published bound in whichever denominations the provider publishes.
 
+A route the provider won't serve surfaces as {{#enum SdkError::CrossChainRouteUnavailable}}. {{#name temporary}} tells a route that should be back shortly, where the same request can succeed later, from one that is refused, where another route is the way forward.
```
dangeross: nit: I think this needs tightening, something like

> {{#name temporary}} indicates that a route is currently unavailable and the same request may succeed later, rather than requiring another route.

crates/flashnet/src/orchestra/client.rs
```diff
@@ -385,6 +386,19 @@ fn error_from_body(body: &str, status: u16) -> FlashnetError {
             reason,
             too_small: false,
         },
+        Some(
+            "route_unavailable"
+            | "estimate_unavailable"
```
dangeross, line 391: Are these observed error codes? I don't see them documented

dangeross (commented):
Looking good

docs/breez-sdk/src/guide/cross_chain.md
erdemyerebasmaz: Taken, pushed.

crates/flashnet/src/orchestra/client.rs
erdemyerebasmaz: `unsupported_route`, `route_disabled`, `service_unavailable` and `spot_unavailable` are from Flashnet's documented error list at: https://docs.flashnet.xyz/api/errors.

`route_unavailable` is referenced by `/limits` and has been observed in live 400 responses, while `estimate_unavailable` has only been observed in 503 responses and is not documented.

dangeross (approved):
LGTM

PR #1143, opened 2026-09-19
add an SSP daemon

JssDWt:
An attempt at an open source implementation of an SSP. Lightspark's SSP is closed source, so our tests have to point at the live regtest environment for many flows, and Lightning is hard to test. I did my best to isolate this change to only the added ssp.

It automatically creates leaf trees from on-chain funds, much like the deployed SSP does.

- Swaps
- Lightning receive
- Lightning send
- Cooperative exit
- Static deposit
- Zero-conf static deposit
- On-chain wallet

The SSP uses bitcoind and ldk-server. It serves an internal gRPC API, with `ssp-cli` as a small CLI over it.

ldk-server runs from the `ssp` branches of the breez forks of ldk-server, ldk-node and rust-lightning, which:
- Support self-payments
- Take `min_final_cltv_expiry_delta` on a hold invoice
- Report a held payment's claim deadline and claimable amount
- Report a failed send under its own error code
- Refuse to pay an invoice twice

The SSP needs an operator fork to expose SO-SSP communications: breez/spark `ssp-enabled` (#1144), which:
- Serves `SparkSspInternalService`, the API Lightspark's operators serve their SSP on: tree creation, counter transfers and returning a stuck transfer
- Restores Lightspark's static deposit claim, zero-conf static deposit claim and static deposit sweep
- Answers a retried zero-conf static deposit reservation or claim with its result
- Has a knob that shortens a scheduled task's interval, for tests


roeierez (approved):
Fantastic!

dangeross (approved):
LGTM 👏

PR #1147, opened 2026-09-19
Move to spark-primitives 0.2.0

peterrojs:
`spark-token-primitives` was renamed to `spark-primitives` upstream (lightsparkdev/spark#8426) now that it covers transfer manifests, quote envelopes, and invoices as well as token transactions. `spark-token-primitives` stays at 0.1.3 on crates.io and will not get further releases, so this moves the workspace dependency to `spark-primitives = "0.2.0"` and updates the two module paths in `crates/spark`.

the thirteen symbols used here (`construct_partial_transfer_transaction`, `build_broadcast_transaction_request`, `hash_final_token_transaction`, `prepare_token_invoice`, `finalize_token_invoice`, and their request and output types) are unchanged between 0.1.3 and 0.2.0, so no call sites change. since 0.1.3 the crate switched its token-transaction hashing from a hand-rolled per-message walker to a generic protobuf reflection walker, which is where the new `prost-reflect`, `logos`, and `unicode-normalization` entries in `Cargo.lock` come from. the cross-implementation hash vectors in `crates/spark` pass unchanged, which is the check that matters for that switch.

## Test plan

- `cargo test -p spark` (442 passed, including the token transaction and spark address hash vector suites)
- `make cargo-clippy-check` and `cargo fmt --all -- --check`
- `cargo build -p cli` and ran `cli --help`
- `cargo check -p spark-primitives --target wasm32-unknown-unknown` upstream, since `prost-reflect` and `logos` are new to the wasm build here; the repo's own `make wasm-clippy-check` needs homebrew llvm locally so that is left to ci


roeierez (approved):
LGTM

danielgranhao (approved):
LGTM, just some lockfiles needed updates to keep CI happy. Pushed that change already

Issue #1137, opened 2026-09-18
WASM: cross-chain sends hang


erdemyerebasmaz:
On the WASM package, sending to another chain never completes: preparing the payment neither returns nor fails, so the app waits forever. It started with the route amount limits added in #1110 and affects every cross-chain send paid in sats. 

What fails is passing a route back into the SDK.

What we'd like:
- Routes can be passed back exactly as they come out.
- A call that fails rejects its promise, so apps can show an error instead of hanging.

Issue #347, opened 2025-10-22
Improve Wasm mapping of u128


danielgranhao:
Currently, some u128 fields in the SDK's API need to be exposed as Strings on Wasm due to this [serde issue](https://github.com/serde-rs/json/issues/625). 

We should figure out a fix or workaround so that all u128 values are properly mapped to js `BigInt`.

danielgranhao:
There is an attempt to fix the serde issue: https://github.com/serde-rs/serde/pull/2781

PR #1148, opened 2026-09-19
wasm: catch u128 values JS can't pass back

erdemyerebasmaz:
#1137 happened because a u128 inside a tagged enum reaches JS as a BigInt that can't be passed back, so the call hangs. This adds a test that fails for any such field, so the next one shows up in CI instead of in an app.

It found one more: publishing a signed token batch package hung the same way. `BatchTotal.amount` is now a string in TypeScript, like the package's other token amounts.


crates/breez-sdk/wasm/src/models/tests.rs
```diff
@@ -68,3 +71,74 @@ fn route_filter_accepts_address_details_with_an_amount() {
...
+    "PublishSignedLnurlPayResponse",
+    "PublishSignedTransferPackageResponse",
+];
+
+// The rule behind the tests above, checked across every model: the models
+// macro tags each enum that carries data, so a u128 anywhere below one must use
+// `serde_u128_as_string` or `serde_option_u128_as_string`.
+#[wasm_bindgen_test]
+fn tagged_enums_reach_no_bigint() {
+    let file = syn::parse_file(include_str!("mod.rs")).unwrap();
```
dangeross: There are several other places that expose types also and I looks like issuer.rs is definitely affected. Should we include issuer.rs and signer.rs also?

erdemyerebasmaz: Definitely. I'll check and include the remaining files.

erdemyerebasmaz: Done.
- Added all remaining mirrored files 56bf37d
- Added support for flattened Serde fields for future-proofing 4adc006

JssDWt:
I would say it's problematic that some u128 numbers are BigInt and some are string. I'm looking for a general solution that can make everything BigInt. 

Perhaps:
> Generate the tag dispatch in extern_wasm_bindgen instead of letting serde do it. The macro already owns these enums; it would emit its own from_wasm_abi that reads the type property off the JsValue with Reflect, then deserializes the rest into the variant. No serde-level tagging, no buffer, identical TypeScript shape.

danielgranhao (approved):

danielgranhao:
> I would say it's problematic that some u128 numbers are BigInt and some are string. I'm looking for a general solution that can make everything BigInt.
> 
> Perhaps:
> 
> > Generate the tag dispatch in extern_wasm_bindgen instead of letting serde do it. The macro already owns these enums; it would emit its own from_wasm_abi that reads the type property off the JsValue with Reflect, then deserializes the rest into the variant. No serde-level tagging, no buffer, identical TypeScript shape.

We've looked into this several times in the past and didn't find a better solution then. But yes, it would definitely be great to improve. There's an existing issue tracking it: https://github.com/breez/spark-sdk/issues/347

erdemyerebasmaz:
I’d prefer to keep #347 separate and address the general `u128` handling there rather than expand the scope of this PR.

Merging this PR as-is for the upcoming release.

